PR #236, opened 2018-06-11
Expose the `publish` property on aws.serverless.Function

lukehoban:
Workaround for the issue in https://github.com/pulumi/pulumi-aws-serverless/issues/9.  Longer term we should just support all `aws.lambda.FunctionArgs` inputs here.


lukehoban:
This is no longer needed now that all lambda function options can be passed.

PR #243, opened 2018-06-14
Update READMEs.

pgavlin:
- Remove stale information from example READMEs
- Remove the private beta warning from the package README


joeduffy:
@pgavlin Is this still in progress?

PR #251, opened 2018-06-28
Add autoscaling NotificationType union and overlay

jen20:
This pull request adds a new overlay for nodejs, defining a union type of valid Autoscaling Notification Types, and overrides the `notifications` property of the aws.autoscaling.Notification type to accept an array of these instead of raw strings.

Some outstanding questions:

- [x] What copyright should go on the new file
- [x] Is there work here to be done for other SDKs? Currently only Node appears to have overlays

I've opened the pull request early in order to take advantage of the properly configured Travis CI building for PRs, and don't anticipate this will be ready to merge without further work.

### DIFF
--- a/overlays/nodejs/autoscaling/notificationType.ts
+++ b/overlays/nodejs/autoscaling/notificationType.ts
@@ -1,0 +1,34 @@
+// Copyright 2016-2018, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file contains two categories of exports:
+//
+//     1) A union type, NotificationType, that accepts any valid AutoScaling Notification type
+//     2) Individual constants for each such AutoScaling Notification type
+//
+// These give a better developer experience and are just sugared strings.
+
+export let InstanceLaunchNotification:         NotificationType = "autoscaling:EC2_INSTANCE_LAUNCH";
+export let InstanceTerminateNotification:      NotificationType = "autoscaling:EC2_INSTANCE_TERMINATE";
+export let InstanceLaunchErrorNotification:    NotificationType = "autoscaling:EC2_INSTANCE_LAUNCH_ERROR";
+export let InstanceTerminateErrorNotification: NotificationType = "autoscaling:EC2_INSTANCE_TERMINATE_ERROR";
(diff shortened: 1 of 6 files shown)


resources.go
```diff
@@ -381,9 +381,16 @@ func Provider() tfbridge.ProviderInfo {
...
 				},
 			},
-			"aws_autoscaling_notification": {Tok: awsResource(autoscalingMod, "Notification")},
-			"aws_autoscaling_policy":       {Tok: awsResource(autoscalingMod, "Policy")},
-			"aws_autoscaling_schedule":     {Tok: awsResource(autoscalingMod, "Schedule")},
+			"aws_autoscaling_notification": {
+				Tok: awsResource(autoscalingMod, "Notification"),
+				Fields: map[string]*tfbridge.SchemaInfo{
+					"notifications": {
+						Type: awsType(tokens.ArrayTypePrefix+autoscalingMod+"notificationType", "NotificationType"),
```
lukehoban: Just for consistency with other uses of this approach - could we use this style instead:

```
awsType(apigatewayMod+"/restApi", "RestApi"),
```

overlays/nodejs/autoscaling/notificationType.ts
```diff
@@ -0,0 +1,19 @@
+// This file contains two categories of exports:
```
lukehoban: Could you please include the standard copyright/license banner on the new file.

overlays/nodejs/autoscaling/notificationType.ts
```diff
@@ -0,0 +1,19 @@
...
+//
+//     1) A union type, NotificationType, that accepts any valid AutoScaling Notification type
+//     2) Individual constants for each such AutoScaling Notification type.
+//
+// These just give a better developer experience and are just sugared strings.
```
lukehoban: Could we add a link to primary source in AWS docs for where these strings are defined?

overlays/nodejs/autoscaling/notificationType.ts
jen20: I've added the copyright notice from other files in the latest commit.

resources.go
```diff
@@ -381,9 +381,16 @@ func Provider() tfbridge.ProviderInfo {
...
 				},
 			},
-			"aws_autoscaling_notification": {Tok: awsResource(autoscalingMod, "Notification")},
-			"aws_autoscaling_policy":       {Tok: awsResource(autoscalingMod, "Policy")},
-			"aws_autoscaling_schedule":     {Tok: awsResource(autoscalingMod, "Schedule")},
+			"aws_autoscaling_notification": {
+				Tok: awsResource(autoscalingMod, "Notification"),
+				Fields: map[string]*tfbridge.SchemaInfo{
+					"notifications": {
+						Type: awsType(tokens.ArrayTypePrefix+autoscalingMod+"/notificationType", "NotificationType"),
```
jen20: Is the `tokens.ArrayTypePrefix` here necessary? And if it is, would it be better abstracted behind some kind of `awsArrayType` function that adds it automatically?

I've made the other modification to add the missing `/` to the tag.

jen20:
I've fixed up the CI here and committed the generated code. Hopefully this should now be good to go, pending a Travis run.

PR #254, opened 2018-06-30
Add `environment` to `aws.serverless.Function`

jen20:
This is targeted fix to add an environment variable map to the `aws.serverless.Function` type, as an interim step to allowing it to configure all facets of a Lambda function.

### DIFF
--- a/overlays/nodejs/serverless/function.ts
+++ b/overlays/nodejs/serverless/function.ts
@@ -84,6 +84,10 @@ export interface FunctionOptions {
         subnetIds: pulumi.Input<string[]>,
     };
     /**
+     * The Lambda environment's configuration settings.
+     */
+    environment?: pulumi.Input<{ variables?: pulumi.Input<{[key: string]: pulumi.Input<string>}> }>;
+    /**
      * The paths relative to the program folder to include in the Lambda upload.  Default is `[]`.
      */
     includePaths?: string[];
@@ -169,6 +173,7 @@ export class Function extends pulumi.ComponentResource {
             code: new pulumi.asset.AssetArchive(codePaths),
             handler: serializedFileName + "." + handlerName,
             runtime: options.runtime || lambda.NodeJS8d10Runtime,
+            environment: options.environment,
             role: this.role.arn,
             timeout: options.timeout === undefined ? 180 : options.timeout,
             memorySize: options.memorySize,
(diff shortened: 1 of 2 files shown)


ellismg:
Thanks for the patch, @jen20.  The Travis build is red due to some infrastructure issues on our end.  I'm working on fixing them up.

overlays/nodejs/serverless/function.ts
```diff
@@ -83,6 +83,10 @@ export interface FunctionOptions {
         securityGroupIds: pulumi.Input<string[]>,
         subnetIds: pulumi.Input<string[]>,
     };
+    /**
+     * The Lambda environment's configuration settings.
+     */
+    environment?: {[key: string]: string}
```
lukehoban: We haven't been consistent about this in other types here, but you could make this flexible to use with outputs from other Pulumi resources by using the same type (with `pulumi.Input` at each layer) as on the underlying `aws.lambda.FunctionArgs`:

```
environment?: pulumi.Input<{ variables?: pulumi.Input<{[key: string]: pulumi.Input<string>}> }>;
```

overlays/nodejs/serverless/function.ts
```diff
@@ -155,12 +159,13 @@ export class Function extends pulumi.ComponentResource {
...
 
         let codePaths = computeCodePaths(closure, serializedFileName, options.includePaths, options.includePackages);
-        
+
         // Create the Lambda Function.
         this.lambda = new lambda.Function(name, {
             code: new pulumi.asset.AssetArchive(codePaths),
             handler: serializedFileName + "." + handlerName,
             runtime: options.runtime || lambda.NodeJS8d10Runtime,
+            environment: options.environment,
```
lukehoban: Note that you'll either need to make the `options.environment` use the nested `variables` property as on the underlying `aws.lambda.FunctionArgs`, or you will need to pass the `options.environment` as the `variables` property of this object.

Probably worth aligning with `aws.lambda.FunctionArgs`, which will require changing the signature above.

jen20: I changed the signature above, so this should no longer apply.

jen20:
I've pushed a new commit here which includes the generated code to fix the CI build.

joeduffy:
@lukehoban , looks like @jen20 addressed the CR.  Shall we merge this in time to make our next package release tomorrow?

PR #255, opened 2018-06-30
Update sources and ensure they can't get out of sync

ellismg:


ellismg:
Due to an issue on my end when preparing this (I had not run `dep ensure`) I think the first commit which updates the sources is incorrect.  So hopefully, we'll see this PR go /red/ which will prove things are working as expected.  Then I'll actually check in the correct sources.

PR #259, opened 2018-07-03
Bump terraform reference

ellismg:


ellismg:
Some build failures (see: https://travis-ci.com/pulumi/pulumi-aws/builds/78028831) due typescript compilation issues because a type could not be named.  I think that removing: `declaration: true` will address this and it is an artifact of the crazy environment we test in (where we have a mix of installed packages and a yarn linked stuff).

I will push another commit, but we should talk about this in person @lukehoban.

Issue #260, opened 2018-07-04
Could not include required dependency warning


joeduffy:
I receive the warning

```
info: Could not include required dependency '@slack/client' in '/Users/joeduffy/temp/post-sqs-slack'.
```

for the below lambda.  I can't figure out why it's missing it, since it definitely does exist.  Is there a bug in the way we regex match scoped packages, because they exist inside of subdirectories?

```typescript
import * as pulumi from "@pulumi/pulumi";
import * as aws from "@pulumi/aws";
import * as serverless from "@pulumi/aws-serverless";

let config = new pulumi.Config(pulumi.getProject());
let queue = aws.sqs.Queue.get("my-queue", config.require("queueName"));
let slackChannel = config.get("slackChannel") || "#general";
let slackToken = config.require("slackToken");
serverless.queue.subscribe("my-slack-poster", queue, async (e: serverless.queue.QueueEvent) => {
    const slack = await import("@slack/client");
    const client = new slack.WebClient(slackToken);
    for (let rec of e.Records) {
        await client.chat.postMessage({
            channel: slackChannel,
            text: `*SQS message ${rec.messageId}* in region ${rec.awsRegion}:\n`+
                `${rec.body}\n`+
                `(with :love_letter: from Pulumi)`,
            as_user: true,
        });
    }
});
```

joeduffy:
I suspect https://github.com/pulumi/pulumi-aws/blob/7786683cdeb1679cacd9752b4b88a6ebc7d2caf8/overlays/nodejs/serverless/function.ts#L315 is the culprit here, but I really don't follow what that code is trying to do.

Attempting a fix now, since this is blocking the demo I had wanted to do, ... please shout if anybody sees or knows anything about the underlying issue here.

joeduffy:
Yep ☹️ 

Notice the leading slash in `/\/\@[^\/]*\/[^\/]*$/.exec(child.path)`.  This turns names like `@protobufjs/fetch` into `/@protobufjs/fetch`, which of course will not match the desired package name.

Simple one-line fix, which I have locally, and will push.  We need tests.

lukehoban:
Argh: https://github.com/pulumi/pulumi-aws/pull/230#discussion_r193900791

The attempt to address that feedback led to this incorrect leading slash in the regexp.  I believe it works for nested dependencies in organizations, but not for top level dependencies in organizations.

> We need tests.

:100:

joeduffy:
[...]

PR #262, opened 2018-07-04
Fix detection of org-scoped packages

lukehoban:
Alternative fix for #260, adds tests and makes a more targeted fix than #261 which appears to introduce other issues.

Use path APIs instead of regexp in workaround for incorrect package names coming from `read-package-tree` dependency when seeing organization-scoped packages.

Add tests for additional cases of organization-scoped packages (at the root as well as nested).

### DIFF
--- a/overlays/nodejs/serverless/function.ts
+++ b/overlays/nodejs/serverless/function.ts
@@ -315,11 +315,13 @@ function findDependency(root: Package, name: string) {
         for (var child of root.children) {
             let childName = child.name;
             // Note: `read-package-tree` returns incorrect `.name` properties for packages in an orgnaization - like
-            // `@types/express` or `@protobufjs/path`.  Compute the correct name from the `path` property instead.
-            // Match any name that ends with something that looks like `@foo/bar`.
-            const match = /\@[^\/]*\/[^\/]*$/.exec(child.path);
-            if (match) {
-                childName = match[0];
+            // `@types/express` or `@protobufjs/path`.  Compute the correct name from the `path` property instead. Match
+            // any name that ends with something that looks like `@foo/bar`, such as `node_modules/@foo/bar` or
+            // `node_modules/baz/node_modules/@foo/bar.
+            const childFolderName = filepath.basename(child.path);
+            const parentFolderName = filepath.basename(filepath.dirname(child.path));
+            if (parentFolderName[0] == "@") {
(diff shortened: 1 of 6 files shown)


overlays/nodejs/serverless/function.ts
```diff
@@ -309,17 +311,17 @@ function addPackageAndDependenciesToSet(s: Set<string>, root: Package, pkg: stri
...
 function findDependency(root: Package, name: string) {
-    for(;root;root = root.parent) {
+    for (; root; root = root.parent) {
         for (var child of root.children) {
             let childName = child.name;
             // Note: `read-package-tree` returns incorrect `.name` properties for packages in an orgnaization - like
             // `@types/express` or `@protobufjs/path`.  Compute the correct name from the `path` property instead.
             // Match any name that ends with something that looks like `@foo/bar`.
-            const match = /\/\@[^\/]*\/[^\/]*$/.exec(child.path);
+            const match = /\@[^\/]*\/[^\/]*$/.exec(child.path);
```
joeduffy: I personally vote against using regexes to match paths like this.  I'd prefer to see us using real filesystem operations that parse the path more precisely.  (For example, this regex will now match `some@foo/bar`, which I'm not sure is intentional?  And if it is, I'd rather state this more precisely than just "happening" to match it using a regex.)

lukehoban: Yes - we’ll move this over to real path operations next. Just opened this to verify that things were working correctly with the minimal change.

joeduffy:
Great to have a test (thought it probably makes sense to pull it out of `examples/` and into `tests/` -- realize this is a work in progress).  I closed out #261.

I assume you didn't see the other issue I saw with a missing `loglevel` library?  (Since the tests passed, I assume not ... however, it could just be that a different set of packages lead to a different layout.)

Do you think we can land today?  I was hoping to focus on this stuff with my blog post, although worst case I can workaround the issue with some package renaming (or just not use Slack).

joeduffy:
> Do you think we can land today?

By "land", I meant commit and cut a new release of `@pulumi/aws`.

If not today, tomorrow is fine also.

lukehoban:
> I assume you didn't see the other issue I saw with a missing loglevel library? (Since the tests passed, I assume not ... however, it could just be that a different set of packages lead to a different layout.)

Yeah - per

joeduffy:
Awesome 🎉

PR #265, opened 2018-07-06
Only apply AutoName to inputs

### LINKED ISSUE #250: aws:apigateway/usagePlanKey:UsagePlanKey resource 'MyUsagePlanKey' has a problem: "name": this field cannot be set
The following program fails with: `aws:apigateway/usagePlanKey:UsagePlanKey resource 'MyUsagePlanKey' has a problem: "name": this field cannot be set`

```javascript
const aws = require("@pulumi/aws");
const apiKey = new aws.apigateway.ApiKey("key", {});
const usagePlan = new aws.apigateway.UsagePlan("plan", {});
exports.usagePlanKey = new aws.apigateway.UsagePlanKey('MyUsagePlanKey',{
    keyId: apiKey.id,
    keyType: 'API_KEY',
    usagePlanId: usagePlan.id
});
```

This appears to be due to the AWS provider trying to set the `name` property on the resource (presumably due to AutoName configuration), even though `name` is an output-only property.

lukehoban:
We previously by default applied AutoName to all poperties named `name`, even if they were not input properties.  We now only add AutoName logic if the TF schema for the `name` property is an input.

Also adds a `regression` test case.

Fixes #250.


lukehoban:
Note that an unfortunate consequence of fixing this in the AWS provider instead of `pulumi-terraform` is that we'll need to port this fix to all other TF providers.

PR #267, opened 2018-07-09
Include README in Python Package

ellismg:
Requires https://github.com/pulumi/pulumi-terraform/pull/197 to land and be consumed.


sdk/python/.gitignore
```diff
@@ -0,0 +1,2 @@
+# generated by the build
+README.rst
```
lukehoban: Given that everything in this folder is "generated by the build" - is there any specific need to `.gitignore` this file?  Won't it be unchanged unless there are corresponding README changes?

ellismg: I suppose there is no reason we couldn't check it in like we do for the other generated code.  I'll make that change.

ellismg:
> `pandoc` is a new build requirement, can we document that somewhere? Is there any option here that would avoid the need to introduce the requirement to have `pandoc` to be able to build this package? (it's yet-another friction point for our build)

If `pandoc` is not installed, we'll just generate an empty README.rst file and issue a warning (of course, that warning is lost in the spew of output that is actually just doing a pip build), but you do not need `pandoc` installed to build the package.  It just means if you do and publish that package there won't be a readme displayed on PiPy.

Note that the fact that `README.rst` will have different content depending on if you've installed `pandoc` or not makes me a little wary about checking in the file (I realized this after replying to your previous comment).

PR #268, opened 2018-07-10
Switch package inclusion from whitelist to blacklist.

CyrusNajmabadi:
Related PR: https://github.com/pulumi/pulumi/pull/1612

This changes how we do closure serialization for lambdas.  Previously, we would try to analyzer user source to figure out what packages were being 'required', and we used that to determine what packages to include.  Now, we just include all dependent packages we find (by examinign package.json) but we blacklist certain packages that we do not want included at runtime.  Specifically, we blacklist anything starting with ```@pulumi``` and we also exclude aws-sdk (since it's automatically available in Lambda thanks to amazon).  

Users can also override this behavior, whitelisting packages if they really want them.  Or blacklisting additional packages (for example, a large package they only need at deployment time and not runtime).


overlays/nodejs/serverless/function.ts
```diff
@@ -254,55 +270,51 @@ interface Package {
...
-            addPackageAndDependenciesToSet(s, child, dep);
+    // addPackageAndDependenciesToSet adds all required dependencies for the requested pkg name from the given root package
+    // into the set.  It will recurse into all dependencies of the package.
+    function addPackageAndDependenciesToSet(s: Set<string>, root: Package, pkg: string) {
+        // Don't process this packages if it was in the set the user wants to exclude.
+
+        // Also, exclude it if it's an @pulumi package.  These packages are intended for deployment
+        // time only and will only bloat up the serialized lambda package.
+        if (excludedPackages.has(pkg) ||
+            pkg.startsWith("@pulumi")) {
```
joeduffy: Does this mean you can't capture config by default?

CyrusNajmabadi:
retest this please

CyrusNajmabadi:
@lukehoban Can you take a look please?  Thanks!

overlays/nodejs/serverless/function.ts
```diff
@@ -88,10 +88,17 @@ export interface FunctionOptions {
...
     /**
-     * The packages relative to the program folder to include in the Lambda upload.  The version of the package
-     * installed in the program folder and it's dependencies will all be included. Default is `[]`.
+     * The packages relative to the program folder to include in the Lambda upload.  The version of
+     * the package installed in the program folder and it's dependencies will all be included.
+     * Default is `[]`.
      */
     includePackages?: string[];
+    /**
+     * The packages relative to the program folder to not in the Lambda upload. This can be used to
```
lukehoban: to not **include** in

overlays/nodejs/serverless/function.ts
```diff
@@ -88,10 +88,17 @@ export interface FunctionOptions {
...
+     * override the default serialization logic that includes all packages referenced by
+     * project.json (except @pulumi packages).
```
lukehoban: Defaults to `[]`.

sdk/nodejs/serverless/function.ts
```diff
@@ -254,55 +270,51 @@ interface Package {
...
+function allFoldersForPackages(includedPackages: Set<string>, excludedPackages: Set<string>): Promise<Set<string>> {
     return new Promise((resolve, reject) => {
-        readPackageTree(path, undefined, (err: any, root: Package) => {
+        readPackageTree(".", undefined, (err: any, root: Package) => {
             if (err) {
                 return reject(err);
             }
+
+            const allPackages = new Set<string>(includedPackages);
+            for (const depName of Object.keys(root.package.dependencies)) {
```
lukehoban: Perhaps a comment on this - since it is really the crux of the new behavior that we root the list based on the explicit `dependencies` of the root package.  (And in particular, don't include `devDependencies` or `peerDependencies`).

sdk/nodejs/serverless/function.ts
```diff
@@ -254,55 +270,51 @@ interface Package {
 
 // allFolders computes the set of package folders that are transitively required by the given list of package
 // dependencies rooted in a package at the provided path.
```
lukehoban: This comment looks like it should be update to reflect the new behaviour.

sdk/nodejs/serverless/function.ts
```diff
@@ -254,55 +270,51 @@ interface Package {
...
-function allFoldersForPackages(path: string, packages: string[]): Promise<Set<string>> {
+function allFoldersForPackages(includedPackages: Set<string>, excludedPackages: Set<string>): Promise<Set<string>> {
     return new Promise((resolve, reject) => {
-        readPackageTree(path, undefined, (err: any, root: Package) => {
+        readPackageTree(".", undefined, (err: any, root: Package) => {
             if (err) {
                 return reject(err);
             }
+
+            const allPackages = new Set<string>(includedPackages);
```
lukehoban: Perhaps just `packages` - since it is not actually "all" of the packages - it's goal is explicitly to pick out a subset of all of the packages that are available under `node_modules`.

sdk/nodejs/serverless/function.ts
```diff
@@ -254,55 +270,51 @@ interface Package {
...
-        console.warn(`Could not include required dependency '${pkg}' in '${filepath.resolve(root.path)}'.`)
-        return;
-    }
-    s.add(child.path);
-    if (child.package.dependencies) {
-        for (let dep of Object.keys(child.package.dependencies)) {
-            addPackageAndDependenciesToSet(s, child, dep);
+    // addPackageAndDependenciesToSet adds all required dependencies for the requested pkg name from the given root package
+    // into the set.  It will recurse into all dependencies of the package.
+    function addPackageAndDependenciesToSet(s: Set<string>, root: Package, pkg: string) {
```
lukehoban: Nit - I'll admit I don't personally like these nested functions - it makes me nervous that these is some shared closure-captured state that I'm not aware of, and makes the contract between functions less "obvious".

lukehoban: Oh - and indeed - this does now subtly have a closure captured dependency on `excludedPackages` even though it's similar dependency on `s` could have been closure captured but was not.  I would prefer to just force these to be explicit dependencies by unnesting the functions.

overlays/nodejs/serverless/function.ts
```diff
@@ -254,55 +270,51 @@ interface Package {
...
+
+        // Also, exclude it if it's an @pulumi package.  These packages are intended for deployment
+        // time only and will only bloat up the serialized lambda package.
+        if (excludedPackages.has(pkg) ||
+            pkg.startsWith("@pulumi")) {
+
+            return;
+        }
+
+        var child = findDependency(root, pkg);
```
lukehoban: This was my bad - but looks like this should be `const`.

sdk/nodejs/ec2/vpcIpv4CidrBlockAssociation.ts
```diff
@@ -0,0 +1,89 @@
+// *** WARNING: this file was generated by the Pulumi Terraform Bridge (tfgen) Tool. ***
```
lukehoban: Huh - I guess I missed this file in the PR that generated these new types?

tests/serverless_functions/index.ts
```diff
@@ -10,15 +11,13 @@ function getContentType() {
...
 }, async (ev, ctx, cb) => {
   var aws = await import('aws-sdk');
   var express = await import('express');
   var os = require('os');
   var slack = require('@slack/client');
-  // TODO[pulumi/pulumi#463] Its reasonable to expect that `./other` would work here, but it currently does not.  For
-  // now, validate that the approach that does currently work will serialize the dependencies correctly.
-  var answer = require('./bin/other').answer;
+
+  var answer = other.answer;
```
lukehoban: This no longer really tests any of the logic in this repo.  Should we just remove this part of the test case?

overlays/nodejs/serverless/function.ts
joeduffy: Got it, thanks!

sdk/nodejs/ec2/vpcIpv4CidrBlockAssociation.ts
CyrusNajmabadi: Seems like it :)

overlays/nodejs/serverless/function.ts
CyrusNajmabadi: Done.

overlays/nodejs/serverless/function.ts
CyrusNajmabadi: Done.

sdk/nodejs/serverless/function.ts
CyrusNajmabadi: sure

sdk/nodejs/serverless/function.ts
CyrusNajmabadi: done.

sdk/nodejs/serverless/function.ts
CyrusNajmabadi: sure.  the capture of excludedPackages was exactly why i did this.  but i'm happy to just pass it along.

overlays/nodejs/serverless/function.ts
CyrusNajmabadi: hunh... surprised linting didn't say anything...

CyrusNajmabadi:
retest this please

PR #274, opened 2018-07-14
Add notes about generated code to CONTRIBUTING.md

jen20:
Hopefully this should prevent @joeduffy from having to answer the same questions repeatedly!


joeduffy:
Perfect!  Thank you 😄 

@ellismg How do you want to propagate to other repos?

PR #275, opened 2018-07-16
Use "Apache-2.0" for NPM licenses as per SPDX

jen20:
This fixes the following warning from yarn when building:

```
warning package.json: License should be a valid SPDX license expression
```

I'm not 100% sure that this has no negative impact on Python, however.


ellismg:
LGTM.  Thanks @jen20!

PR #276, opened 2018-07-16
Fix Python naming collisions

joeduffy:
Ingest the changes from our TFgen package,
https://github.com/pulumi/pulumi-terraform/commit/f7064bdeddd27ee2a96d3556e8e7c9cc7689a1f5,
so that we stop generating names -- like `lambda` -- that clash with
built-in Python keywords.


pgavlin:
Were the file renames intentional/are they necessary?

joeduffy:
This is all code-gen and this is the primary aim of the change -- to generate a `lambda_` module that can actually be imported in a Python project, since the current `lambda` name is a reserved keyword.

PR #277, opened 2018-07-17
Add Tags typedef as a Node SDK overlay

jen20:
This gives a consistent definition which can be used across components.

### DIFF
--- a/resources.go
+++ b/resources.go
@@ -1197,6 +1197,7 @@ func Provider() tfbridge.ProviderInfo {
 					"description": {
 						Default: managedByPulumi,
 					},
+					"tags": {Type: awsType(awsMod, "Tags")},
 				},
 			},
 			// Route53
--- a/resources.go
+++ b/resources.go
@@ -758,9 +758,14 @@ func Provider() tfbridge.ProviderInfo {
 			"aws_subnet":                       {Tok: awsResource(ec2Mod, "Subnet")},
 			"aws_volume_attachment":            {Tok: awsResource(ec2Mod, "VolumeAttachment")},
 			"aws_vpc_dhcp_options_association": {Tok: awsResource(ec2Mod, "VpcDhcpOptionsAssociation")},
-			"aws_vpc_dhcp_options":             {Tok: awsResource(ec2Mod, "VpcDhcpOptions")},
 			"aws_default_vpc_dhcp_options": {
 				Tok: awsResource(ec2Mod, "DefaultVpcDhcpOptions"),
+				Fields: map[string]*tfbridge.SchemaInfo{
+					"tags": {Type: awsType(awsMod, "Tags")},
+				},
+			},
+			"aws_vpc_dhcp_options": {
+				Tok: awsResource(ec2Mod, "VpcDhcpOptions"),
 				Fields: map[string]*tfbridge.SchemaInfo{
 					"tags": {Type: awsType(awsMod, "Tags")},
 				},
--- a/resources.go
+++ b/resources.go
@@ -989,10 +989,15 @@ func Provider() tfbridge.ProviderInfo {
 					"tags": {Type: awsType(awsMod, "Tags")},
 				},
 			},
-			"aws_lb_listener":                {Tok: awsResource(elbv2Mod, "Listener")},
-			"aws_lb_listener_certificate":    {Tok: awsResource(elbv2Mod, "ListenerCertificate")},
-			"aws_lb_listener_rule":           {Tok: awsResource(elbv2Mod, "ListenerRule")},
-			"aws_lb_target_group":            {Tok: awsResource(elbv2Mod, "TargetGroup")},
+			"aws_lb_listener":             {Tok: awsResource(elbv2Mod, "Listener")},
+			"aws_lb_listener_certificate": {Tok: awsResource(elbv2Mod, "ListenerCertificate")},
+			"aws_lb_listener_rule":        {Tok: awsResource(elbv2Mod, "ListenerRule")},
+			"aws_lb_target_group": {
+				Tok: awsResource(elbv2Mod, "TargetGroup"),
+				Fields: map[string]*tfbridge.SchemaInfo{
+					"tags": {Type: awsType(awsMod, "Tags")},
+				},
+			},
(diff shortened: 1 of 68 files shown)


overlays/nodejs/tags.ts
```diff
@@ -0,0 +1,21 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Tags represents a set of key-value string pairs to which can be applied
+ * to an AWS resource.
+ */
+export interface Tags {
```
lukehoban: It's a little odd to define a type in this library which is never used.  

We could in principle override the type of properties that accept this shape.  Are there a few we should apply this type to as part of the initial PR?

jen20: Good point, it hadn't occurred to me to go through and replace the usages. I've made a start on doing so, with each resource as a separate commit. If you'd rather they were a single commit for all resources it should be easy enough to squash them.

jen20: I think all the resources are now done, but there is still some work to do to make the data sources work.

jen20:
Right, this is an interesting one and was the least clear when I did the work here. I agree it is definitely a breaking change, but I'm not sure the fix here is to change it to `Tags[]`.

The need to have `tagsCollection` at all comes from [here][1] in the Terraform provider source code, where both `tag` (singular) and `tags` (plural) are specified, but marked as conflicting with one another. In the original Terraform provider, this was done (if memory serves correctly) such that either:

```hcl
resource "aws_autoscaling_group" "some_group" {
    // some options

    tag {
        key = "asgRoute53:record"
        value = "${var.vpc_zone_id}:${var.dns_name}"
        propagate_at_launch = false
    }

    tag {
        key = "system:log_group"
        value = "${var.log_group_name}"
        propagate_at_launch = true
    }
}
```

Or:

```hcl
resource "aws_autoscaling_group" "some_group" {
    // some options

    tags {
        "asgRoute53:record" = "${var.vpc_zone_id}:${var.dns_name}"
        "system:log_group"  = "${var.log_group_name}"
    }
}
```

Would both be valid and usable. The latter of these was originally the only option supported by Terraform, and was likely maintained for backwards compatibility when [terraform#1319][2] was merged.

The implementation of CRUD in the provider, in the knowledge that both `tag` and `tags` cannot both be set, constructs the `createOpts.Tags` value either from the list on the schema (`tag`) or the map (`tags`).

I'm not sure that Pulumi needs to maintain this legacy - since TypeScript has first-class support for maps where the values are complex objects, it would be better to introduce a _new_ type here and go back to having a single `tags` property which can accept either type: something like:

```ts
interface AutoScalingTagValue {
	value: string;
	propagateAtLaunch: true;
}

// We assume that `propagateAtLaunch` is false if a string value is provided here
type AutoScalingTag = string | AutoScalingTagValue;

interface AutoScalingTags {
	[name: string]: AutoScalingTag;
}
```

We could then define `tags` on `aws.autoscaling.Group` as being of type `AutoScalingTagValue`, and remove the legacy of having both `tag` and `tags` with slightly different semantics.

That said, I'm not (yet) sure that this is possible to implement simply using the overlay system in `tfgen`, and in any case it would be a breaking change, though one I think would be beneficial for reducing confusion as an API consumer. If this is the route we want to look at, I'll remove the change from this pull request and follow up with this change separately.


[1]: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_autoscaling_group.go#L270-L277
[2]: https://github.com/hashicorp/terraform/pull/1319
[3]: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_autoscaling_group.go#L411-L427

lukehoban:
> That said, I'm not (yet) sure that this is possible to implement simply using the overlay system in tfgen, and in any case it would be a breaking change, though one I think would be beneficial for reducing confusion as an API consumer. If this is the route we want to look at, I'll remove the change from this pull request and follow up with this change separately.

Thanks for the background on this.  The ideal you describe would likely require new capabilities in `pulumi-terraform`, and seems a good direction to go over time.  But would be great to get this in in current form first.

For right now, I'm not so worried about a breaking change here as I am about the projected shape working.  I believe to ensure this continues to work after the proposed changes, we'll want to add `MaxItemsOne: boolRef(true),` to the `tagsCollection` field.  That will ensure that it gets projected as a singleton map instead of an array of maps.  That is needed so that at runtime the conversion between Terraform and Pulumi converts this singleton to array back and forth correctly.  Without that, the TypeScript types won't match the Terraform object shapes. (Note that everywhere that we use `MaxItemsOne` here, the Terraform provider probably intended to apply `MaxItems: 1` in it's schema (most cases we've used this the Terraform provider implementation doesn't correctly support more than one item).

I think with that one tweak this should be ready to merge.

jen20:
Thanks for the pointer, @lukehoban. I've made the change you recommended and it appears to now be correct (and changed in the Go and Python SDKs as a result).

Regarding `MaxItems: 1` in Terraform - again that is something that was introduced fairly late on, and was never back ported to older resources. This disproportionately affects core AWS resources since they were implemented early on in Terraform's life. Generally, parameters with `MaxItems: 1` set are to give a "block"-style syntax for HCL  which has the unfortunate side effect of actually ending up as an array underneath.

lukehoban:
> Regarding MaxItems: 1 in Terraform - again that is something that was introduced fairly late on, and was never back ported to older resources. This disproportionately affects core AWS resources since they were implemented early on in Terraform's life. Generally, parameters with MaxItems: 1 set are to give a "block"-style syntax for HCL which has the unfortunate side effect of actually ending up as an array underneath.

Yeah - that’s why we project `MaxItems: 1` as a scalar instead of an array, but because some (especially older) resources don’t have that annotation, we also allow force overriding it in the Pulumi projection layer.  At some point it would likely make sense to upstream those annotations to the Terraform providers (I think in all cases so far, that would just allowing catching an existing but earlier in the provider).

lukehoban:
> I've made the change you recommended and it appears to now be correct (and changed in the Go and Python SDKs as a result).

Great - this looks good.

PR #282, opened 2018-07-22
Improve the pulumi/aws experience when building locally.

CyrusNajmabadi:


Makefile
```diff
@@ -27,8 +27,10 @@ build::
 	done
 	cd ${PACKDIR}/nodejs/ && \
 		yarn install && \
+		# pick up the local versions of these if possible.
+		(yarn link @pulumi/pulumi || true) && \
```
CyrusNajmabadi: when building a library, we attempt to pick up any locally built dependencies if possible.  Note the ```|| true``` bit.  if we aren't able to, that's not an issue.  we'll then just build against the version of the lib provided by ```yarn install```.

Makefile
```diff
@@ -27,8 +27,10 @@ build::
...
 		yarn run tsc && \
-		cp ../../README.md ../../LICENSE package.json yarn.lock ./bin/ && \
```
CyrusNajmabadi: it's unclear to me if we should be including yarn.lock's in our libraries.  yarn.lock should be locking your dependencies at the application level.  however, libraries should just say what their constraints are in their package.json.  Then npm/yarn should figure out what to do at the application level with all of this.

Makefile
```diff
@@ -52,8 +54,8 @@ install::
 	mkdir -p "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)"
 	cp -r ${PACKDIR}/nodejs/bin/. "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)"
 	rm -rf "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)/node_modules"
+	cp -r ${PACKDIR}/nodejs/node_modules "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)"
```
CyrusNajmabadi: after blowing away the previous node modules, we copy over the existing ones we installed/linked for this module when building.  

This is necessary, otherwise we break transitive updates.  i.e. consider hte case of pulumi/pulumi, pulumi/aws, and pulumi/cloud.   With the previous logic, even if we linked ```@pulumi/pulumi``` into ```@pulumi/aws```, once we build ```@pulumi/aws```, we'd toss its node_modules, and then do a yarn install, giving us the modules it was writtne against, not what we'd linked locally.

If we then linked @pulumi/pulumi and @pulumi/aws into @pulumi/cloud, we'd be able to observe different @pulumi/pulumi libs.

Makefile
```diff
@@ -52,8 +54,8 @@ install::
...
 	cd "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)" && \
-		yarn install --offline --production && \
```
CyrusNajmabadi: We don't do this since it will install the set of dependencies listed in package.json, ignoring any linking you've done.

Note: if we *do* want tis behavior (perhaps on travis runs) we could consider a new make target (or something else) to get this done.

ellismg:
I understand the direction and agree we need this safety valve to get back on a link plan for cases like you where you work across repositories.

I'm concerned about going all in on linking however.  The `yarn link @pulumi/pulumi || true` thing feels like spooky action at a distance.  If I happened to have built `@pulumi/pulumi` on my machine at any point, I fall onto the link plan.  If my version of `@pulumi/pulumi` starts lagging behind the "official" versions, then I'm likely to end up in a weird place (stuff is missing, for example) and I think it will be hard to determine why.

I also have a little bit of worry around how this interacts with our tests, where we'd have both installed and linked versions of dependent packages floating around (since the tests today list `@pulumi/pulumi` and friends in their package.json's).

But looking at this made me realize that perhaps we can do something like this:

https://github.com/pulumi/pulumi-aws/commit/55a61deff12eb259dd347fb5183c508ac3c4effe

Here, we allow you to set `PULUMI_LINK_PACKAGES=1` before doing a build.  When it is set, we do `yarn link` for all the `@pulumi/*` packages in your project.json before invoking `tsc`.  In addition, we continue to install as normal, but when `PULUMI_LINK_PACKAGES=1` is set, we do a link again after installing to wack all the `@pulumi/*` packages you had downloaded as part of `yarn install`.  This feels a little cleaner to me.  If this is the direction we want to go, I imagine we'd teach the integration testing framework about `PULUMI_LINK_PACKAGES` as well and have it do `yarn link` for all the dependent packages after `yarn install`.

Overall this makes me feel a little better and I think it still meets your needs.  Do you think this is worthwhile?  Will it work for you?

ellismg:
BTW: Thanks a ton for doing the work here.  Reading the diff made me think about the problem in a different light and helped me develop what I linked to above.

CyrusNajmabadi:
@ellismg All of the above makes sense to me.  And i like the idea of this being an environment variable that i can set for local development.  

> When it is set, we do yarn link for all the @pulumi/* packages in your project.json before invoking tsc

That's a lot cleaner than what i was doing.  I far prefer that to how i was hardcoding things.

Note: one niggle here is that the ```|| true``` is necessary (at least for anything not in the local repo).  Or, we can not have that, but it will require everyone to have built these dependent libraries locally.  Otherwise, these link calls will fail.

> we do a link again after installing to wack all the @pulumi/* packages you had downloaded as part of yarn install.

I wasn't certain if that would work.  But, if it does, that's totally fine with me.  :)

> Overall this makes me feel a little better and I think it still meets your needs. Do you think this is worthwhile? Will it work for you?

1000% yes.

My approach was not really intended to be shippable.  But just to show that at least it's *possible* to do local dev as long as the linking stuff isn't totally wrecked during intermediary steps.  However we can accomplish that, while also making sure we don't introduce anything bad during normal/CI builds is great for me.

CyrusNajmabadi:
closing @ellismg has an approach for this.

PR #283, opened 2018-07-24
Downstream impact of module capture branch.

CyrusNajmabadi:


examples/minimal/package.json
```diff
@@ -7,7 +7,7 @@
         "build": "tsc"
     },
     "dependencies": {
-        "@pulumi/pulumi": "^0.14.0-rc1"
+        "@pulumi/pulumi": "^0.14.4-dev-1532134746-g61324d55"
```
CyrusNajmabadi: @ellismg Having to update this in basically every downstream project is really a PITA.  Note that this is necessary so that things like tests actually pick up the real published upstream package.  If we don't do this, tests won't necessarily actually be using the upstream package we're trying to test out.  

Note: this gets even more obnoxious every downstream package you want to update.  i.e. in the `@pulumi/aws-serverless` repo, i'll have to do all of this for both the `@pulumi/pulumi` *and* `@pulumi/aws` packages.  as you go further downstream, the worse it gets.

Effectively, i feel like we really need a way/flag to say: when i'm in a feature-branch, then use that feature-branch name across repos to coordinate which packages are pulled in.  Any `@pulumi` deps should pull the latest version of that feature-branch.  Perhaps that could be done with tags of some sort.  

But it really needs to be some system where we don't have to manually make these sorts of changes (and then undo them once we're satisfied things are working).

CyrusNajmabadi:
Closing out.  @ellismg is finding ways to help us effectively test these things.

PR #286, opened 2018-08-01
Adopt pulumi-terraform#217

lukehoban:
Bring in the changes from pulumi/pulumi-terraform#217.


pgavlin:
> So I think the only test we could add is that the previous way fails, but not sure that adds much value.

I wasn't thinking of the example as a test per se--just an example of how to use data sources. But maybe that's obvious :)

PR #287, opened 2018-08-01
Update serverless example to use new module capturing semantics.

CyrusNajmabadi:


CyrusNajmabadi:
NOte: this will fail until https://github.com/pulumi/pulumi/pull/1685 goes through.

tests/serverless_functions/index.ts
```diff
@@ -2,20 +2,26 @@
...
 const testFunc = new aws.serverless.Function("f", {
   policies: [aws.iam.AWSLambdaFullAccess],
   includePaths: ['./Pulumi.yaml'],
 }, async (ev, ctx, cb) => {
-  var aws = await import('aws-sdk');
-  var express = await import('express');
-  var os = require('os');
-  var slack = require('@slack/client');
+  // These variables exist only to ensure that capturing modules doesn't cause any problems.
+  var _aws = aws;
```
lukehoban: This looks like it's capturing `@pulumi/aws` - that won't work - right?

PR #294, opened 2018-08-06
Use core helper exposed from @pulumi/pulumi.

CyrusNajmabadi:


sdk/nodejs/apigateway/vpcLink.ts
```diff
@@ -75,7 +75,7 @@ export interface VpcLinkState {
     /**
      * The list of network load balancer arns in the VPC targeted by the VPC link. Currently AWS only supports 1 target.
      */
-    readonly targetArn?: pulumi.Input<string>;
+    readonly targetArn?: pulumi.Input<pulumi.Input<string>>;
```
CyrusNajmabadi: @lukehoban this doesn't look like a good change.  Did something change in our tf layer that is causing this?

CyrusNajmabadi: I think this is: https://github.com/pulumi/pulumi-terraform/pull/217

lukehoban:
I suspect you did not `make ensure` before `make`.

PR #296, opened 2018-08-07
Explicilty map tag => tags on aws.autoscaling.Group

### LINKED ISSUE #293: Issue using tags: tag.0: expected object, got slice
Issue report for pulumi-community slack channel:

------

I found another fun issue today with `aws.autoscaling.Group`. The following property value:

```
tags: [
    {
        key: "Name",
        value: "Server",
        propagateAtLaunch: true,
    },
],
```

gives an error when updating:

```
Diagnostics:
  global: global
    error: aws:autoscaling/group:Group resource 'autoscaling-group' has a problem: tag.0: expected object, got slice

  pulumi:pulumi:Stack: pulumi-test-dev
    error: One or more resource validation errors occurred; refusing to proceed

```

lukehoban:
We have an awkward mapping of tag=>tags and tags=>tagsCollection on `aws.autoscaling.Group`.  This leads to some confusion in our tfbridge mapping.

For now, we can fix #293 by making the tag=>tags part of the mapping explicit.  Longer term, we likely need to make deeper improvements in tfbridge to handle this and/or come up with a different approach for aws.autoscaling.Group projection of tag/tags.

Fixes #293.

### DIFF
--- a/resources.go
+++ b/resources.go
@@ -380,6 +380,10 @@ func Provider() tfbridge.ProviderInfo {
 			"aws_autoscaling_group": {
 				Tok: awsResource(autoscalingMod, "Group"),
 				Fields: map[string]*tfbridge.SchemaInfo{
+					"tag": {
+						// Explicitly map tag => tags to avoid confusion with tags => tagsCollection below.
+						Name: "tags",
+					},
 					"tags": {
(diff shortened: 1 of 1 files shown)


jen20:
I can confirm that this fixes the reproduction of the issue I had from #293! Thanks!

PR #297, opened 2018-08-07
Ingest the latest pulumi-terraform.

pgavlin:
These changes bring in support for first-class provider definitions in
Node.JS.


resources.go
```diff
@@ -1780,7 +1780,7 @@ func Provider() tfbridge.ProviderInfo {
 		},
 		JavaScript: &tfbridge.JavaScriptInfo{
 			Dependencies: map[string]string{
-				"@pulumi/pulumi":    "^0.14.0",
+				"@pulumi/pulumi":    "^0.14.4-dev",
```
lukehoban: Does this work?

pgavlin: It does locally.

pgavlin:
> Can we add a test? Ideally of multi-region deployment?

Already working on it.

examples/examples_test.go
```diff
@@ -37,6 +37,7 @@ func TestExamples(t *testing.T) {
 
 	examples := []integration.ProgramTestOptions{
 		baseJS.With(integration.ProgramTestOptions{Dir: path.Join(cwd, "minimal")}),
+		baseJS.With(integration.ProgramTestOptions{Dir: paht.Join(cwd, "multiple-regions")}),
```
pgavlin: I've added this to the "short" CI list for this PR. Happy to move it out before merging.

examples/multiple-regions/index.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright 2016-2017, Pulumi Corporation.  All rights reserved.
+
+import * as aws from "@pulumi/aws";
+import { Output } from "@pulumi/pulumi";
+import { getLinuxAMI } from "./linuxAmi";
+
+let size: aws.ec2.InstanceType = "t2.micro";
+
+function createServer(region: aws.Region): aws.ec2.Instance {
+    const provider = new aws.Provider(`provider-${region}`, { region: region });
```
swgillespie: This is awesome.

examples/multiple-regions/linuxAmi.ts
```diff
@@ -0,0 +1,161 @@
...
+    "cc2.8xlarge": "HVM64",
+};
+
+/**
+ * regionArchLinuxAMI is a map from region to inner maps from architecture to the recommended Linux AMI.
+ */
+export let regionArchLinuxAMI: {
+    [region: string]: { [arch: string]: string; },
+} = {
+    "us-east-1": {
```
jen20: Is there a reason to do hard code AMIs here rather than use `getImage` and friends?

pgavlin: Laziness on my part. :)

Are you referring to https://pulumi.io/reference/pkg/nodejs/@pulumi/aws/index.html#getAmiIds?

jen20: That's the one! That function exists from Terraform partly because it was boring to go update the docs periodically when the AMIs rolled out of availability...

lukehoban: FWIW - also slightly interesting related to this:

- https://aws.amazon.com/blogs/compute/query-for-the-latest-amazon-linux-ami-ids-using-aws-systems-manager-parameter-store/
- https://www.terraform.io/docs/providers/aws/d/ssm_parameter.html

pgavlin: Ended up using the parameter store API.

PR #298, opened 2018-08-07
Update dependencies

ellismg:
This will need to be harmonized with Pat's in-flight changes that do something similar, but opening now to get an early read.


lukehoban:
I think this is out of date - @ellismg can we close?

PR #303, opened 2018-08-21
Update to terraform-provider-aws v1.32.0

lukehoban:


sdk/nodejs/ec2/subnet.ts
```diff
@@ -101,6 +108,11 @@ export class Subnet extends pulumi.CustomResource {
  * Input properties used for looking up and filtering Subnet resources.
  */
 export interface SubnetState {
+    /**
+     * The ARN of the subnet.
+     * * `availability_zone`- The AZ for the subnet.
```
pgavlin: Bit of an odd comment here... I wonder if pulumi/pulumi-terraform#235 would help?

lukehoban: Yeah - hopefully it will - though I suspect there is further work needed as well in general to improve this doc text extraction.

PR #304, opened 2018-08-21
Expose 'factory functions' from pulumi-aws

CyrusNajmabadi:
This is just exposing the support we added here: https://github.com/pulumi/pulumi/pull/1804

Also includes an example showing how you could use this with express.


examples/multiple-regions/index.ts
```diff
@@ -16,13 +16,13 @@ function createServer(region: aws.Region): aws.ec2.Instance {
         ingress: [
             { protocol: "tcp", fromPort: 80, toPort: 80, cidrBlocks: ["0.0.0.0/0"] },
         ],
-    }, { provider: provider });
-    
+    }, <pulumi.CustomResourceOptions>{ provider: provider });
```
swgillespie: why do we need these casts now?

CyrusNajmabadi: tagging @pgavlin .  It looks like we split ResourceOptions into ResourceOptions and CustomResourceOptions.  This resource still takes ResourceOptions, but 'provider' is in CustomResourceOptions.  So, without this cast, typescript thinks something fishy is going on.

I'm running into this because i'm pulling in a new @pulumi/pulumi version.  I'm not sure what our expectation is supposed to be here.

joeduffy: How close are we to landing https://github.com/pulumi/pulumi/issues/1802? I think that would have let the code just work as written.

joeduffy: Absent that, I think you need to write

```
{ providers: { aws: provider } }
```

CyrusNajmabadi: This is custom resource, not a component resource.  so the `{ providers: { aws: provider } }` version seems... wrong.

overlays/nodejs/serverless/function.ts
```diff
@@ -106,9 +129,10 @@ export class Function extends pulumi.ComponentResource {
 
     constructor(name: string,
         options: FunctionOptions,
-        func: Handler,
+        func: FuncSignature,
```
lukehoban: This will, I'm nearly certain - lead to much worse intellisense - since TypeScript can't infer the signature for callers.  I'm not a huge fan of losing this.

Should we consider:
1. A different class that inherits from `aws.serverless.Function` that accepts a different argument?
2. Moving the function parameter inside the args bag and allowing either `func: ` or `factory: `?
3. Something else?

CyrusNajmabadi:
I agree with your concerns luke.  I think to approaches can be taken here:

1. roll back and redo.
2. adjust things with a new PR to handle the problems you identified.

Because the issues seem small enough, i'm going to go with '2'.

PR #309, opened 2018-08-29
Remove incorrectly cased file

ellismg:
This is the old name but when we renamed it we did not delete
it. Having two files with the same name but different cases throws git
for a loop on macOS and Windows.


lukehoban:
:pray:

PR #310, opened 2018-08-29
Remove the need for /opt/pulumi

ellismg:


stack72:
@ellismg Can we revisit this? I'd love to get rid of the need for this :)

I presume we would be needing to do this across everything right?

stack72:
Going to close this out for now - this is something we will want to do wholescale when we do it

PR #311, opened 2018-08-29
Use stronger IAM doc types

joeduffy:
This change uses the iam.PolicyDocument type in a few areas as an
optional input, so that users don't need to manually run JSON.stringify.

This is part of pulumi/pulumi-terraform#10.


joeduffy:
The diff is pretty large, I assume because we didn't ingest some of the more recent commits from the Terraform repo yet. The only ones I triggered are those implied by the `resources.go` changes.

PR #314, opened 2018-09-05
Use the codepaths helper function to determine required codepaths to include with an aws lambda.

CyrusNajmabadi:


overlays/nodejs/serverless/function.ts
```diff
@@ -195,7 +195,10 @@ export class Function extends pulumi.ComponentResource {
         });
 
         const codePaths = computeCodePaths(
-            closure, serializedFileNameNoExtension, options.includePaths, options.excludePackages);
+            closure, serializedFileNameNoExtension,
+            options.includePaths || [],
+            options.includePackages || [],
+            options.excludePackages || []);
```
CyrusNajmabadi: note: we had a bug here where we weren't passing along `options.includePackages`.  This is why i seriously dislike optional parameters.  Instead of optional parameters, we shoudl use *required* parameters that allow *undefined* to be passed in.

CyrusNajmabadi:
CLosing out.  We're taking a different approach.

PR #315, opened 2018-09-06
Update pulumi-terraform.

pgavlin:
This brings in support for default config and property values provided
by environment variables.


pgavlin:
> I assume these were previously b info filled in by the provider instead?

Correct.

PR #323, opened 2018-09-15
Add EC2 PlacementStrategy union and overlay

### LINKED ISSUE #288: Add overlaid constants or sum type for PlacementGroup::strategy
As per the AWS docs for [`create-placement-group`](https://docs.aws.amazon.com/cli/latest/reference/ec2/create-placement-group.html), valid values are:
```
--strategy (string)

The placement strategy.
Possible values:
- cluster
- spread
```

These would be useful constants and/or a sum type to overlay such that autocomplete prevents the need for looking up the values every time!

jen20:
This commit adds a new overlay for nodejs, defining a union type of valid strategies for placement groups, and overrides the `strategy` property of the aws.ec2.PlacementGroup type to accept an array of these instead of raw strings.

Fixes #288.

### DIFF
--- a/overlays/nodejs/ec2/placementStrategy.ts
+++ b/overlays/nodejs/ec2/placementStrategy.ts
@@ -1,0 +1,41 @@
+// Copyright 2016-2018, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file contains two categories of exports:
+//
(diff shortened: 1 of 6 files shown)


overlays/nodejs/ec2/placementStrategy.ts
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+// This file contains two categories of exports:
+//
+//     1) A union type, PlacementStrategy, that accepts any valid Placement Strategy type
+//     2) Individual constants for each such Placement Strategy type
+//
+// These give a better developer experience and are just sugared strings.
+
+export let SpreadStrategy: PlacementStrategy = "spread";
```
joeduffy: I would encourage adding JSDoc comments on these, since it will get picked up by our doc generation process. Honestly, I usually just copy the existing CloudFormation documentation, since it's usually decent, but in this case I guess we need to go to the CLI docs (since CloudFormation docs aren't very good here).

```
/**
 *  A `spread` placement group places instances on distinct hardware.
 */
```

from https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreatePlacementGroup.html.

jen20: I've added the comments to the constants and the union, and rebased this on top of `master`.

jen20:
:+1:, forgot about the need for re-generating here. Looks like the comments made it through into the relevant files in the `sdk` directory now.

PR #324, opened 2018-09-15
Allow API objects in addition to IDs for several properties

### LINKED ISSUE #289: Allow LaunchConfiguration::instanceProfile to accept an InstanceProfile
It's often the case that an instance profile and launch configuration are defined close to each other, such that the instance profile object is still in scope. It should be allowed to pass the object into LaunchConfigurationArgs as well as the ID of the resulting instance profile.

This may also apply to `aws.ec2.Instance` as well as `aws.ec2.LaunchConfiguration`.

### LINKED ISSUE #290: Allow aws.autoscaling.Group::placementGroup to refer to an aws.ec2.PlacementGroup
This is another case where passing in an object could be preferable to referring to the object by ID, similar to #289.

### LINKED ISSUE #292: Allow aws.autoscaling.Group::launchConfiguration to refer to an aws.ec2.LaunchConfiguration
This is another case where passing in an object could be preferable to referring to the object by ID, similar to #289.

jen20:
This pull request allows various properties in AWS resources related to autoscaling and IAM to accept relevant API objects in addition to the string representation of their ID.

- `launchConfiguration` on `aws.autoscaling.Group` will accept an `aws.ec2.LaunchConfiguration` (Fixes #292)
- `placementGroup` on `aws.autoscaling.Group` will accept an `aws.ec2.PlacementGroup` (Fixes #290)
- `iamInstanceProfile` on `aws.ec2.Instance` and `aws.ec2.LaunchConfiguration` will accept an `aws.iam.InstanceProfile` (Fixes #289)

### DIFF
--- a/resources.go
+++ b/resources.go
@@ -387,6 +387,10 @@ func Provider() tfbridge.ProviderInfo {
 			"aws_autoscaling_group": {
 				Tok: awsResource(autoscalingMod, "Group"),
 				Fields: map[string]*tfbridge.SchemaInfo{
+					"launch_configuration": {
+						Type:     "string",
+						AltTypes: []tokens.Type{awsResource(ec2Mod, "LaunchConfiguration")},
+					},
 					"placement_group": {
 						Type:     "string",
 						AltTypes: []tokens.Type{awsResource(ec2Mod, "PlacementGroup")},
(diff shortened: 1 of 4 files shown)


joeduffy:
:+1: :100:

PR #327, opened 2018-09-19
Move 'aws-serverless' components down into 'aws'

CyrusNajmabadi:
There are many components to this change.  So it's worth outlining all of them:

1. Moving 'EventSubscription' down to aws.lambda.EventSubscription.
2. Move 'createLambdaFunction' down to aws.lambda.createLambdaFunction.  Note: this should perhaps just be called .createFunction.
3. Deprecate aws.serverless.Function (and all corresponding types it exports), with documentation saying to use the corresponding functions and types in aws.lambda instead.
4. Make the helpers/types surrounding creating aws.lambda.Functions strongly typed.  
5. Move Bucket-Subscription code down from aws-serverless into aws.s3.  
6. Use mixins to add that subscription functionality onto aws.Bucket itself.

All of the above has been done in a backward compatible fashion. i.e. we have deprecated, but kept around, all existing types/functions/classes.  Though, to ensure we don't have massive duplication, the old code sits on top of the new code now.


examples/bucket/index.ts
```diff
@@ -0,0 +1,54 @@
...
+        rule: {
+            applyServerSideEncryptionByDefault: {
+                sseAlgorithm: "AES256",
+            },
+        },
+    },
+    forceDestroy: true,
+});
+
+bucket.onObjectCreated("test", bucket, async (event) => {
```
CyrusNajmabadi: Shows using an mixin method on aws.s3.Bucket directly.

overlays/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,308 @@
...
+    readonly functionVersion: string;
+    readonly invokedFunctionArn: string;
+    readonly memoryLimitInMB: string;
+    readonly awsRequestId: string;
+    readonly logGroupName: string;
+    readonly logStreamName: string;
+    readonly identity: any;
+    readonly clientContext: any;
+    getRemainingTimeInMillis(): string;
+}
```
CyrusNajmabadi: these types moved over from aws.serverless.  IMO, we should not have/need aws.serverless (except for back-compat purposes).  All functionality can and should be more cleanly exposed from aws.lambda (with mixins if necessary).

For back compat, the orignal types remain, but forward to these types.

overlays/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,308 @@
...
+ */
+export type EntryPoint<E, R> = (event: E, context: Context, callback: (error: any, result: R) => void) => Promise<R> | void;
+
+/**
+ * EntryPointFactory is the signature for a function that will be called once to produce the
+ * serverless function that AWS Lambda will invoke.  It can be used to initialize expensive state
+ * once that can then be used across all invocations of the Lambda (as long as the Lambda is using
+ * the same warm node instance).
+ */
+export type EntryPointFactory<E, R> = () => EntryPoint<E, R>;
```
CyrusNajmabadi: these types were renamed as generic names like Func/Handler were getting quite confusing.  "EntryPoint" clearly indicates (along with teh docs) that these represent the initial entrypoints for a Lambda.  

These have also been made strongly typed, as we had a weak typed and strong typed version of these before.  Now, we'll just have the strong typed version.  If someone wants the weak typed version they can use `Entrypoint<any, any>`

overlays/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,308 @@
...
+    /**
+     * The Javascript function instance that will be called to produce the function that is the
+     * entrypoint for the AWS Lambda. Either [func] or [factoryFunc] must be provided.
+     *
+     * This form is useful when there is expensive initialization work that should only be executed
+     * once.  The factory-function will be invoked once when the final AWS Lambda module is loaded.
+     * It can run whatever code it needs, and will end by returning the actual function that Lambda
+     * will call into each time the Lambda is invoked.
+     */
+    factoryFunc?: EntryPointFactory<E, R>;
```
CyrusNajmabadi: Note: i would like to rename these properties o `entryPoint` and `entryPointFactory`.  however, with backcompat needs, that would mean still looking for these properties as well.  it seemed like too much effort, and 'func/factoryFunc' are acceptable.

overlays/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,308 @@
...
+     * Default is `[]`.
+     */
+    includePackages?: string[];
+
+    /**
+     * The packages relative to the program folder to not include the Lambda upload. This can be
+     * used to override the default serialization logic that includes all packages referenced by
+     * project.json (except @pulumi packages).  Default is `[]`.
+     */
+    excludePackages?: string[];
```
CyrusNajmabadi: Note: i would like to take includePaths/includePackage/excludePackages and make a new CodePathOptions type for them.  That way, if we need to ever add/change options to pulumi.computeCodePaths, we can just update that type (and reference it here), instead of needing to downstream flow all those changes.

overlays/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,308 @@
...
+ * Creates an AWS Lambda from a provided Handler.  If the Handler is already an AWS Lambda, this
+ * does nothing and simply returns that instance.  Otherwise, if it is a JavaScript Function
+ * instance, this will serialize that Function instance into a usable packaged form that is then
+ * exposed through an AWS Lambda.
+ *
+ * If no [policies] are included in [functionOptions] a default policy of
+ * [aws.iam.AWSLambdaFullAccess] will be used.
+ */
+export function createLambdaFunction<E, R>(
+    name: string, options: FunctionOptions<E, R>, opts?: pulumi.ResourceOptions): lambdaFunction.Function {
```
CyrusNajmabadi: Note: this majority of this code was pulled out of aws.serverless.Function.  That type now just forwards to this helper.

overlays/nodejs/s3/s3Mixins.ts
```diff
@@ -0,0 +1,231 @@
...
+ * Arguments to help customize a notification subscription for a bucket.
+ */
+export interface CommonBucketSubscriptionArgs {
+    /**
+     * An optional prefix or suffix to filter down notifications.  See
+     * aws.s3.BucketNotification.lambdaFunctions for more details.
+     */
+    filterPrefix?: string;
+    filterSuffix?: string;
+}
```
CyrusNajmabadi: All this code moved down from aws-serverless.  However, it was tweaked to point at aws.lambda, not aws.serverless at this layer.

joeduffy:
Cool -- definitely eager to hear what @jen20 thinks!

overlays/nodejs/lambda/lambdaMixins.ts
CyrusNajmabadi: Note: i woudl definitely recommend understanding the types in this file as they are criticla to understanding the overall Lambda/Serverless/Mixin shape we're exposing at this layer.

overlays/nodejs/s3/s3Mixins.ts
```diff
@@ -0,0 +1,232 @@
...
+            key: string;
+            size: number;
+            eTag: string;
+            versionId?: string;
+            sequencer: string;
+        };
+    };
+}
+
+export type BucketEventHandler = lambda.EventHandler<BucketEvent, void>;
```
CyrusNajmabadi: By being a lambda.EventHandler, this is saying: you can either provide a JS function that we will serialize for you, or you can just provide the actual aws.lambda you want to be called.

overlays/nodejs/s3/s3Mixins.ts
```diff
@@ -0,0 +1,232 @@
...
+                events: subscription.events,
+                filterPrefix: subscription.filterPrefix,
+                filterSuffix: subscription.filterSuffix,
+                lambdaFunctionArn: subscription.lambdaFunctionArn,
+            })),
+        }, { parent: bucket, dependsOn: permissions });
+    }
+});
+
+// Mixin event handling functionality onto Bucket.
```
CyrusNajmabadi: this part is where this gets a bit interesting.  Below is the TS/JS way to effectively have 'partial types'.  Here we augment the shape of hte 'Bucket' type by stating that it has these three extra subscription methods.  Then below that we actually provide the implementation of those methods by directly updating the .prototype of the Bucket class.

overlays/nodejs/serverless/function.ts
```diff
@@ -12,41 +12,27 @@
...
-    readonly invokedFunctionArn: string;
-    readonly memoryLimitInMB: string;
-    readonly awsRequestId: string;
-    readonly logGroupName: string;
-    readonly logStreamName: string;
-    readonly identity: any;
-    readonly clientContext: any;
-    getRemainingTimeInMillis(): string;
-}
+export type Context = lambda.Context;
```
CyrusNajmabadi: all this code is effectively deprecated in favor of the corresponding types in aws.lambda.XXX.  However, for back compat, htey have been marked deprecated, but kept here.

CyrusNajmabadi:
Note: i've only moved one set of subscriptions down as i'd like to get consensus on what people see here first.  Once i do, i can move the rest of https://github.com/pulumi/pulumi-aws-serverless/pull/31 down into this (or do it in an immediately following PR).

overlays/nodejs/s3/s3Mixins.ts
```diff
@@ -0,0 +1,231 @@
...
+declare module "./bucket" {
+    interface Bucket {
+        onObjectCreated(
+            name: string, handler: BucketEventHandler,
+            args?: ObjectCreatedSubscriptionArgs, opts?: pulumi.ResourceOptions): BucketEventSubscription;
+
+        onObjectRemoved(
+            name: string, handler: BucketEventHandler,
+            args?: ObjectRemovedSubscriptionArgs, opts?: pulumi.ResourceOptions): BucketEventSubscription;
```
CyrusNajmabadi: the above names are an opinionated change than those cloud/serverless used to use.  Instead of onPut/onDelete, we use the names that AWS uses for these operations.  i.e. when you register for these events you use the "s3:ObjectCreated" tag.  

This fits into our view that @pulumi/aws should be feel very aws native.

overlays/nodejs/s3/s3Mixins.ts
```diff
@@ -0,0 +1,231 @@
...
+
+        /**
+         * Creates a new subscription to the given bucket using the lambda provided, along with
+         * optional options to control the behavior of the subscription.  This function should be
+         * used when full control over the subscription is wanted, and other helpers (like
+         * onObjectCreated/onObjectRemoved) are not sufficient.
+         */
+        onEvent(
+            name: string, handler: BucketEventHandler,
+            args: BucketEventSubscriptionArgs, opts?: pulumi.ResourceOptions): BucketEventSubscription;
```
CyrusNajmabadi: this is also a part of our opinionated view on events.  In general, most resources will have a generic onEvent entrypoint that gives them full control over whatever AWS exposes.  For example, you may want to specifically list to ObjectCreated:Post vs ObjectCreated:Put.  Or you might want a single event that hears about both creation and deletes.  These onEvent helpers try to expose the full subscription functionality.  

In general, we will try to provide other suitable helpers (like onObjectCreated and onObjectDeleted) when AWS has high level event bucketing themselve.  These helpers should almost always call into the more generic onEvent entrypoint, just with the right values set for the user.

CyrusNajmabadi: Fundamentally though, that means that all these helpers may look very different across cloud providers.  The Azure versions here will have a very different naming/shape.  Unification (if any) happens at the @pulumi/cloud layer.

CyrusNajmabadi:
@lukehoban You had an open question about how things like 'Roles' would be parented in this new formalization (where we deprecate aws.serverless.Function).  THere are two answers, depending on which part of the API you call into.

1. If you call directly into `aws.lambda.createFunction(name, handler, opts)` then the Role created will be parented by whatever 'parent' is set by 'opts'.  This is somewhat unsatisfying if opts is not provided, as that means it will be parented by the stack.  *However*, this is alleviated by:
2. If you are creating the lambda for something like an event-subscription (like a BucketEventSubcription).  IN this case the subscription itself is the parent resource as it is the component that logically creates and owns the Lambda and connects it to hte bucket.  You can see that in this representation:

```
     Type                                Name                          Status      Info
 +   pulumi:pulumi:Stack                 bucket-cybucketawstest        created     1 warning
 +   ├─ aws:s3:Bucket                    testbucket                    created     
 +   │  └─ aws:s3:BucketNotification     test-object-created           created     
 +   └─ aws:s3:BucketEventSubscription   test-object-created           created     
 +      ├─ aws:iam:Role                  test-object-created           created     
 +      ├─ aws:lambda:Function           test-object-created           created     
 +      ├─ aws:iam:RolePolicyAttachment  test-object-created-32be53a2  created     
 +      └─ aws:lambda:Permission         test-object-created           created
```

In general, our advice should be that those component that actually end up causing the lambda to be created should pass a reasonable parent for that lambda (often themselves).

overlays/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,321 @@
...
+
+    /**
+     * The packages relative to the program folder to not include the Lambda upload. This can be
+     * used to override the default serialization logic that includes all packages referenced by
+     * project.json (except @pulumi packages).  Default is `[]`.
+     */
+    excludePackages?: string[];
+
+    /**
+     * @internal.  TODO(cyrusn): Remove if not needed by higher layers.
```
swgillespie: Is this still pending? Do we know if it's needed?

overlays/nodejs/s3/s3Mixins.ts
```diff
@@ -0,0 +1,232 @@
...
+            events: args.events,
+            filterPrefix: args.filterPrefix,
+            filterSuffix: args.filterSuffix,
+            lambdaFunctionArn: this.func.arn,
+            permission: permission,
+        });
+    }
+}
+
+process.on("beforeExit", () => {
```
swgillespie: Is this still necessary, you think? Do you know why this has to be run before exit?

CyrusNajmabadi: Yeah, this is necessary.  The main issue here is that AWS requires all bucket subscriptions to be set in one swell foop.  So we have to collect them all over the life of the app, and then only apply right at the end.

overlays/nodejs/lambda/lambdaMixins.ts
CyrusNajmabadi: yeah, i can't see any place where we still use ths.  will remove.

CyrusNajmabadi:
@jen20 @lukehoban Can you guys take a look?  My primary PR concern here is if anything stands out to you as "oh no... i don't like that".  This is less about the core implementation and more about if the design/shapes for the API we'll be exposing feel appropriate.  

On a personal level, i like it a *lot*.  I've been using it both here and in https://github.com/pulumi/pulumi-aws/pull/328 and it feels much better and much more natural.  But i want some separate eyes here before proceeding.  

Note: i would very much like the reviews done today if possible.  That would make it possible for me to do the whole work of moving all of 'serverless' down to 'aws' and to get 'cloud-aws' ported over onto this this weekend. 

Thanks!

examples/bucket/bin/index.js
```diff
@@ -0,0 +1,57 @@
+"use strict";
```
joeduffy: Do we intend to commit the `bin/` directory?

examples/bucket/index.ts
```diff
@@ -0,0 +1,54 @@
...
+        rule: {
+            applyServerSideEncryptionByDefault: {
+                sseAlgorithm: "AES256",
+            },
+        },
+    },
+    forceDestroy: true,
+});
+
+bucket.onObjectCreated("test", async (event) => {
```
joeduffy: I might call this example `bucket-lambda`.

overlays/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,302 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as pulumi from "@pulumi/pulumi";
+import * as crypto from "crypto";
+
+import * as arn from "../arn";
+import * as iam from "../iam";
+import * as runtime from "./runtimes";
```
joeduffy: Nit, strange sorting, I'd expect `runtime` to be next to `lambdaFunction` and `permission` since both are in the current directory.

overlays/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,302 @@
...
+import * as utils from "../utils";
+
+import * as lambdaFunction from "./function";
+import * as permission from "./permission";
+
+/**
+ * Context is the shape of the context object passed to a Function callback.
+ */
+export interface Context {
+    callbackWaitsForEmptyEventLoop: boolean;
```
joeduffy: Nit, wish we had comments for these properties.

overlays/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,302 @@
...
+    readonly functionName: string;
+    readonly functionVersion: string;
+    readonly invokedFunctionArn: string;
+    readonly memoryLimitInMB: string;
+    readonly awsRequestId: string;
+    readonly logGroupName: string;
+    readonly logStreamName: string;
+    readonly identity: any;
+    readonly clientContext: any;
+    getRemainingTimeInMillis(): string;
```
joeduffy: Why not `readonly`?

overlays/nodejs/lambda/lambdaMixins.ts
joeduffy: How about something like `FunctionCallback` or, to make it look better alongside `EventHandler`, perhaps `EventCallback`.

overlays/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,302 @@
...
+export type EntryPointFactory<E, R> = () => EntryPoint<E, R>;
+
+/**
+ * An EventHandler is either a JavaScript Function instance or an aws.lambda.Function that can be
+ * used to handle an event triggered by some resource.  If just a JavaScript function is provided
+ * the AWS Lambda will be created by calling [createFunction] on it.  If more control over the
+ * resultant AWS Lambda is required, clients can call [createFunction] directly and pass the result
+ * of that to any code that needs an EventHandler.
+ */
+export type EventHandler<E, R> = EntryPoint<E, R> | lambdaFunction.Function;
```
joeduffy: I always find it strange when I see `function.Function`. We seem to use that pattern a lot. As part of moving these abstractions down, do you think we can come up with a better scheme?

overlays/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,302 @@
...
+
+
+export interface CodePathOptions {
```
joeduffy: Nit, type-level comment and maybe properties too?

overlays/nodejs/s3/s3Mixins.ts
```diff
@@ -0,0 +1,232 @@
...
+/**
+ * Arguments to help customize a notification subscription for a bucket.
+ */
+export interface CommonBucketSubscriptionArgs {
+    /**
+     * An optional prefix or suffix to filter down notifications.  See
+     * aws.s3.BucketNotification.lambdaFunctions for more details.
+     */
+    filterPrefix?: string;
+    filterSuffix?: string;
```
joeduffy: Feels like we should have independent comments for each.

overlays/nodejs/s3/s3Mixins.ts
```diff
@@ -0,0 +1,232 @@
...
+            principal: "s3.amazonaws.com",
+            sourceArn: bucket.id.apply(bucketName => `arn:aws:s3:::${bucketName}`),
+        }, { parent: this });
+
+        this.permission = permission;
+
+        // We must create only a single BucketNotification per Bucket per AWS API limitations.  See
+        // https://github.com/terraform-providers/terraform-provider-aws/issues/1715.  So we push
+        // the subscription information here, and then actually create the BucketNotification if
+        // needed on process `beforeExit`.
```
joeduffy: Eek, I guess this is ok -- but kind of scary to create resources just before the process exits like this.

overlays/nodejs/s3/s3Mixins.ts
```diff
@@ -0,0 +1,232 @@
...
+    // The type of event to listen for.  '*' is the default if not provided.
+    event?: "*" | "Delete" | "DeleteMarkerCreated";
+}
+
+// See https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html.
+export interface BucketEvent {
+    Records?: BucketRecord[];
+}
+
+export interface BucketRecord {
```
joeduffy: Nit, could use some better comments. I like where we landed on this general topic in https://github.com/pulumi/pulumi-aws/pull/323.

overlays/nodejs/s3/s3Mixins.ts
joeduffy: I agree 100% with this.

overlays/nodejs/s3/s3Mixins.ts
```diff
@@ -0,0 +1,232 @@
...
+}
+
+/**
+ * Arguments to specifically control a subscription to 'ObjectCreated' notifications on a bucket.If
+ * more events than just 'ObjectCreated' events are desired, the 'subscribe' function should be used
+ * instead.
+ */
+export interface ObjectCreatedSubscriptionArgs extends CommonBucketSubscriptionArgs {
+    // The type of event to listen for.  '*' is the default if not provided.
+    event?: "*" | "Put" | "Post" | "Copy" | "CompleteMultipartUpload";
```
joeduffy: Might be nice to document each option e.g. from https://docs.aws.amazon.com/AmazonS3/latest/dev/NotificationHowTo.html.

joeduffy: In fact, any reason not to have an enum of options here? Seems nicer to be able to say

`lambda.ObjectCreatedEvent` versus `"*"`

and similarly `lambda.ObjectCreatedPutEvent` versus `"Put"`, especially given case sensitivity.

overlays/nodejs/s3/s3Mixins.ts
```diff
@@ -0,0 +1,232 @@
...
+    args = args || {};
+    args.event = args.event || "*";
+
+    const argsCopy = {
+        filterPrefix: args.filterPrefix,
+        filterSuffix: args.filterSuffix,
+        events: ["s3:ObjectCreated:" + args.event],
+    };
+
+    return this.onEvent(name + "-object-created", handler, argsCopy, opts);
```
joeduffy: It's not clear we need to append `-object-created`, given that the developer is explicitly naming their registration.

examples/bucket/bin/index.js
CyrusNajmabadi: Nope!  Will fix.

overlays/nodejs/lambda/lambdaMixins.ts
CyrusNajmabadi: Docs added.

overlays/nodejs/lambda/lambdaMixins.ts
CyrusNajmabadi: I really like lambda.EntryPoint.  It makes it really clear waht it's describing. i.e. it's not the shape of an arbitrary feature-level callback.  It's expressing hte shape of hte *Lambda* entrypoint.  It's very lambda specific (since it gets that wacky callback object and has that specific shape).  EntryPointFactory also reads well to me.

CyrusNajmabadi: >  always find it strange when I see function.Function

I was able to improve things here.  The fundamental problem is that aws calls this "Function" which is a very important type/name already in JS.  We don't want to collide with that as it can make things super wrong.  Previously, we did this by qualifying with the module name.  However, a nicer approach is to just import this as "Function as LambdaFunction", so that we can refer to it in code as LambdaFunction which is simpler but still clearly indicates what this is.

overlays/nodejs/s3/s3Mixins.ts
CyrusNajmabadi: indeed.  i was being lazy.

overlays/nodejs/s3/s3Mixins.ts
CyrusNajmabadi: We've been doing this for months.  This is how Bucket.put has had to work since day 1.  It actually hasn't been a problem for us at all :)

overlays/nodejs/s3/s3Mixins.ts
CyrusNajmabadi: A large problem here is trying to find acutal docs to refer to for these guys.  For example, this is the best i could find from amazon themselves: https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html

Note that htey don't really explain what the fields are in their own docs.  It's hard to give better comments here when AWS isn't even really documenting this stuff in any explicit fashion (that i can find at least).

overlays/nodejs/s3/s3Mixins.ts
CyrusNajmabadi: > and similarly lambda.ObjectCreatedPutEvent versus "Put", especially given case sensitivity.

Note: case sensitivity isn't really a problem.  TS will complain if you give hte wrong casing as this code explicitly defines the allowed string values you can provide.  

We can use an enum here though. I don't feeel strongly about that one way or the other.  Let me go look at how that would look.

CyrusNajmabadi: TypeScript will also give you intellisense here, allowing you to complete out to these names easily.  So it's not like having an enum actually makes things easier to type, or easier to check.

CyrusNajmabadi: This form is actually how TS does things with their library files, and is idiomatic for the TS community.  i.e.:  `    normalize(form: "NFC" | "NFD" | "NFKC" | "NFKD"): string;`  There's no need for an enum indirection as the "union" and "string" (or number) types subsume that.

overlays/nodejs/s3/s3Mixins.ts
CyrusNajmabadi: indeed.  holdover from how we did things before.  removing now.

overlays/nodejs/lambda/lambdaMixins.ts
CyrusNajmabadi: I really like lambda.EntryPoint. It makes it really clear waht it's describing. i.e. it's not the shape of an arbitrary feature-level callback. It's expressing hte shape of hte Lambda entrypoint. It's very lambda specific (since it gets that wacky callback object and has that specific shape). EntryPointFactory also reads well to me.

overlays/nodejs/lambda/lambdaMixins.ts
CyrusNajmabadi: This was removed.  it was temporary until i updated this PR to reference hte latest @pulumi/pulumi.

overlays/nodejs/lambda/lambdaMixins.ts
lukehoban: I'll admit I find `EntryPoint` a bit surprising here as well - in part because I've never seen that term used in Lambda docs.  In their docs, this is a "handler" or "callback".  

Note even above where the doc comments say "Context is the shape of the context object passed to a Function callback." and include a property named `callbackWaitsForEmptyEventLoop`.  Given that, I think `FunctionCallback` would feel more natural here.

overlays/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,365 @@
...
+ */
+export type EventHandler<E, R> = EntryPoint<E, R> | LambdaFunction;
+
+/**
+ * FunctionOptions provides configuration options for the serverless Function.  It is effectively
+ * equivalent to [aws.lambda.FunctionArgs] except with a few important differences documented at the
+ * property level.  For example, [role] is an actual iam.Role instance, and not an ARN. Properties
+ * like [runtime] are now optional.  And some properties (like [code]) are entirely disallowed.
+ */
+export type FunctionOptions<E, R> = utils.Overwrite<FunctionArgs, {
```
lukehoban: It seems a little strange now that this module has both `aws.lambda.FunctionArgs` and `aws.lambda.FunctionOptions`.  Should the latter be `aws.lambda.CallbackFunctionArgs`?

overlays/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,365 @@
...
+     * Not allowed when creating an aws.serverless.Function.  The [code] will be generated from the
+     * passed in JavaScript callback.
+     */
+    code?: never;
+
+    /**
+     * Not allowed when creating an aws.serverless.Function.  The [code] will be generated from the
+     * passed in JavaScript callback.
+     */
+    handler?: never;
```
lukehoban: This comment again refers to the `callback` - I feel like I do keep expecting that to be the name of the property here.

Also slightly notable that the one place "entrypoint" is used in AWS docs is in the original doc comment on this `handler` property:

```
     /**
     * The function [entrypoint][3] in your code.
     */
    readonly handler: pulumi.Input<string>;
```

In that context, it's really referring to the name of the export on the JS module (subtle difference).

overlays/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,365 @@
...
+    else {
+        return handler;
+    }
+}
+
+/**
+ * Creates an AWS Lambda out of the provider options.func or options.factoryFunc.  Configuration of
+ * this lambda can be provided through [options].
+ */
+export function createFunction<E, R>(
```
lukehoban: Along the lines of other comments, I do wonder whether this should use some other identifying word - like `createCallbackFunction`?  It's argument bag doesn't allow for creating a Function from normal handler/code data - so it's really disjoint from `new aws.lambda.Function`, not something that encapsulates that plus additional use cases.

Frankly, it's almost like this should just be `new aws.lambda.CallbackFunction()` for full consistency with our normal pattern?  With the design of things as they are in this library - what's the main argument against that?

lukehoban: > Frankly, it's almost like this should just be new aws.lambda.CallbackFunction() for full consistency with our normal pattern? With the design of things as they are in this library - what's the main argument against that?

I guess one question is what does it return - but a subclass of `aws.lambda.Function` seems a natural answer - right?

overlays/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,365 @@
...
+}
+
+// Mixin the Role we potentially create into the Function instances we return.
+
+declare module "./function" {
+    interface Function {
+        // Actual Role instance value for this Function.  Will only be set if this function
+        // was created from [createFunction]
+        roleInstance?: iam.Role;
+    }
```
lukehoban: This feels a little awkward - is this another reason to consider making this a subclass of `aws.lambda.Function`?

overlays/nodejs/s3/s3Mixins.ts
lukehoban: Perhaps just including that doc link in the interface docs would be nice.

overlays/nodejs/s3/s3Mixins.ts
lukehoban: Specifically - a Bucket can only have one BucketNotification resource attached to it (which can itself specify multiple handler).  Since we want to allow these to be provided at mutiple points in time via a more imprative API, we have to collect these and create the BucketNotification at shutdown.

Arguably, this is one of the pieces of evidence suggesting we should be generally delaying more things, to allow other imperative models - but that's something for another day.

sdk/nodejs/lambda/lambdaMixins.ts
```diff
@@ -0,0 +1,365 @@
+// Copyright 2016-2018, Pulumi Corporation.
```
lukehoban: Not for this commit - but the more we add these overlays - the more I feel like we should just move "overlays" into the `sdk` folder next to the generated code instead of generating them from themselves.  That will make authoring more sane, and avoid this strange duplication of files.  Now that we check in generated files, there are few downsides to doing this.

overlays/nodejs/lambda/lambdaMixins.ts
CyrusNajmabadi: Ok.  Based on this the above feedack, i'm going to be naming these to FunctionCallback, FunctionCallbackFactory, and FunctionCallbackArgs

overlays/nodejs/lambda/lambdaMixins.ts
CyrusNajmabadi: That would be an interesting approach.  I like it, let me go try that out!

overlays/nodejs/lambda/lambdaMixins.ts
CyrusNajmabadi: Sure.  That seems very doable.

sdk/nodejs/lambda/lambdaMixins.ts
CyrusNajmabadi: YES.  I 100% agree.  it's super annoying to do things like we do today :)

PR #333, opened 2018-09-22
Move away from needing overlays in the aws layer.

CyrusNajmabadi:


CyrusNajmabadi:
>  How do I now regenerate the entire package from scratch? 

Isn't the entire package always generated from scratch? 

> And what happens when a type is renamed in the TF provider?

That should still work fine.  If any of the mixin types reference those, you'd just need to update after regenerating things.

PR #340, opened 2018-09-26
Move 'API' code down into aws.apigateway.x

CyrusNajmabadi:
This is the last piece of aws-serverless.  Moving this down allows us to completely deprecate that package.
Module is currently called 'x' because i'm not confident this is in the final form we'll want.  There are still pieces of APIGateway that we don't support (like connecting APIGateway to arbitrary AWS services).

We'll want to do a full review to make sure we are exposing all hte functionality/capabilities we want, in a cohesive manner.


CyrusNajmabadi:
@lukehoban This is the move down of the API code.  As per our discussion, i'm putting it in an 'x' module for now.  Review when you can.  I'd like to merge in asap so i can retire this part of aws-serverless, and also resit pulumi/cloud-aws onto this.

CyrusNajmabadi:
Ok.  Going to merge this in.  If you have any changes you want me to make, let me know.

PR #341, opened 2018-09-27
Update terraform-provider-aws to v1.38.0.

pgavlin:


lukehoban:
I don't see any updates to `resources.go` - but I do see a new data source at https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1380-september-26-2018.  Are there new warnings in the build now about these new data sources being unmapped?

In general, we should add support for all newly exposed APIs as part of updating the TF provider.

sdk/nodejs/applicationloadbalancing/listener.ts
```diff
@@ -33,17 +33,17 @@ export class Listener extends pulumi.CustomResource {
     /**
      * An Action block. Action blocks are documented below.
      */
-    public readonly defaultActions: pulumi.Output<{ targetGroupArn: string, type: string }[]>;
+    public readonly defaultAction: pulumi.Output<{ fixedResponse?: { contentType: string, messageBody?: string, statusCode: string }, redirect?: { host?: string, path?: string, port?: string, protocol?: string, query?: string, statusCode: string }, targetGroupArn?: string, type: string }>;
```
lukehoban: Interesting.  This will be a breaking change.  Looks like a case there `MaxItems==1` was added in TF, which is not really breaking, but that causes us to project this differently.  

I don't think there is anything we can/should do here - but worth marking this as breaking and adding a note to the PR.

pgavlin:
> Are there new warnings in the build now about these new data sources being unmapped?

I did not look, to be honest. I will check.

pgavlin:
@lukehoban I've added the missing resources and data sources.

PR #343, opened 2018-10-04
Change to a path-based approach to API routes in the experimental apigateway API.

CyrusNajmabadi:
I recommend reviewing with whitespace off: https://github.com/pulumi/pulumi-aws/pull/343/files?w=1

Instead of passing multiple differnet arrays of route types, callers can just supply a 'route mapping object' where the keys are the route paths they care about, and the values are the specific types of routes.

### DIFF
--- a/examples/api/index.ts
+++ b/examples/api/index.ts
@@ -44,23 +44,20 @@ const lambda = new aws.lambda.Function("myfunction", {
 });
 
 const api = new x.API("myapi", {
-    routes: {
-        "/a": {
-            kind: "EventHandler",
-            method: "GET",
-            eventHandler: async (event) => {
-                return {
-                    statusCode: 200,
-                    body: "<h1>Hello world!</h1>",
-                };
-            }
+    routes: [{
+        path: "/a",
+        method: "GET",
+        eventHandler: async (event) => {
+            return {
+                statusCode: 200,
+                body: "<h1>Hello world!</h1>",
+            };
         },
-        "/b": {
-            kind: "EventHandler",
-            method: "GET",
-            eventHandler: lambda
-        },
-    },
+    }, {
+        path: "/b",
+        method: "GET",
+        eventHandler: lambda,
+    }],
 });
 
 export const url = api.url;
--- a/examples/api/index.ts
+++ b/examples/api/index.ts
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 import * as aws from "@pulumi/aws";
-import { x } from "@pulumi/aws/apigateway";
 import * as pulumi from "@pulumi/pulumi";
 
 const policy: aws.iam.PolicyDocument = {
@@ -43,7 +42,7 @@ const lambda = new aws.lambda.Function("myfunction", {
     runtime: aws.lambda.NodeJS8d10Runtime,
 });
 
-const api = new x.API("myapi", {
+const api = new aws.apigateway.x.API("myapi", {
     routes: [{
         path: "/a",
         method: "GET",
--- a/examples/api/index.ts
+++ b/examples/api/index.ts
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import * as aws from "@pulumi/aws";
+import { x } from "@pulumi/aws/apigateway";
 import * as pulumi from "@pulumi/pulumi";
 
 const policy: aws.iam.PolicyDocument = {
@@ -42,16 +43,24 @@ const lambda = new aws.lambda.Function("myfunction", {
     runtime: aws.lambda.NodeJS8d10Runtime,
 });
 
-const api = new aws.apigateway.x.API("myapi", {
-    routes: [
-        { method: "GET", path: "/a", handler: async (event) => {
-            return {
-                statusCode: 200,
-                body: "<h1>Hello world!</h1>",
-            };
-        }},
-        { method: "GET", path: "/b", handler: lambda },
-    ],
+const api = new x.API("myapi", {
+    routes: {
+        "/a": {
+            kind: "EventHandler",
+            method: "GET",
+            eventHandler: async (event) => {
+                return {
+                    statusCode: 200,
+                    body: "<h1>Hello world!</h1>",
+                };
+            }
+        },
+        "/b": {
+            kind: "EventHandler",
+            method: "GET",
+            eventHandler: lambda
+        },
+    },
 });
 
 export const url = api.url;
--- a/examples/api/step2/index.ts
+++ b/examples/api/step2/index.ts
@@ -15,18 +15,16 @@
 import * as aws from "@pulumi/aws";
 
 const api = new aws.apigateway.x.API("myapi", {
-    routes: {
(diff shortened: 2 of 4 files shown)


examples/api/index.ts
```diff
@@ -42,16 +43,24 @@ const lambda = new aws.lambda.Function("myfunction", {
...
-                statusCode: 200,
-                body: "<h1>Hello world!</h1>",
-            };
-        }},
-        { method: "GET", path: "/b", handler: lambda },
-    ],
+const api = new x.API("myapi", {
+    routes: {
+        "/a": {
+            kind: "EventHandler",
```
CyrusNajmabadi: So, i'm not a huge fan that users of the API have to specify this 'kind' themselves.  We could have a model where this was elided, and we used just the properties of the object to determine what sort of op this was.  However, we'd have to be super careful to make sure that each type of 'route' didn't overlap in terms of the properties they had.

@lukehoban what are your thoughts here?

sdk/nodejs/apigateway/experimental/api.ts
```diff
@@ -69,17 +69,39 @@ export interface Response {
...
+ */
+export type Route = EventHandlerRoute | StaticRoute | ProxyRoute;
+
+/**
+ * A mapping of [path]s to [Route]s.  The [path] is the incoming url path segment to match on.  The
+ * [route] is what behavior should be taken for that path.  See:
+ * https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-method-settings-method-request.html
+ * for more information on the types of paths allowed by apigateway.
+ */
+export type Routes = { [path: string]: Route };
```
CyrusNajmabadi: other potential names: RouteMapping, or RouteMap.  Thoughts?

sdk/nodejs/apigateway/experimental/api.ts
```diff
@@ -69,17 +69,39 @@ export interface Response {
...
+
+export type EventHandlerRoute = {
+    kind: "EventHandler";
```
CyrusNajmabadi: we coudl get rid of 'kind' if we're careful.  see first comment in PR.  Thoughts?

sdk/nodejs/apigateway/experimental/api.ts
```diff
@@ -292,46 +315,72 @@ interface ApigatewayIntegration {
...
+                    "application/json": "{\"message\": \"404 Not found\" }",
+                },
+            },
+            "ACCESS_DENIED": {
+                "statusCode": 404,
+                "responseTemplates": {
+                    "application/json": "{\"message\": \"404 Not found\" }",
+                },
+            },
+        },
```
CyrusNajmabadi: this just moved since it was conceptually simpler to just supply this at creation time, rather than adding it in after the fact.

sdk/nodejs/apigateway/experimental/api.ts
```diff
@@ -292,46 +315,72 @@ interface ApigatewayIntegration {
...
+                    continue;
+                case "Static":
+                    addStaticRouteToSwaggerSpec(name, swagger, path, route, opts);
+                    continue;
+                case "Proxy":
+                    addProxyRouteToSwaggerSpec(name, swagger, path, route, opts);
+                    continue;
+                default:
+                    const exhaustiveMatch: never = route;
+                    throw new Error('Non-exhaustive match for route');
```
CyrusNajmabadi: nifty TS trick so that if we add a new route kind, we'll be forced to update this switch.

sdk/nodejs/apigateway/experimental/api.ts
```diff
@@ -292,46 +315,72 @@ interface ApigatewayIntegration {
...
 
+function addEventHandlerRouteToSwaggerSpec(
+    name: string, swagger: SwaggerSpec, path: string, route: EventHandlerRoute, opts: pulumi.ResourceOptions) {
+
+    const method = swaggerMethod(route.method);
+    const lambdaFunc = lambda.createFunctionFromEventHandler(
+        name + sha1hash(method + ":" + path), route.eventHandler, opts);
+
+    addSwaggerOperation(swagger, path, method, createSwaggerOperationForLambda(lambdaFunc));
     return;
```
CyrusNajmabadi: same logic as before, just withotu the foreach loop over the routes since that is handled by our caller.

sdk/nodejs/apigateway/experimental/api.ts
```diff
@@ -351,19 +400,29 @@ function addRoutesToSwaggerSpec(
...
+    // gateway route to all the s3 bucket objects we create for the files in these directories.
+    const stat = fs.statSync(route.localPath);
+    if (stat.isFile()) {
+        processFile(route);
+    }
+    else if (stat.isDirectory()) {
+        processDirectory(route);
+    }
+
+    return;
```
CyrusNajmabadi: same logic as before, just withotu the foreach loop over the routes since that is handled by our caller.

sdk/nodejs/apigateway/experimental/api.ts
```diff
@@ -550,53 +585,44 @@ function addStaticRoutesToSwaggerSpec(
...
-                        // NLB, which will only be the case for cloud.Service ports exposed as
-                        // type "tcp".
-                        throw new pulumi.ResourceError(
-                            "AWS endpoint proxy requires an Endpoint on a service port of type 'tcp'", opts.parent);
-                    }
-                    return endpoint.loadBalancer.arn;
-                });
+                return endpoint.loadBalancer.arn;
             });
+        });
```
CyrusNajmabadi: i don't know why this indentation change is completely borked in github.

examples/api/index.ts
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import * as aws from "@pulumi/aws";
+import { x } from "@pulumi/aws/apigateway";
```
lukehoban: Nit - but this feels a little odd - `x.API` isn't particularly descriptive.

examples/api/index.ts
lukehoban: Yes - no kind.

examples/api/step2/index.ts
```diff
@@ -15,14 +15,18 @@
...
-    routes: [
-        { method: "GET", path: "/b", handler: async (event) => {
-            return {
-                statusCode: 200,
-                body: "<h1>Hello world!</h1>",
-            };
-        }},
-    ],
+    routes: {
+        "/b": {
```
lukehoban: Hmm - this isn't what I was imagining we would do here.  I thought we were going to keep `routes` as a single list and then just have entries for each kind of route.  Any reason we can't or don't want to do that?

examples/api/index.ts
CyrusNajmabadi: fair enough.  :)

CyrusNajmabadi: done.

CyrusNajmabadi:
Ok.  have incorporated PR feedback.

PR #344, opened 2018-10-05
Add input validation to routes.

CyrusNajmabadi:
We really need to do this all the time with all our APIs.  Things work fine if you're using TS.  But if you use JS, you don't know it won't work and you'll often get a cryptic crash.

### DIFF
--- a/sdk/nodejs/apigateway/experimental/api.ts
+++ b/sdk/nodejs/apigateway/experimental/api.ts
@@ -400,7 +400,7 @@ function addSwaggerOperation(swagger: SwaggerSpec, path: string, method: string,
 }
 
 function checkRoute<TRoute>(route: TRoute, propName: keyof TRoute, opts: pulumi.ResourceOptions) {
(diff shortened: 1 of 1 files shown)


sdk/nodejs/apigateway/experimental/api.ts
```diff
@@ -399,9 +399,18 @@ function addSwaggerOperation(swagger: SwaggerSpec, path: string, method: string,
     swagger.paths[path][method] = operation;
 }
 
+function checkRoute<TRoute>(route: TRoute, propName: keyof TRoute, opts: pulumi.ResourceOptions) {
```
CyrusNajmabadi: using `propName: keyof TRoute` is a cute trick that ensures that hte string literals we pass in are known property names for the typeof route we're checing.  that way we don't rename one, and forget to update the checkRoute calls.

swgillespie: nice!

Issue #348, opened 2018-10-11
Lambda callback value is required by should be optional


lukehoban:
The callback to a lambda should be okay to call with only a single value (the error).  Currently that is not allowed because the signature is:

```
export declare type Callback<E, R> = (event: E, context: Context, callback: (error: any, result: R) => void) => Promise<R> | void;
```

instead of

```
export declare type Callback<E, R> = (event: E, context: Context, callback: (error: any, result?: R) => void) => Promise<R> | void;
```

CyrusNajmabadi:
I think technically both could be optional.  you can just do `callback()` to indicate success (with no result), right?

CyrusNajmabadi:
have confirmed with the docs.

PR #349, opened 2018-10-11
Do not make callback args mandatory as AWS does not require them.

CyrusNajmabadi:
fixes https://github.com/pulumi/pulumi-aws/issues/348

### DIFF
--- a/sdk/nodejs/lambda/lambdaMixins.ts
+++ b/sdk/nodejs/lambda/lambdaMixins.ts
@@ -112,13 +112,23 @@ export interface Context {
 /**
  * Callback is the signature for an AWS Lambda function entrypoint.
  *
+ * [event] is the data passed in by specific services calling the Lambda (like s3, or kinesis).  The
+ * shape of it will be specific to individual services.
+ *
+ * [context] AWS Lambda uses this parameter to provide details of your Lambda function's execution.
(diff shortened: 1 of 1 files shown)


sdk/nodejs/lambda/lambdaMixins.ts
```diff
@@ -112,13 +112,23 @@ export interface Context {
...
+ * for details.
+ *
  * This function can be synchronous or asynchronous function, though async is only supported with an
  * AWS Lambda runtime of 8.10 or higher.  On those runtimes a Promise can be returned, 'callback'
  * parameter can be ignored, and AWS will appropriately handle things. For AWS lambda pre-8.10, a
  * synchronous function must be provided.  The synchronous function should return nothing, and
  * should instead invoke 'callback' when complete.
  */
-export type Callback<E, R> = (event: E, context: Context, callback: (error: any, result: R) => void) => Promise<R> | void;
+export type Callback<E, R> = (event: E, context: Context, callback: (error?: any, result?: R) => void) => Promise<R> | void;
```
CyrusNajmabadi: Note: both args are optional as per the docs:

```
Where:
    error – is an optional parameter that you can use to provide results of the failed Lambda function execution. 
```

PR #354, opened 2018-10-19
getCluster kubeconfig

mazamats:


lukehoban:
Unfortunately the files changed here are auto-generated, so this approach won't work (it'll get overwritten any time we update the provider).

cc @jen20 or @pgavlin to suggest an alternate approach for incorporating this functionality in.

PR #378, opened 2018-11-16
add p3 instance to instanceType

houqp:
Not sure if this file is manually maintained or auto-generated


lukehoban:
> Not sure if this file is manually maintained or auto-generated

Currently manually maintained - so thanks for the PR!  But hoping to move to something auto-generated soon so that this and other enums are kept up to date.

PR #385, opened 2018-11-29
Update to Terraform provider v1.50.0

jen20:
This pull request updates the Terraform provider backing `pulumi-aws` to v1.50.0 (via v1.49.0). There are a number of new resources, and a few errors corrected:

- Gamelift resources were erroneously projected into the Glacier namespace
- Guardduty resources were erroneously projected into the Glacier namespace

We pull in the latest `pulumi-terraform` to fix an error regarding detection of H1 tags in markdown.


Gopkg.toml
```diff
@@ -76,6 +76,12 @@
   [override.metadata]
     govendor-overriden=true
 
+[[override]]
+  name = "github.com/boombuler/barcode"
```
pgavlin: Sheesh, I wonder what's pulling this in...

jen20: Looks to me to be Vault's client.

pgavlin:
We should probably put this in a new release series due to the breaking changes. We'll also want to update `CHANGELOG.md` to reflect this.

sdk/nodejs/glacier/index.ts
```diff
@@ -2,9 +2,5 @@
 // *** Do not edit by hand unless you're certain you know what you are doing! ***
 
 // Export members:
-export * from "./detector";
```
ellismg: We probably want to delete these files from the source tree as well (both here and and in Python and Go).  Because we no longer have an external "overlays" folder, this process is a little tedious. We probably want some sort of script that uses the resources.go file to understand what hand authored overlays and removes all other files before running `tfgen`)

jen20: Done in 1644b43. This is probably worth investigating in `tfgen` at some point though.

jen20:
cc @CyrusNajmabadi re the change in d543db4.

PR #387, opened 2018-11-30
 Automatically set event parent to the resource it was created off of. 

CyrusNajmabadi:
Followup to https://github.com/pulumi/pulumi-aws/pull/386
To discuss with @pgavlin if this is an approach we want to take.

General idea: our subscriptions allow clients to pass in ResourceOptions.  If they do, we respect them.  However, if they don't, we set the parent of hte subscription to the resource being subscribed to. This has the benefit of ensuring that the provider/providers can be picked up automatically without the client needing to explicitly specify them.


CyrusNajmabadi:
Not merging in without discussing things with @pgavlin first.  However, this might be a nice pattern to ensure good behavior, without the specific requirement that clients have to explicitly pass along these opts arguments.

CyrusNajmabadi:
@pgavlin Would like your thoughts here.  Thanks!

pgavlin:
Apologies for taking so long on this...

My primary concern is that it might be a bit difficult for a user to get the contents of the options bag correct w.r.t. to the provider in use. In order for this component to work, at least some of the provider config (the region in particular IIRC) must be shared between the component's resources and the target of the subscription. Perhaps rather than `opts || { parent: foo }`, use `{ parent: foo, ...(opts || {}) }`. This still allows the user to override the parent and other options if desired, but continues to get the parent "right" if it was not specified.

CyrusNajmabadi:
Closing out as this didn't work out due to asynchrony allowing people to observe ciruclarities, which then caused deadlocks.

PR #393, opened 2018-12-05
Introduce Gopkg.template.toml

jen20:
This brings the AWS repository into line with the other Terraform-backed providers (excluding GCP), and allows the workflow for updating this repository to use `govendor-override < Gopkg.template.toml > Gopkg.toml` and removes the need for removing overrides for protocol buffers.


jen20:
It’s already up to date, this is just to bring the workflow in line with other repos.

PR #401, opened 2018-12-14
Don't use aws.Tags for aws.autoscaling.Group::tagsCollection

### LINKED ISSUE #400: aws.autoscaling.Group.tagsCollection does not work
The shape of this input does not match the shape that the underlying Terraform provider expects. This results in the following error when attempting to use `tagsCollection` in an ASG:
```
Diagnostics:
  aws:autoscaling:Group (asg):
    error: Plan apply failed: asg-f5cf5f9: invalid tag attributes: key missing
```

jen20:
The existence of `tagsCollection` on `aws.autoscaling.Group` is to work around a limitation of HCLv1 configuration for Terraform, and while it is not the preferable way to define tags in Pulumi, is sometimes generated by tf2pulumi.

This commit fixes the type generation such that a list of maps with keys consisting of `key`, `value` and `propagate_at_launch` will work for tagging an autoscaling group.

Use of `tags` continues to be preferred.

Fixes #400.

### DIFF
--- a/resources.go
+++ b/resources.go
@@ -433,9 +433,7 @@ func Provider() tfbridge.ProviderInfo {
 						// Conflicts with the pluralized `tag` property, which is the more strongly typed option for
 						// providing tags.  We keep this dynamically typed collection of tags as an option as well, but
 						// give it a different name.
-						Name:        "tagsCollection",
-						Type:        awsType(awsMod, "Tags"),
-						MaxItemsOne: boolRef(true),
+						Name: "tagsCollection",
 					},
 				},
 			},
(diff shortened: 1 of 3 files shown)


jen20:
@lukehoban I'm not sure (yet) if there is a way to override this, but the type we would need here is:

```
interface AsgTag {
    key: string
    value: string
    propagate_at_launch: boolean
}
```

I did this initially (with normal camel cased names) but the `propagateAtLaunch` name was not being converted to `propagate_at_launch` automatically - thought it was better to get the blocker fixed before investigating that further.

lukehoban:
> I did this initially (with normal camel cased names) but the propagateAtLaunch name was not being converted to propagate_at_launch automatically - thought it was better to get the blocker fixed before investigating that further.

Hub - yeah - I'm not actually even sure what we expect to happen to TypeMap fields during translation - but I think https://github.com/pulumi/pulumi-terraform/issues/111 is related here.

PR #402, opened 2018-12-15
Follow new patterns for component resources.

CyrusNajmabadi:
As per our friday meeting:

1. Components do not need to pass 'props' up to pulumi.ComponentResource (and, even if they did, pulumi.ComponentResource would ignore them thanks to https://github.com/pulumi/pulumi/pull/2296).
2. ComponentResources should call `registerOutputs` at the end of processing.  But should generally not include any properties with that call.  actually reporting outputs is really only intended for 'Stack' to do.


CyrusNajmabadi:
Preemptively merging in.  let me know if there are any changes you'd like me to make.

PR #435, opened 2019-01-14
Update downstream proejcts with api quality-of-life improvements in @pulumi/pulumi

CyrusNajmabadi:


resources.go
```diff
@@ -1805,7 +1805,7 @@ func Provider() tfbridge.ProviderInfo {
 		},
 		JavaScript: &tfbridge.JavaScriptInfo{
 			Dependencies: map[string]string{
-				"@pulumi/pulumi":    "^0.16.9",
+				"@pulumi/pulumi":    "=0.16.11-dev.1547665405",
```
CyrusNajmabadi: @ellismg is this ok?  i could not figure out any way to get npm to actually use the latest dev version without doing this.  both `^0.16.11` and `dev` did not work.

resources.go
```diff
@@ -1805,7 +1805,7 @@ func Provider() tfbridge.ProviderInfo {
 		},
 		JavaScript: &tfbridge.JavaScriptInfo{
 			Dependencies: map[string]string{
-				"@pulumi/pulumi":    "^0.16.9",
+				"@pulumi/pulumi":    "dev",
```
CyrusNajmabadi: @ellismg is it ok to reference dev here?

ellismg: We should not check in something referencing dev here. It's fine for doing PRs to get validation, but landing it means we are in a pretty awful state if we ever release this package (in fact, automation now would prevent it from being released).

If I am following along, it should be safe to just do a release of `@pulumi/pulumi` with your most recent changes, right?  If that's the case, we could then update this constraint to be something like ^0.16.11, which is what you want.

If so, we can do the release today (I would like to include https://github.com/pulumi/pulumi/pull/2362 in it)

sdk/nodejs/s3/s3Mixins.ts
```diff
@@ -139,7 +140,7 @@ export class BucketEventSubscription extends lambda.EventSubscription {
             function: this.func,
             action: "lambda:InvokeFunction",
             principal: "s3.amazonaws.com",
-            sourceArn: bucket.id.apply(bucketName => `arn:aws:s3:::${bucketName}`),
+            sourceArn: interpolate `arn:aws:s3:::${bucket.id}`,
```
ellismg: Dumb question.  Why don't we just use `bucket.arn` here?

PR #452, opened 2019-01-29
Deprecate `aws.apigateway.x.Api`

CyrusNajmabadi:
It has moved to `awsx.apigateway.Api`.


lukehoban:
I think we should be careful to hold off on this until we have delivered the replacement in a release. The `aws` package ships frequently, and will likely ship before we do the major release of the `awsx` package so we may want to hold this for a little longer.

PR #462, opened 2019-02-08
Update to upstream v1.58.0

jen20:
This PR updates to version 1.58.0 of upstream.

This is the first version using `gomod-override` for generating `Gopkg.toml`, and brings in the latest `pulumi-terraform`, improving all the examples - hence the high number of files modified.

PR #463, opened 2019-02-11
Update pulumi dependencies for Python package

jen20:


jen20:
Yes - it was already at 50c3039457 from #462, which is the latest revision.

PR #481, opened 2019-02-25
Update to upstream version 1.60.0

jen20:
There are no new resources upstream in v1.60.0 - but several bug fixes.


sdk/nodejs/ec2/defaultNetworkAcl.ts
```diff
@@ -30,65 +30,6 @@ import * as utilities from "../utilities";
  * For more information about Network ACLs, see the AWS Documentation on
  * [Network ACLs][aws-network-acls].
  * 
- * ## Basic Example Usage, with default rules
```
pgavlin: Did the upstream TF lose these examples? cc @joeduffy

jen20: This particular one became invalid: it was edited to stop specifying "0.0.0.0/0" as the CIDR block, and instead has a comment saying "# set a CIDR block here", which causes parsing in `tf2pulumi` to fail with:

```
error: creating tree module: Error parsing main.tf: At 13:5: Unknown token: 13:5 IDENT from_port
```

sdk/nodejs/ec2/defaultNetworkAcl.ts
```diff
@@ -30,65 +30,6 @@ import * as utilities from "../utilities";
...
- * import * as aws from "@pulumi/aws";
- * 
- * const mainvpc = new aws.ec2.Vpc("mainvpc", {
- *     cidrBlock: "10.1.0.0/16",
- * });
- * const defaultDefaultNetworkAcl = new aws.ec2.DefaultNetworkAcl("default", {
- *     defaultNetworkAclId: mainvpc.defaultNetworkAclId,
- *     ingress: [{
- *         action: "allow",
- *         cidrBlock: "0.0.0.0/0",
```
jen20: This is the same as the previous one: the example was changed and is now in valid HCL.

sdk/nodejs/ec2/networkAclRule.ts
```diff
@@ -12,29 +12,6 @@ import * as utilities from "../utilities";
...
- * 
- * ```typescript
- * import * as pulumi from "@pulumi/pulumi";
- * import * as aws from "@pulumi/aws";
- * 
- * const barNetworkAcl = new aws.ec2.NetworkAcl("bar", {
- *     vpcId: aws_vpc_foo.id,
- * });
- * const barNetworkAclRule = new aws.ec2.NetworkAclRule("bar", {
- *     cidrBlock: "0.0.0.0/0",
```
jen20: This is the same as the others - the example has changed to read:

```
  # Opening to 0.0.0.0/0 can lead to security vulnerabilities.
  cidr_block = # add a CIDR block here
```

jen20:
Looks like the lost examples here come from this commit: https://github.com/terraform-providers/terraform-provider-aws/commit/890f63bcac76ae318f2852b2902bd7af38e37169 which removes `0.0.0.0/0` as a CIDR assignment in examples. I guess if this is going to be a thing, we should probably make `tf2pulumi` more resilient to this.

lukehoban:
That’s an unfortunate change - makes it much less clear what a legal value looks like. 

But given it’s an intentional change upstream, seems like this is good to merge on our side. 

Would be nice to open a tracking issue in tf2pulumi to try to handle these cases - though not clear these even parse correctly.

pgavlin:
> Would be nice to open a tracking issue in tf2pulumi to try to handle these cases - though not clear these even parse correctly.

Yeah, these won't even parse. Given that these are examples, we could probably do something clever w.r.t. rewriting the HCL before passing it off.

PR #486, opened 2019-02-27
Implement query primitives

hausdorff:


stack72:
@hausdorff do we still need this?

stack72:
Closing out for now as per convo on slack with @lukehoban - we can reopen when we need to

PR #500, opened 2019-03-12
Update pulumi-terraform

jen20:
This PR pulls in the default value fixes from `pulumi-terraform`. As yet, we are not bumping the version of the upstream provider until it is safe to have breaking changes.


jen20:
I've pushed an update here to bump `pulumi-terraform` up to 129ae82 (unmerged into `pulumi-terraform`) to verify that the tests are fixed.

PR #506, opened 2019-03-27
Fix typing of Role properties

jen20:
Fixes pulumi/pulumi-eks#83


jen20:
pulumi/pulumi-terraform#358 tracks this.

Issue #460, opened 2019-02-07
[appsync] support for pipeline resolvers missing


BerndWessels:
It seems that the support for [pipeline resolvers](https://docs.aws.amazon.com/appsync/latest/devguide/pipeline-resolvers.html) is missing.

If so please take this as a feature request.

If not please point me to the documentation.

Thank you

austin43:
Is there support for resolvers at all? I've not seen a Pulumi method for it and am wondering if I'm missing something.  

https://docs.aws.amazon.com/appsync/latest/APIReference/API_CreateResolver.html

woeps:
Yesterday resolver an schema fuctionality was merged into Terraform 2.4: https://github.com/terraform-providers/terraform-provider-aws/pull/6451
I hope it will land in Pulumi soon.

austin43:
@woeps thanks for the reply, looks like it was merged in to the pulumi provider today. https://github.com/pulumi/terraform-provider-aws

jen20:
This was released in v0.18.1 of `pulumi-aws` - please feel free to reopen if you think something else is missing with respect to this feature!

lukehoban:
> This was released in v0.18.1 of pulumi-aws

@jen20 Were the docs updated for that release?  I believe the feedback here is that these are not yet showing up in the docs.

PR #510, opened 2019-03-31
Update to v2.4.0 of upstream TF provider

jen20:
Somewhat annoyingly, there are breaking changes despite only a minor version bump upstream, so this might want to go out as v0.19.0 just a day after v0.18.0! The breaking changes are listed as:

- `service/appmesh`: Changes to support AppMesh General Availability (GA) release. The AppMesh resources were added while the API was under Public Preview and the GA release earlier this month introduced breaking changes which prevented further AWS Go SDK updates. This is a very atypical situation for the Terraform AWS Provider as most AWS API and SDK changes are additive. To prevent this situation in the future we may introduce separate Terraform AWS Provider(s) specifically for Public Preview or Beta APIs. If or when this occurs, it will be separately announced. The maintainers will continue following Terraform Provider compatibility promises outlined on the HashiCorp Blog and Extending Terraform documentation as best as possible except other existing Public Preview resources.
- `resource/aws_appmesh_virtual_node`: Replace backends configuration block(s) with backend configuration blocks
- `resource/aws_appmesh_virtual_node`: Replace service_discovery configuration block service_name argument with service_discovery configuration block dns configuration block
- `resource/aws_appmesh_virtual_router`: Remove spec configuration block service_names argument


lukehoban:
Upstream didn’t ship this as a major release, so I think we’re okay with a patch release. The break is in an API that was very new and not yet GA.

lukehoban:
@jen20 If this ready to merge and publish?

lukehoban:
I believe this will fix https://github.com/pulumi/pulumi-aws/issues/460?

PR #517, opened 2019-04-10
Introduces a new formalization for creating CloudWatch Metrics and Alarms.

CyrusNajmabadi:
The general approach here is to have packages expose `metricXXX` functions that users can use discover and configure a metric they care about.  For example, a user can now call:

```ts
const metric = aws.lambda.metricDuration();
````

This gives them a metric for the duration, which defaults to the 'average' statistic, and "milliseconds" for the unit, for all the lambdas in the account.  You can easily override the defaults by doing things like:

```ts
const metric = aws.lambda.metricDuration({ statistic: "Sum", unit: "Seconds" });
````

Which would now be tracking the total amount of seconds consumed for all lambdas over a specific period of time.  One can then setup an cloudwatch alarm easily by doing this:

```ts
const metric = aws.lambda.metricDuration({ unit: "Seconds" });
const alarm = metric.createAlarm("highduration", { threshold: 100, period: 120, evaluationPeriods: 5});
```

Because the 'duration' metric defaults to 'Average' statistic and 'Milliseconds' unit, and because alarms default to 'GreaterThanOrEqualToThreshold', this is saying "trigger a cloudwatch alarm when the average duration of the lambdas go over 100s in a 2minute period at least 5 periods in a row".

Currently, if you want to know about the metrics/duration for a specific function, you would do this:

```ts
const func = /* get or create aws.lambda.Function somehow */;
const metric = aws.lambda.metricDuration({ unit: "Seconds", dimensions: { FunctionName: func.Name } });
const alarm = metric.createAlarm("highduration", { threshold: 100, period: 120, evaluationPeriods: 5 });
```

Note that this is somewhat clunky.  I will open a bug to track having instance methods for metrics exposed on teh right resource types.  That way you should be able to write the above as:

```ts
const func = /* get or create aws.lambda.Function somehow */;
const metric = func.metricDuration({ unit: "Seconds" });
const alarm = metric.createAlarm("highduration", { threshold: 100, period: 120, evaluationPeriods: 5 });
```

Note: this PR only includes metrics for the lambda package.  [A followup PR](https://github.com/pulumi/pulumi-aws/pull/518) includes the metrics for all applicable aws packages.


resources.go
```diff
@@ -1875,17 +1875,45 @@ func Provider() tfbridge.ProviderInfo {
 					"utils.ts",  // Helpers,
 				},
 				Modules: map[string]*tfbridge.OverlayInfo{
+					"acmpca": {
+						DestFiles: []string{
+							"metrics.ts",
```
CyrusNajmabadi: general pattern will be that packages expose appropriate metrics through the 'metrics.ts' file/module.

sdk/nodejs/acmpca/metrics.ts
```diff
@@ -0,0 +1,74 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as pulumi from "@pulumi/pulumi";
+import * as cloudwatch from "../cloudwatch";
+
+export type AcmpcaMetricName =
+    "CRLGenerated" | "MisconfiguredCRLBucket" | "Time" | "Success" | "Failure";
```
CyrusNajmabadi: we start wiht the names of the metrics this AWS subsystem supports.

sdk/nodejs/acmpca/metrics.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Creates an AWS/ACMPrivateCA metric with the requested [metricName]. See
+ * https://docs.aws.amazon.com/acm-pca/latest/userguide/PcaCloudWatch.html for list of all
+ * metric-names.
+ *
+ * Note, individual metrics can easily be obtained without supplying the name using the other
+ * [metricXXX] functions.
+ */
+export function metric(metricName: AcmpcaMetricName, change: cloudwatch.MetricChange = {}) {
```
CyrusNajmabadi: always have teh general 'metric' helper that all specific metricXXX helpers defer to.  This has the most docs on how the metrics for this package work (some packages go very much in depth).  We should likely have these become README.md info for that package.

sdk/nodejs/acmpca/metrics.ts
CyrusNajmabadi: Skip this file and go to apigateway, it's more interesting.

sdk/nodejs/apigateway/metrics.ts
```diff
@@ -0,0 +1,124 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as pulumi from "@pulumi/pulumi";
+import * as cloudwatch from "../cloudwatch";
+
+export type ApigatewayMetricName =
+    "4XXError" | "5XXError" | "CacheHitCount" | "CacheMissCount" |
+    "Count" | "IntegrationLatency" | "Latency";
```
CyrusNajmabadi: we start wiht the names of the metrics this AWS subsystem supports.

sdk/nodejs/apigateway/metrics.ts
```diff
@@ -0,0 +1,124 @@
...
+ *    Enabling such metrics will incur additional charges to your account. For pricing information,
+ *    see Amazon CloudWatch Pricing.
+ * 3. "ApiName, Stage": Filters API Gateway metrics for an API stage of the specified API and stage.
+ */
+export function metric(metricName: ApigatewayMetricName, change: cloudwatch.MetricChange = {}) {
+    return new cloudwatch.Metric({
+        namespace: "AWS/ApiGateway",
+        name: metricName,
+        ...change,
+    });
```
CyrusNajmabadi: always have teh general 'metric' helper that all specific metricXXX helpers defer to. This has the most docs on how the metrics for this package work (some packages go very much in depth). We should likely have these become README.md info for that package.

sdk/nodejs/apigateway/metrics.ts
```diff
@@ -0,0 +1,124 @@
...
+ *    specified API, stage, resource, and method.
+ *
+ *    API Gateway will not send such metrics unless you have explicitly enabled detailed CloudWatch
+ *    metrics. You can do this in the console by selecting Enable CloudWatch Metrics under a stage
+ *    Settings tab. Alternatively, you can call the stage:update action of the API Gateway REST API
+ *    to update the metricsEnabled property to true.
+ *
+ *    Enabling such metrics will incur additional charges to your account. For pricing information,
+ *    see Amazon CloudWatch Pricing.
+ * 3. "ApiName, Stage": Filters API Gateway metrics for an API stage of the specified API and stage.
```
CyrusNajmabadi: when we know, we include information about the dimensions that are supported for the metrics for this packag.e  In general, these dimensions should also translate to us providing instance methods on the appropriate resource ot handle this for the user.  For example, since there is an ApiName dimension, we should likely also have all these metrics exposed as instance methods on aws.apigateway.RestApi.  That way you can get the metrics about that specific Api instance, instead of getting it for everything in the account.

That said, it's good to have the top level functions so you have total control adn can get all that aggregated info if you want.  IN other words we should *not* solely provide metrics through instance mehtods on resources.

sdk/nodejs/apigateway/metrics.ts
```diff
@@ -0,0 +1,124 @@
...
+ * The number of client-side errors captured in a specified period.
+ *
+ * The Sum statistic represents this metric, namely, the total count of the 4XXError errors in the
+ * given period. The Average statistic represents the 4XXError error rate, namely, the total count
+ * of the 4XXError errors divided by the total number of requests during the period. The denominator
+ * corresponds to the Count metric (below).
+ */
+export function metric4XXError(change: cloudwatch.MetricChange = {}) {
+    return metric("4XXError", { unit: "Count", ...change });
+}
```
CyrusNajmabadi: now we follow with all the metricXXX methods.  Each one is named so that it is basically `"metric" + metricName`.  So, in this case, because the metric name in apigateway is `"4XXError"`, we name it `metric4XXError`.  When the aws docs have made it clear what the preferred unit/statistic is, we've put that info in here so you can just call `metricXXX` and not have to tweak anything to get what is likely the most useful metric.

sdk/nodejs/apigateway/metrics.ts
```diff
@@ -0,0 +1,124 @@
...
+export function metricCount(change: cloudwatch.MetricChange = {}) {
+    return metric("Count", { unit: "Count", ...change });
+}
+
+/**
+ * The time between when API Gateway relays a request to the back end and when it receives a
+ * response from the back end.
+ */
+export function metricIntegrationLatency(change: cloudwatch.MetricChange = {}) {
+    return metric("IntegrationLatency", { unit: "Milliseconds", ...change });
```
CyrusNajmabadi: an example of a different type of 'unit' here.

sdk/nodejs/cloudwatch/metric.ts
```diff
@@ -0,0 +1,402 @@
...
+        this.dimensions = pulumi.output(args.dimensions);
+        this.namespace = pulumi.output(args.namespace);
+        this.period = utils.ifUndefined(args.period, 60).apply(validatePeriod);
+        this.statistic = pulumi.all([args.statistic, args.extendedStatistic])
+                               .apply(([statistic, extendedStatistic]) => validateStatistics(statistic, extendedStatistic));
+        this.extendedStatistic = pulumi.output(args.extendedStatistic).apply(validateExtendedStatistic);
+        this.unit = pulumi.output(args.unit);
+    }
+
+    public with(change: MetricChange | undefined) {
```
CyrusNajmabadi: helper method so that one can easily take the default metrics exposed by packages, and tweak what representation you want.  Common things to do would be to change the unit, period or statistic you care about hte metric for.  for example, while the default period for metrics is 60 seconds, you could always call `metricXXX({ period: 120 })` to ask for the same metric for a 120 second period.

sdk/nodejs/cloudwatch/metric.ts
```diff
@@ -0,0 +1,402 @@
...
+
+            dimensions: this.dimensions,
+            extendedStatistic: this.extendedStatistic.apply(s => s === undefined ? undefined : `p${s}`),
+            metricName: this.name,
+            namespace: this.namespace,
+            period: this.period,
+            statistic: this.statistic,
+            unit: this.unit,
+        }, { parent: this.resource, ...opts });
+    }
```
CyrusNajmabadi: needs docs, but this makes it easy to create a real cloudwatch alarm from a given metric.

sdk/nodejs/cloudwatch/metric.ts
```diff
@@ -0,0 +1,402 @@
...
+
+export type MetricStatistic = "SampleCount" | "Average" | "Sum" | "Minimum" | "Maximum";
+export type MetricUnit =
+    "Seconds"       | "Microseconds"        | "Milliseconds"        |
+    "Bytes"         | "Kilobytes"           | "Megabytes"           | "Gigabytes"           | "Terabytes"           |
+    "Bits"          | "Kilobits"            | "Megabits"            | "Gigabits"            | "Terabits"            |
+    "Percent"       | "Count"               |
+    "Bytes/Second"  | "Kilobytes/Second"    | "Megabytes/Second"    | "Gigabytes/Second"    | "Terabytes/Second"    |
+    "Bits/Second"   | "Kilobits/Second"     | "Megabits/Second"     | "Gigabits/Second"     | "Terabits/Second"     |
+    "Count/Second"  | "None";
```
CyrusNajmabadi: standard values so that TypeScript will typecheck you.

sdk/nodejs/cloudwatch/eventMetrics.ts
```diff
@@ -0,0 +1,94 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as pulumi from "@pulumi/pulumi";
+import { Metric, MetricChange } from "./metric";
+
+export type CloudwatchEventMetricName =
+    "DeadLetterInvocations" | "Invocations" | "FailedInvocations" | "TriggeredRules" |
+    "MatchedEvents" | "ThrottledRules";
```
CyrusNajmabadi: humorously (but for good purpose) CloudWatch publishes its own events on how it is being used.  So the aws.cloudwatch package is both whre we define things like the Metric/Alarm types, but it's also where you can get metrics for, and create alarms from, how your actual cloudwatch subsystem is being used.

examples/metrics/index.ts
```diff
@@ -0,0 +1,40 @@
...
+        try {
+            const res = await fetch.default(url);
+        } catch (err) {
+            console.log(`${url}: Failed to GET`);
+            return;
+        }
+    }
+});
+
+const metric = aws.lambda.metricDuration({ unit: "Seconds", dimensions: { FunctionName: subscription.func.name } });
```
ekrengel: Is there a way to simplify this so we create the alarm and metric at the same time? Or is that something that would go in the AWSX layer?

sdk/nodejs/cloudwatch/metric.ts
```diff
@@ -0,0 +1,402 @@
...
+ * points that are published to CloudWatch. Think of a metric as a variable to monitor, and the data
+ * points as representing the values of that variable over time. For example, the CPU usage of a
+ * particular EC2 instance is one metric provided by Amazon EC2. The data points themselves can come
+ * from any application or business activity from which you collect data.
+ *
+ * AWS services send metrics to CloudWatch, and you can send your own custom metrics to CloudWatch.
+ * You can add the data points in any order, and at any rate you choose. You can retrieve statistics
+ * about those data points as an ordered set of time-series data.
+ *
+ * Metrics exist only in the region in which they are created. Metrics cannot be deleted, but they
```
ekrengel: If a metric cannot be deleted, how will Pulumi handle this?

CyrusNajmabadi: discussed offline.

examples/metrics/index.ts
CyrusNajmabadi: it's a good question.  we could consdier just having similar things like "alarmDuration" to go along with all the metrics.  The things that are pushign me away from doing that are:

1. we'd end up with every 'metrixXXX' method duplicated.
2. we'll eventually want to tie these metrics in cloudwatch dashboards/widgets.  And then i'm loathe to then have widgetXXX methods to go along with all the metricXXX and alarmXXX methods.  

I think it's good to have 'metric' as the underlying primitive, and then provide the helpers on top of that.

Issue #520, opened 2019-04-11
Expose suitable metrics as instance methods off their respective resources.


CyrusNajmabadi:
We expose many metrics off off the 'metrics' sub-module for many modules.  However, those metrics tend to be for the entirety of that AWS sub-system.  Many metrics supply  'dimensions' to allow you to suitably narrow down the result to some particular AWS resource.  

In the cases where AWS does provide this, we should expose these metrics as instance methods on the resources themselves.  

A non-exhaustive list of where we can do this is:

- [x] lambda function metrics. https://github.com/pulumi/pulumi-aws/pull/522
- [x] apigateway restapi. https://github.com/pulumi/pulumi-aws/pull/524
- [x] sns topic. https://github.com/pulumi/pulumi-aws/pull/526
- [x] sqs queue. https://github.com/pulumi/pulumi-aws/pull/525
- [x] dynamodb table. https://github.com/pulumi/pulumi-aws/pull/527
- [x] s3 bucket. https://github.com/pulumi/pulumi-aws/pull/529
- [x] ec2 isntances. https://github.com/pulumi/pulumi-aws/pull/530
- [x] cloudfront distribution. https://github.com/pulumi/pulumi-aws/pull/531
- [x] ecs cluster. https://github.com/pulumi/pulumi-aws/pull/532
- [x] cognito user pool. https://github.com/pulumi/pulumi-aws/pull/533
- [x] cloudwatch rule, loggroup, filter, destination. https://github.com/pulumi/pulumi-aws/pull/534
- [x] codebuild project. https://github.com/pulumi/pulumi-aws/pull/536
- [ ] ecs service. Needs to be done in awsx.
- [ ] loadbalancers+targetgroups.  needs to be done in awsx.

CyrusNajmabadi:
This has been completed to our desired level.

PR #522, opened 2019-04-11
 Draft on how we can expose metrics as instance methods on appropriate resources.

CyrusNajmabadi:
Issue tracking all the resources we'd likely want to target is here:

https://github.com/pulumi/pulumi-aws/issues/520


examples/metrics/index.ts
```diff
@@ -34,7 +34,7 @@ const subscription = topic.onEvent("for-each-url", async (event) => {
     }
 });
 
-const metric = aws.lambda.metrics.duration({ unit: "Seconds", dimensions: { FunctionName: subscription.func.name } });
-const alarm = metric.createAlarm("LongDuration", { threshold: 60, evaluationPeriods: 2 });
+const metric = subscription.func.metricDuration({ unit: "Seconds" });
```
CyrusNajmabadi: Note how there is no need ot provide 'dimensions' anymore, we do that on the user's behalf.

examples/metrics/index.ts
```diff
@@ -34,7 +34,7 @@ const subscription = topic.onEvent("for-each-url", async (event) => {
     }
 });
 
-const metric = aws.lambda.metrics.duration({ unit: "Seconds", dimensions: { FunctionName: subscription.func.name } });
-const alarm = metric.createAlarm("LongDuration", { threshold: 60, evaluationPeriods: 2 });
+const metric = subscription.func.metrics.duration({ unit: "Seconds" });
```
CyrusNajmabadi: Note how there is no need ot provide 'dimensions' anymore, we do that on the user's behalf.

sdk/nodejs/acmpca/metrics.ts
```diff
@@ -38,7 +38,7 @@ export module metrics {
     /**
      * A certificate revocation list (CRL) was generated. This metric applies only to a private CA.
      */
-    export function crlGenerated(change: cloudwatch.MetricChange) {
+    export function crlGenerated(change?: cloudwatch.MetricChange) {
```
CyrusNajmabadi: mistake in signature i'm just fixing up in this PR.

sdk/nodejs/lambda/metrics.ts
```diff
@@ -139,3 +141,97 @@ export module metrics {
...
+        return {
+            invocations: (change: cloudwatch.MetricChange) => getMetric(lambda, change, metrics.invocations),
+            errors: (change: cloudwatch.MetricChange) => getMetric(lambda, change, metrics.errors),
+            deadLetterErrors: (change: cloudwatch.MetricChange) => getMetric(this, change, metrics.deadLetterErrors),
+            duration: (change: cloudwatch.MetricChange) => getMetric(lambda, change, metrics.duration),
+            throttles: (change: cloudwatch.MetricChange) => getMetric(lambda, change, metrics.throttles),
+            iteratorAge: (change: cloudwatch.MetricChange) => getMetric(lambda, change, metrics.iteratorAge),
+        }
+    }
+});
```
CyrusNajmabadi: need to define hte 'metrics' property this way as opposed to just attaching something to the prototype.  This is necessary so we can have 'metrics' be a 'getter' function, allowing us access to the right 'this' value.  We need to stash that away so we can pass it along to all the sub-functions properly.

PR #547, opened 2019-04-24
Bump pulumi/pulumi to 0.17.8

lukehoban:
Pulls in the `ignoreChanges` and adds a test case.

### DIFF
--- a/examples/ignoreChanges/index.ts
+++ b/examples/ignoreChanges/index.ts
@@ -1,0 +1,8 @@
+// Copyright 2016-2017, Pulumi Corporation.  All rights reserved.
+
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+
+const bucket = new aws.s3.Bucket("mybucket", {
+    bucketPrefix: "foo"
+}, { ignoreChanges: ["bucketPrefix"]});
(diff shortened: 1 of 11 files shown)


examples/ignoreChanges/index.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright 2016-2017, Pulumi Corporation.  All rights reserved.
+
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+
+const bucket = new aws.s3.Bucket("mybucket", {
+    bucketPrefix: "foo"
+}, { ignoreChanges: ["bucketPrefix"]});
```
ellismg: What do you think about removing the `ignoreChanges` member here (I see why you need it in the second step, but it shouldn't be needed here (and we want to validate that even if you forgot to set it on the initial update, you could set it on the subsequent one, right)?

lukehoban: I could, and it makes it a little more targeted.  But the normal use case for this would be that the property is set both before and after the change, and in some sense this provides additional coverage that the `ignoreChanges` doesn't impact the creation, but does impact the update.  I think I slightly prefer it as is.

PR #549, opened 2019-04-30
Add hypothetical IAM helpers

joeduffy:
Per Slack conversation with @CyrusNajmabadi, this is one hypothetical approach to https://github.com/pulumi/pulumi-awsx/issues/242. This helps to reduce examples like

```typescript
const baseIamRole = new aws.iam.Role("baseIamRole", {
    path: "/",
    assumeRolePolicy: {
        "Version": "2012-10-17",
        "Statement": [{
            "Action": [
                "sts:AssumeRole",
            ],
            "Effect": "Allow",
            "Principal": {
                "Service": [
                    "ec2.amazonaws.com",
                ],
            },
        }],
    },
});
```

to the far simpler

```typescript
const baseIamRole = new aws.iam.Role("baseIamRole", {
    path: "/",
    assumeRolePolicy: aws.iam.createPolicy({
        action: "sts:AssumeRole",
        servicePrincipal: "ec2.amazonaws.com",
    }),
});
```

Of course, this is a vast subset of what we can do -- I think there are many interesting examples for things like `attachPolicy` on `User`, etc. But anyway, just an example approach for consideration.

### DIFF
--- a/sdk/nodejs/iam/documents.ts
+++ b/sdk/nodejs/iam/documents.ts
@@ -44,12 +44,18 @@
  */
 export interface PolicyDocument {
     // The version of the policy language that you want to use. As a best practice, use the latest `2012-10-17` version.
-    Version: "2008-10-17" | "2012-10-17";
+    Version: PolicyVersion;
     // An optional document ID.
     Id?: string;
     // One or more policy statements, describing the effect, principal, action, resource, and condition.
     Statement: PolicyStatement[];
 }
+
+/**
+ * PolicyVersion is a union type of all possible IAM policy document versions supported. As a best practice,
+ * always use the latest supported version.
+ */
+export type PolicyVersion = "2008-10-17" | "2012-10-17";
 
 /**
  * The Statement element is the main element for a policy. This element is required. It can include multiple elements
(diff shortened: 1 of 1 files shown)


joeduffy:
@lukehoban What do you think about this approach? I've been using it on some projects recently and, though simple, it really does clean up some of the strange/hard-coded IAM policies. I can personally never remember things like the version number, etc. This could be a nice/minor add for the launch.

CyrusNajmabadi:
I'd go even further.  i..e have `awsx.iam.assumeRolePolicy("ec2.amazonaws.com")`.  I think there are ample areas where we can drive simple entrypoints to create this stuff.  

However, i don't see a huge rush on this, and i'd really like a good 48-72 hours spent on collecting lots of the use cases and driving toward a nice API for all those cases.

casey-robertson:
upvote - doing AWS policies in code is a dumpster fire.

lukehoban:
@casey-robertson Anything in particular you'd like to see made simpler?

Note that we already have `aws.iam.assumeRolePolicyForPrincipal` - so the original code example above can be:

```
const baseIamRole = new aws.iam.Role("baseIamRole", {
    path: "/",
    assumeRolePolicy: aws.iam.assumeRolePolicyForPrincipal({Service: "ec2.amazonaws.com" }),
});
```

The changes in this PR provide a bit of additional flexibility - and might be useful - but would love to hear your use cases (possibly in a new issue) to make sure we do work that addresses them.

thomas11:
closed after confirming with lukehoban.

PR #552, opened 2019-05-04
Add support for AWS Lambda custom runtimes

j-veiga:
Hi, would like to have the ability to use custom runtimes with Pulumi in nodejs/ts without having to patch the aws module. Press release, but you are likely to be aware: https://aws.amazon.com/about-aws/whats-new/2018/11/aws-lambda-now-supports-custom-runtimes-and-layers/


jen20:
Thanks, @j-veiga!

j-veiga:
@jen20 thanks for accepting.

PR #554, opened 2019-05-06
Upgrade to v2.9.0 of the upstream Terraform provider

jen20:


swgillespie:
This PR contains https://github.com/pulumi/pulumi-terraform/pull/369, so it would be nice to smoke-test this really quick just to make sure it's not totally hosed. Seems like we've got some tests in this repo so LGTM 👍

PR #560, opened 2019-05-14
Update Node Version in serverless-raw examples

ellismg:
Node 6.10 can no longer be used to create new AWS Lambda functions, it
has been EOL'd upstream and Amazon no longers supports creating new
lambdas with this runtime:

```
Error creating Lambda function: InvalidParameterValueException:
The runtime parameter of nodejs6.10 is no longer supported for
creating or updating AWS Lambda functions. We recommend you use the
new runtime (nodejs8.10)
```


ellismg:
This also includes a change to bump the version of pulumi we are using in the webserver-py example, so it gets a newer version that has the custom resource options.  We can change the codegen to now require this, but for now I want to get these builds healthy again and there may be more stuff lingering behind this we also have to change.

PR #565, opened 2019-05-17
Update Python SDK in Webserver example

jen20:
I think there is no need to wait for CI on this one, since a pull request build will not exercise it in any case.


swgillespie:
We need to bump this as well, or else we'll keep seeing this in other Pulumi programs:

https://github.com/pulumi/pulumi-aws/blob/9dbe8eae51f0b58c8d18253141476073d8e12337/resources.go#L1982

swgillespie:
apologies for the poor communication on the breaking-ish change here - I was not anticipating that this was going to cause downstream trouble, although I should have.

jen20:
I've updated the requirement that will go into the package to be the same.

PR #568, opened 2019-05-18
Pull in newer pulumi-terraform

ellismg:
This should fix the python issues we are seeing.  As part of this work, I am going to start running our one Python test as part of PRs, to have some python coverage.


jen20:
I've pushed an additional commit here which re-runs the code generation with the newer dependencies - this was the reason for the CI failure of 55c435c.

PR #569, opened 2019-05-18
Add nodejs10.x

ellismg:
AWS now supports this runtime, see:
https://aws.amazon.com/about-aws/whats-new/2019/05/aws_lambda_adds_support_for_node_js_v10/


ellismg:
Not really sure how we should think about removing the existing Node6.10 value.  AWS no longer allows creating or updating lambdas on the runtime, so it feels like we should remove it at some point.  Maybe the next time we do a minor version update?

PR #570, opened 2019-05-19
Hard code the regions for multiple-regions example

jen20:
The recent changes in `pulumi-terraform` have changed the type of `aws.config.region` to possibly be undefined. Although we could cast here, it seems somewhat neater to just hard-code both region names instead of having one configured and one not.


jen20:
I'll go ahead and merge this (and tag for post-hoc review), since a pull-request build will not exercise this code.

Issue #252, opened 2018-06-29
Cannot authenticate on AWS using switchrole


jnancel:
Hello,

I'm trying to bootstrap my first pulumi project but I'm encountering an issue with the supported authentication method on AWS. 

If I'm using a profile in which my credentials are defined ( with `aws_access_key_id` and `aws_secret_access_key` ), it works like a charm. But that's not how we work in my company, we have a source account and we're using switchrole to jump from that source account to all other accouts ( this is standard AWS procedure when you're dealing with multiple accounts ).

Here's the look of my profiles in `~/.aws/config` : 

```
[profile myprofile]
role_arn = arn:aws:iam::<AWS destination account id>:role/<the role I'm going to impersonate>
mfa_serial = arn:aws:iam::<source AWS account ID>:mfa/<IAM User name on the source account>
source_profile = <the profile name in which my keys are defined for my source account>
```

Now when I'm using one of those profile, it fails : 
```
$ AWS_PROFILE=myprofile pulumi update
Previewing update of stack 'prod'
Previewing changes:

     Type                    Name               Plan          Info
 *   global                  global             no change     1 error
 +   └─ pulumi:pulumi:Stack  first-pulumi-prod  create        
 
Diagnostics:
  global: global
    error: unable to discover AWS AccessKeyID and/or SecretAccessKey - see https://pulumi.io/install/aws.html for details on configuration
 
error: an error occurred while advancing the preview
```

When using the same command line with my source profile ( the one with the keys ), I have the expected result :
```
      Type                 Name               Plan       Info
 +   pulumi:pulumi:Stack  first-pulumi-prod  create
 +   └─ aws:s3:Bucket     my-bucket          create

info: 2 changes previewed:
    + 2 resources to create

Do you want to perform this update?
  yes
> no
  details
```

Is there any configuration I'm missing or is it not possible for the moment to use an AWS profile with switchrole ?

Thanks

jnancel:
Hello Matt,

Thanks for the answer, I tried enabling that option but encountered a `panic` error : 
```
$ AWS_PROFILE=myprofile pulumi preview
Previewing update of stack 'prod'
Previewing changes:

     Type    Name    Plan          Info
 *   global  global  no change     1 error, 2 info messages
 
Diagnostics:
  global: global
[...]

PR #579, opened 2019-05-30
Fixed bug, where assume-role was not working properly (#252)

mi5guided:
From my take, the `provider` portions of the creds.Get() call have been deprecated and full credential functionality taken over by the new `awsbase.GetCredentials()` call. That seems to be the breaking change a few months ago.

The fix is to remove the call to `creds.Get()` and refactor the call to `awsbase.GetGredentials()` to get rid of the unused variable the Go compiler is complaining about.


jen20:
I believe this will fix #252 after testing.

jen20:
Hi @mi5guided! Thanks for opening this PR. I've taken these commits and added a few more to fix up using `AWS_PROFILE` as well as the `~/.aws/config` file and opened #580 with it - your commits will get merged with that. Thanks for contributing!

PR #580, opened 2019-05-30
Allow assuming a role via AWS_PROFILE or configuration files

### LINKED ISSUE #252: Cannot authenticate on AWS using switchrole

jen20:
This PR enhances the work done by @mi5guided in #579 to allow profiles defined via `AWS_PROFILE` to be used as well as the AWS config files.

Thanks, @mi5guided!

Fixes #252.

### DIFF
--- a/resources.go
+++ b/resources.go
@@ -1,4 +1,4 @@
-// Copyright 2016-2018, Pulumi Corporation.(
+// Copyright 2016-2018, Pulumi Corporation.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -191,11 +191,7 @@ func preConfigureCallback(vars resource.PropertyMap, c *terraform.ResourceConfig
 	}
 	config.CredsFilename = credsPath
 
-	// TODO[pulumi/pulumi-terraform#48] We should also be setting `config.AssumeRole*` here, but we are currently
-	// blocked on not being able to read out list-valued provider config.
-
-	_, err = awsbase.GetCredentials(config)
-	if err != nil {
+	if _, err := awsbase.GetCredentials(config); err != nil {
 		return errors.New("unable to discover AWS AccessKeyID and/or SecretAccessKey " +
 			"- see https://pulumi.io/install/aws.html for details on configuration")
 	}
@@ -222,6 +218,11 @@ func Provider() tfbridge.ProviderInfo {
 				Type: awsType("region", "Region"),
 				Default: &tfbridge.DefaultInfo{
 					EnvVars: []string{"AWS_REGION", "AWS_DEFAULT_REGION"},
+				},
+			},
+			"profile": {
+				Default: &tfbridge.DefaultInfo{
+					EnvVars: []string{"AWS_PROFILE"},
 				},
 			},
 		},
--- a/resources.go
+++ b/resources.go
@@ -194,13 +194,7 @@ func preConfigureCallback(vars resource.PropertyMap, c *terraform.ResourceConfig
 	// TODO[pulumi/pulumi-terraform#48] We should also be setting `config.AssumeRole*` here, but we are currently
 	// blocked on not being able to read out list-valued provider config.
 
-	creds, err := awsbase.GetCredentials(config)
-	if err != nil {
-		return errors.New("unable to discover AWS AccessKeyID and/or SecretAccessKey " +
-			"- see https://pulumi.io/install/aws.html for details on configuration")
-	}
-
-	_, err = creds.Get()
+	_, err = awsbase.GetCredentials(config)
 	if err != nil {
 		return errors.New("unable to discover AWS AccessKeyID and/or SecretAccessKey " +
 			"- see https://pulumi.io/install/aws.html for details on configuration")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ### Improvements
 
-- Addition of enums for SSM Parameter Types 
+- Add an enumeration for SSM Parameter Types
+- Fixed a bug where `AWS_PROFILE` and the ~/.aws/config files were not honoured.
 
 ## 0.18.6 (Released May 24th, 2019)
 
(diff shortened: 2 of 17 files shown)


jen20:
Sadly I don’t think so without making the test modify configuration files it probably should never touch.

resources.go
```diff
@@ -191,17 +191,7 @@ func preConfigureCallback(vars resource.PropertyMap, c *terraform.ResourceConfig
...
-
-	creds, err := awsbase.GetCredentials(config)
-	if err != nil {
-		return errors.New("unable to discover AWS AccessKeyID and/or SecretAccessKey " +
-			"- see https://pulumi.io/install/aws.html for details on configuration")
-	}
-
-	_, err = creds.Get()
-	if err != nil {
+	if _, err := awsbase.GetCredentials(config); err != nil {
```
lukehoban: I'm not sure I follow why this is the right fix.

All of the code here is intended to catch and report failures to authenticate prior to the Terraform Provider `Configure` call reporting back failures.  But the goal is to do the same checks the Terraform Provider will do.

Per https://github.com/hashicorp/aws-sdk-go-base/blob/8fac28668e59a2e5afa6b2d034614c648ddc418e/session.go#L31, it appears the call to `creds.Get()` will be made there.  

Is the problem instead that `Profile:   stringValue(vars, "profile"),` isn't correctly reading the profile here?

lukehoban:
> LGTM. Is there any way for us to get test coverage here as well?

Can we test via creating a first-class provider?

jen20:
The build failure is due to missing code gen - I'll push a fix for that now. I don't think testing with a first-class provider doesn't help (and it already worked if `profile` was set as a parameter) - we'd still need to edit the `~/.aws/credentials` or `~/.aws/config` files to create a known profile to use, which is something we should probably not do in a script lest we wipe out someone's local config?

lukehoban:
> I don't think testing with a first-class provider doesn't help (and it already worked if profile was set as a parameter) - we'd still need to edit the ~/.aws/credentials or ~/.aws/config files to create a known profile to use, which is something we should probably not do in a script lest we wipe out someone's local config?

Ahh - yeah - makes sense.

resources.go
jen20: The reason using `AWS_PROFILE` was failing is because we weren't reading the profile correctly before - whether through a profile configured in `~/.aws/config` or `~/.aws/credentials`.

I don't think removing the call to `Get` on the returned `Credentials` struct is _strictly_ necessary to fix the bug, but I also don't think it buys us anything to do it. `GetCredentials` itself calls `Get` here (after constructing a valid credentials chain):

https://github.com/hashicorp/aws-sdk-go-base/blob/master/awsauth.go#L239

It also tests that a role can be assumed, if appropriate:

https://github.com/hashicorp/aws-sdk-go-base/blob/master/awsauth.go#L283

If we hit either of these cases, the call to `GetCredentials` would have failed before _anyway_, so `Get` would not be called. If it worked the first time (in `GetCredentials`), calling it a second time should not yield different behaviour without configuration changes.

It's worth noting that the `aws-sdk-go-base` library is newer than the Pulumi AWS provider, and that the checks done by Terraform have changed as a result of using this library.

Finally, the error message we produce in response to _any_ error from configuration complains about `AccessKeyID` and `SecretAccessKey`. In future, we should change the error to reflect the actual source of the problem. We _could_ duplicate the entire logic of `GetCredentials` in the pre-configure callback to give better error messages here, but it like we'd be better off doing string detection on the error values to avoid doing so.

lukehoban: >  don't think removing the call to Get on the returned Credentials struct is strictly necessary to fix the bug, but I also don't think it buys us anything to do it. 

Got it - makes sense.

> Finally, the error message we produce in response to any error from configuration complains about AccessKeyID and SecretAccessKey. In future, we should change the error to reflect the actual source of the problem. We could duplicate the entire logic of GetCredentials in the pre-configure callback to give better error messages here, but it like we'd be better off doing string detection on the error values to avoid doing so.

Totally agreed.

resources.go
```diff
@@ -230,6 +220,11 @@ func Provider() tfbridge.ProviderInfo {
 					EnvVars: []string{"AWS_REGION", "AWS_DEFAULT_REGION"},
 				},
 			},
+			"profile": {
+				Default: &tfbridge.DefaultInfo{
+					EnvVars: []string{"AWS_PROFILE"},
+				},
+			},
```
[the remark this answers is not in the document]
jen20: I think we should - I'll follow up on this with a new PR soon.

PR #582, opened 2019-06-04
Implement type guards for AWS resources

hausdorff:


jen20:
Hi @hausdorff! Looks like we have a few different problems here:

- I think the intention is to depend on an unmerged PR in `pulumi-terraform` - we should merge that before doing this (though you can check the effects on generation from the `pulumi-terraform` build directly).
- You may be missing the latest version of `tf2pulumi` on your `PATH` locally, so examples get removed
- I think dependencies were updated using `GO111MODULE=on go mod vendor` instead of `make ensure` which has extra steps to copy provider documentation from the module cache - hence the docs being removed
- We seem to have spurious dependency changes: `GO111MODULE=on go get github.com/pulumi/pulumi-terraform@master` once the upstream PR is merged should do the trick for updating this with just one change to `go.mod`.

Happy to help with ensuring your local environment is good once you're online!

hausdorff:
* I put these PRs so they'd be ready when that PR did merge. The intent is not to merge against the dev branch.
* `tf2pulumi` was actually updated updated immediately before building these repositories.
* We used `make ensure`, not `go mod vendor`.
* Why is this different from taking a dependency on the dev version? It seems like they should be the same and we should be able to cut over once it merges.

hausdorff:
Another thing that might be relevant is that adding our `gomod-override` is what resulted in all this stuff getting deleted.

jen20:
Superseded by #586

PR #597, opened 2019-06-11
Adding Route53 RecordType constants

stack72:
//cc @clstokes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 * Add constants for Aurora-based RDS Cluster `EngineType` and `EngineMode`
 * Add constants for EC2 Seurity Group Protocols
 * Add constants for ALB IpAddressType and LoadBalancer type
+* Add constants for Route53 record types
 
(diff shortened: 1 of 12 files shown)


jen20:
@joeduffy @lukehoban Any thoughts on switching to moving to having the constant versions in abstract classes like this to clean up autocomplete?

joeduffy:
I personally love it. TBH, I wished we had this very thing when writing the crosswalk docs. E.g., https://pulumi.io/reference/crosswalk/aws/iam/#using-pre-defined-iam-managed-policies -- I wanted to say, "just look at the `ManagedPolicies` type, and all of the constants" ... but instead they're constants at the top-level, without a common prefix, so it ultimately was hard to say anything other than "well poke around in the package and you'll find some constants." (Unless I am missing an existing class here!)

jen20:
I don't think you're missing one - along with principals they're at the top level in `aws.iam` which makes navigation somewhat unwieldy. Let's go with this and refactor others over time to be in this same format?

PR #598, opened 2019-06-11
Pass a parent along to pick up providers.

CyrusNajmabadi:
Note: this PR makes use of aliases!  I manually validated the appropraite behavior here.  Tagging @ellismg  for views on incorporating this into a fully fledged test in the future!


sdk/nodejs/cloudwatch/logGroupMixins.ts
```diff
@@ -76,7 +76,7 @@ export class LogGroupEventSubscription extends lambda.EventSubscription {
         name: string, logGroup: logGroup.LogGroup, handler: LogGroupEventHandler,
         args: LogGroupEventSubscriptionArgs = {}, opts: pulumi.ComponentResourceOptions = {}) {
 
-        super("aws:cloudwatch:LogGroupEventSubscription", name, opts);
+        super("aws:cloudwatch:LogGroupEventSubscription", name, { parent: logGroup, ...opts });
```
CyrusNajmabadi: this is our pattern elsewhere.  it allows the caller ot have final say (through ...opts) but allows the current type to say "i'm the appropriate parent otherwise" (by being first in the splat)

PR #600, opened 2019-06-12
Move constants into modules

jen20:
As per the pattern established between and #597 and #599 and the discussion around it, this PR adds modules for the related groups of constants for AWS:

- `ec2.InstancePlatforms`
- `ec2.InstanceTypes`
- `ec2.Tenancies`
- `iam.ManagedPolicies`
- `iam.Principals`

The existing constants are left alone, but marked with `@deprecated` in the `tsdoc`, to be removed at some future time.


CyrusNajmabadi:
Hi there!  Could i be included on these reviews?  Thanks!

PR #608, opened 2019-06-18
Disable smart quotes in generated README.rst

ellismg:
`pandoc` will output using "smart quotes" by default, which caues the
README.rst which is included in our pyton package to include some non
ASCII characters, which blocks installs on some machines.

This change removes the use of smart quotes. The correct long term fix
is tracked by: https://github.com/pulumi/pulumi-terraform/pull/388 but
this will unblock installs of this package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ build:: provider tfgen
 		sed -i.bak "s/\$${VERSION}/$(VERSION)/g" ./bin/package.json
 	cd ${PACKDIR}/python/ && \
 		if [ $$(command -v pandoc) ]; then \
-			pandoc --from=markdown --to=rst --output=README.rst ../../README.md; \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -33,7 +33,7 @@ build:: provider tfgen
 		sed -i.bak "s/\$${VERSION}/$(VERSION)/g" ./bin/package.json
 	cd ${PACKDIR}/python/ && \
 		if [ $$(command -v pandoc) ]; then \
-			pandoc --from=markdown --to=rst --output=README.rst ../../README.md; \
+			pandoc --from=markdown-smart --to=rst-smart --output=README.rst ../../README.md; \
```
ellismg: This looks a little confusing, the `-smart` here is not a suffix but an instruction to disable the smart quotes extension.

Issue #610, opened 2019-06-18
Update Lambda Runtime Constants


stack72:
We need to deprecate the old runtime constants and then be able to add the new ones so we allow people to move forward

https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

PR #611, opened 2019-06-19
Update the available Lambda Runtimes

stack72:
Fixes: #610


stack72:
Going to self merge this - this follows the same pattern as that of deprecation of other properties

CyrusNajmabadi:
Hey @stack72 can you pin me on these reviews?  Thanks!

Issue #591, opened 2019-06-07
LambdaLayer doesn't accept src code folder.


lorecrafting:
Currently LambdaLayers only accepts a pointer to a .zip file deployment package through the filename property.  It would be nice to have a 'code' property just like normal Lambdas so that it can point an unzipped source folder for easier development.

stack72:
@lukehoban this is currently not supported functionality in the AWS SDK. The SDK input struct looks as follows:

```
type LayerVersionContentInput struct {
	_ struct{} `type:"structure"`

	// The Amazon S3 bucket of the layer archive.
	S3Bucket *string `min:"3" type:"string"`

	// The Amazon S3 key of the layer archive.
	S3Key *string `min:"1" type:"string"`

	// For versioned objects, the version of the layer archive object to use.
	S3ObjectVersion *string `min:"1" type:"string"`

	// The base64-encoded contents of the layer archive. AWS SDK and AWS CLI clients
	// handle the encoding for you.
	//
	// ZipFile is automatically base64 encoded/decoded by the SDK.
	ZipFile []byte `type:"blob" sensitive:"true"`
}
```

Do you think we can handle this a different way - i.e. have a process that will zip a folder?

lukehoban:
We want to do the same thing we do for `aws.lambda.Function`:

https://github.com/pulumi/pulumi-aws/blob/d49496d31e846e5c06d79b3d2b17d579b6350dc8/resources.go#L1270-L1289

Presumably just:

```go
                         "aws_lambda_layer_version": {
				Tok: awsResource(lambdaMod, "LayerVersion"),
				Fields: map[string]*tfbridge.SchemaInfo{
					"layer_name": tfbridge.AutoName("name", 64),
					// Terraform accepts two sources for lambdas: a local filename or a S3 bucket/object.  To bridge
					// with Pulumi's asset model, we will hijack the filename property.  A Pulumi archive is passed in
					// its stead and we will turn around and emit the archive as a temp file that Terraform can read.
					// We also automatically populate the asset hash property as this is used in diffs/updates/etc.
					"filename": {
						Name: "code",
						Asset: &tfbridge.AssetTranslation{
							Kind:      tfbridge.FileArchive,
							Format:    resource.ZIPArchive,
							HashField: "source_code_hash",
						},
					},
				},
			},
```

stack72:
@lukehoban ooohhhhh ok, I can try and put a reproduction together and try this!

PR #612, opened 2019-06-19
Adding the ability to pass code directly to lambda layer_version

stack72:
Fixes: #591

@lukehoban we should have a chat about this before we merge it - I am not quite sure that this has no unintended side-effects

### DIFF
--- a/examples/lambda-layer/index.ts
+++ b/examples/lambda-layer/index.ts
@@ -1,0 +1,24 @@
+// Copyright 2016-2017, Pulumi Corporation.  All rights reserved.
+
+import * as pulumi from "@pulumi/pulumi";
+import * as aws from "@pulumi/aws";
+import { asset } from "@pulumi/pulumi";
+
+const config = new pulumi.Config("aws");
+const providerOpts = { provider: new aws.Provider("prov", { region: <aws.Region>config.require("envRegion") }) };
+
+let layer = new aws.lambda.LayerVersion("my-layer-version", {
+    code: new asset.AssetArchive({
+        "index.js": new asset.StringAsset(
+            "exports.handler = (e, c, cb) => cb(null, {statusCode: 200, body: 'Hello, world!'});",
+        ),
+    }),
+    compatibleRuntimes: [
+        aws.lambda.NodeJS8d10Runtime,
+        aws.lambda.NodeJS10dXRuntime,
+    ],
+    layerName: "lambda_layer_name",
+}, providerOpts);
+
+export let layerSize = layer.sourceCodeSize;
+export let layerArn = layer.arn;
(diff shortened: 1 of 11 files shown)


examples/lambda-layer/index.ts
```diff
@@ -0,0 +1,18 @@
...
+
+let lambda = new aws.lambda.LayerVersion("my-layer-version", {
+    code: new asset.AssetArchive({
+        "index.js": new asset.StringAsset(
+            "exports.handler = (e, c, cb) => cb(null, {statusCode: 200, body: 'Hello, world!'});",
+        ),
+    }),
+    compatibleRuntimes: [
+        aws.lambda.NodeJS8d10Runtime,
+        aws.lambda.NodeJS8d10Runtime,
```
lukehoban: Should these two be different?

examples/lambda-layer/index.ts
```diff
@@ -0,0 +1,18 @@
...
+    ],
+    layerName: "lambda_layer_name",
+});
```
lukehoban: Is the goal of the test to create a lambda function using this layer?  Or just to create the Layer resource?

examples/lambda-layer/invoke.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+echo "Enter your Lambda's name (usually in the form mylambda-f0780fd95f4) followed by [ENTER]: "
+read LAMBDA
+echo "Attempting to invoke $LAMBDA ..."
+aws lambda invoke --function-name $LAMBDA --log-type Tail out.txt | jq '.LogResult' -r | base64 --decode
```
lukehoban: I don't think we use this file anymore - it was a very old artifact of back when these really were "examples".  And for the Layer in particular, this won't even work.

examples/lambda-layer/package.json
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "serverless-raw",
```
lukehoban: Nit - but this name probably should change.

examples/lambda-layer/index.ts
stack72: yes - good spot!

examples/lambda-layer/invoke.sh
stack72: gone - good spot!

examples/lambda-layer/package.json
stack72: done!

examples/lambda-layer/index.ts
stack72: The goal of the test is just to make sure that we can create the layer based on the new parameters - I have added some outputs to verify that the layer has been created as expected

stack72:
@CyrusNajmabadi FYI here ☝️

Issue #592, opened 2019-06-08
Add constants for RDS instance classes


jen20:
It would be nice to have an enumeration of valid RDS instance types such that we avoid magic strings in various places.

PR #613, opened 2019-06-19
Adding constants for RDS InstanceTypes

stack72:
Fixes: #592

### DIFF
--- a/sdk/nodejs/rds/instanceType.ts
+++ b/sdk/nodejs/rds/instanceType.ts
@@ -1,0 +1,134 @@
+// Copyright 2016-2018, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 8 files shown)


sdk/nodejs/rds/instanceType.ts
```diff
@@ -0,0 +1,133 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file contains two categories of exports:
+//
+//     1) A union type, InstanceType, that accepts any valid RDS instance type.
+//     2) Individual constants for each such RDS instance type.
+//
+// These just give a better developer experience and are just sugared strings.
```
lukehoban: Can we add a link to the primary source for this list?

stack72: Good idea - just added https://aws.amazon.com/rds/instance-types/ thanks for reminding me of that!

PR #616, opened 2019-06-19
Strongly-typed S3 bucket website routingRules

justinvp:
I've been working on a program that adds some redirects for the GREAT WEBSITE MIGRATION OF 2019, and wished there were stronger types for this. Decided to go ahead and add these while it was fresh in my mind...

This augments the `routingRules` property to accept a strongly-typed array of objects, which helps to know which properties can be set and avoids the need to use `JSON.stringify`. This is especially helpful as some properties can only be set if other properties are not set.

Depends on a change in `pulumi-terraform` to allow JSON arrays in `TransformJSONDocument`: https://github.com/pulumi/pulumi-terraform/pull/393

Details:
https://www.terraform.io/docs/providers/aws/r/s3_bucket.html#routing_rules
https://docs.aws.amazon.com/AmazonS3/latest/dev/how-to-page-redirect.html#advanced-conditional-redirects (scroll down for schema)

### DIFF
--- a/sdk/nodejs/s3/bucket.ts
+++ b/sdk/nodejs/s3/bucket.ts
@@ -6,6 +6,7 @@ import * as utilities from "../utilities";
 
 import {PolicyDocument} from "../iam/documents";
 import {CannedAcl} from "./cannedAcl";
+import {RoutingRule} from "./routingRules";
 
 /**
  * Provides a S3 bucket resource.
@@ -474,7 +475,7 @@ export interface BucketState {
     /**
      * A website object (documented below).
      */
-    readonly website?: pulumi.Input<{ errorDocument?: pulumi.Input<string>, indexDocument?: pulumi.Input<string>, redirectAllRequestsTo?: pulumi.Input<string>, routingRules?: pulumi.Input<string> }>;
+    readonly website?: pulumi.Input<{ errorDocument?: pulumi.Input<string>, indexDocument?: pulumi.Input<string>, redirectAllRequestsTo?: pulumi.Input<string>, routingRules?: pulumi.Input<string | RoutingRule[]> }>;
     /**
      * The domain of the website endpoint, if the bucket is configured with a website. If not, this will be an empty string. This is used to create Route 53 alias records.
      */
@@ -568,7 +569,7 @@ export interface BucketArgs {
     /**
      * A website object (documented below).
      */
-    readonly website?: pulumi.Input<{ errorDocument?: pulumi.Input<string>, indexDocument?: pulumi.Input<string>, redirectAllRequestsTo?: pulumi.Input<string>, routingRules?: pulumi.Input<string> }>;
+    readonly website?: pulumi.Input<{ errorDocument?: pulumi.Input<string>, indexDocument?: pulumi.Input<string>, redirectAllRequestsTo?: pulumi.Input<string>, routingRules?: pulumi.Input<string | RoutingRule[]> }>;
     /**
      * The domain of the website endpoint, if the bucket is configured with a website. If not, this will be an empty string. This is used to create Route 53 alias records.
      */
--- a/sdk/nodejs/s3/routingRules.ts
+++ b/sdk/nodejs/s3/routingRules.ts
@@ -1,0 +1,99 @@
+// Copyright 2016-2018, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * A rule that identifies a condition and the redirect that is applied when the condition is met.
+ * If a condition is not included, the rule is applied to all requests.
+ *
+ * For more details, please refer to the AWS documentation online:
+ * https://docs.aws.amazon.com/AmazonS3/latest/dev/how-to-page-redirect.html#advanced-conditional-redirects
+ */
+export interface RoutingRule {
+    /**
+     * A condition that must be met for the specified redirect to be applied.
+     * If not included, the rule is applied to all requests.
+     */
+    Condition?: Condition;
+
+    /**
+     * Provides instructions for redirecting the request. You can redirect requests to another host or another page,
+     * or you can specify another protocol to use.
+     */
+    Redirect: Redirect;
+}
+
+/**
+ * A condition that must be met for the specified redirect to be applied.
+ */
+export interface Condition {
+    /**
+     * The HTTP error code that must match for the redirect to apply. If an error occurs, and if the error code meets
+     * this value, then the specified redirect applies.
+     *
+     * `HttpErrorCodeReturnedEquals` is required if `KeyPrefixEquals` is not specified. If both `KeyPrefixEquals` and
+     * `HttpErrorCodeReturnedEquals` are specified, both must be true for the condition to be met.
+     */
+    HttpErrorCodeReturnedEquals?: string;
+
+    /**
+     * The prefix of the object key name from which requests are redirected.
+     *
+     * `KeyPrefixEquals` is required if `HttpErrorCodeReturnedEquals` is not specified. If both `KeyPrefixEquals` and
+     * `HttpErrorCodeReturnedEquals` are specified, both must be true for the condition to be met.
+     */
+    KeyPrefixEquals?: string;
+}
+
+/**
+ * Provides instructions for redirecting the request. You can redirect requests to another host or another page, or you
+ * can specify another protocol to use. At least one property must be set.
+ */
+export interface Redirect {
+    /**
+     * The hostname to be used in the Location header that is returned in the response.
+     *
+     * If another property is set, `HostName` is not required.
+     */
+    HostName?: string;
+
+    /**
+     * The HTTP redirect code to be used in the Location header that is returned in the response.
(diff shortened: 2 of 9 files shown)


sdk/nodejs/s3/routingRules.ts
```diff
@@ -0,0 +1,182 @@
...
+     */
+    Condition?: Condition;
+
+    /**
+     * Provides instructions for redirecting the request. You can redirect requests to another host or another page, or you can specify another protocol to use.
+     */
+    Redirect: Redirect;
+}
+
+// Multiple types are used for `Condition` to enforce at least one property being set.
```
justinvp: Is there a more succinct way to express this in TypeScript while still providing nice JSDoc comments?

sdk/nodejs/s3/routingRules.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     * The prefix of the object key name from which requests are redirected.
+     */
+    KeyPrefixEquals: string;
+}
+
+// Multiple types are used for `Redirect` to enforce at least one property being set.
+// If `ReplaceKeyPrefixWith` is set, then `ReplaceKeyWith` cannot be set.
+// If `ReplaceKeyWith` is set, then `ReplaceKeyPrefixWith` cannot be set.
```
justinvp: Is there a more succinct way to express this in TypeScript while still providing nice JSDoc comments?

sdk/nodejs/s3/bucket.ts
```diff
@@ -474,7 +475,7 @@ export interface BucketState {
     /**
      * A website object (documented below).
      */
-    readonly website?: pulumi.Input<{ errorDocument?: pulumi.Input<string>, indexDocument?: pulumi.Input<string>, redirectAllRequestsTo?: pulumi.Input<string>, routingRules?: pulumi.Input<string> }>;
+    readonly website?: pulumi.Input<{ errorDocument?: pulumi.Input<string>, indexDocument?: pulumi.Input<string>, redirectAllRequestsTo?: pulumi.Input<string>, routingRules?: pulumi.Input<string | RoutingRules> }>;
```
justinvp: Ideally, I'd like this to be expressed as:

```typescript
routingRules?: pulumi.Input<string | RoutingRule[]>
```

Rather than:

```typescript
routingRules?: pulumi.Input<string | RoutingRules>

// with the following from routingRules.ts:
export type RoutingRules = RoutingRule[];
```

But I had to create the `RoutingRules` placeholder type because I couldn't express `"RoutingRule[]"` as an `AltTypes` in the tfbridge.SchemaInfo`.

When trying to do so:

```go
AltTypes:  []tokens.Type{awsType(s3Mod+"/routingRules", "RoutingRule[]")},
```

It tries to import the array (which isn't valid TypeScript):

```typescript
import {RoutingRule[]} from "./routingRules";
```

(It should just import `RoutingRule`).

I could make a targeted change in `pulumi-terraform`'s `generate_nodejs.go` to strip the `[]` suffix when importing the type.

https://github.com/pulumi/pulumi-terraform/blob/2bd38b436a69b4056ae58ce8accc0dd5ff63a980/pkg/tfgen/generate_nodejs.go#L959

Changing this:

```go
imports[relmod][string(ct.Name())] = true
```

To this:

```go
imports[relmod][strings.TrimSuffix(string(ct.Name()), "[]")] = true
```

Trying out that change appears to work perfectly, but I'm slightly hesitant to do so as `ct.Name()` is supposed to conform to the `[A-Za-z_.][A-Za-z0-9_.]*` regex, even though the `[]` suffix appears to be getting through when I set it in the `AltTypes`.

lukehoban: Honestly, `RoutingRules` seems fine to me. 

But if we can fix this in `pulumi-Terraform` to allow `[]` that’s great too.

justinvp: I'll just fix `pulumi-terraform` to all `[]` so we don't have to add unnecessary types that just alias the array. I'll add a commit to https://github.com/pulumi/pulumi-terraform/pull/393

sdk/nodejs/s3/routingRules.ts
```diff
@@ -0,0 +1,182 @@
...
+     * If specified, both `HttpErrorCodeReturnedEquals` and `KeyPrefixEquals` must be true for the condition to be met.
+     */
+    KeyPrefixEquals?: string;
+}
+
+export interface KeyPrefixEqualsCondition {
+    /**
+     * The HTTP error code that must match for the redirect to apply. If an error occurs, and if the error code meets this value, then the specified redirect applies.
+     *
+     * If specified, both `HttpErrorCodeReturnedEquals` and `KeyPrefixEquals` must be true for the condition to be met.
```
CyrusNajmabadi: i'm confused what you mean by 'true' here since these are strings.

sdk/nodejs/s3/routingRules.ts
```diff
@@ -0,0 +1,182 @@
...
+     */
+    HttpErrorCodeReturnedEquals?: string;
+
+    /**
+     * The prefix of the object key name from which requests are redirected.
+     */
+    KeyPrefixEquals: string;
+}
+
+// Multiple types are used for `Redirect` to enforce at least one property being set.
```
CyrusNajmabadi: this feels contorted to me.  i woudl just make optional and have a runtime check TBH.

CyrusNajmabadi:
TBH, it feels a little overboard on trying to be typesafe instead of just having a single bag with a few optional properties that we just check that at least one is set...

sdk/nodejs/s3/routingRules.ts
justinvp: These are routing conditions. The conditions are defined by strings, and depending on the route to the bucket, they'll either be true or false.

This text comes directly from the AWS docs at https://docs.aws.amazon.com/AmazonS3/latest/dev/how-to-page-redirect.html#advanced-conditional-redirects (scroll down a bit).

sdk/nodejs/s3/routingRules.ts
justinvp: Would this be a runtime check we implement (if so how?), or just leave it up to AWS to do any checking?

CyrusNajmabadi: Yes.  I twould be something like `if (a.x === undefined && a.y === undefined && a.z === undefined) { throw new ResourceError("At least one of blah blah blah must be set"); }`

justinvp: Yeah, but where would such a check go? Is there a place on the TS/JS side to put it?

I suppose I could write it on the Go side... instead of defining the field as:

```go
Fields: map[string]*tfbridge.SchemaInfo{
        "routing_rules": {
            Name:      "routingRules",
            Type:      "string",
            AltTypes:  []tokens.Type{awsType(s3Mod+"/routingRules", "RoutingRules")},
            Transform: tfbridge.TransformJSONDocument,
        },
    },
```

I could write a custom `Transform` that wraps the call to `TransformJSONDocument` and adds custom validation, e.g. something like:

```go
Transform: func(val resource.PropertyValue) (resource.PropertyValue, error) {
	result, err := tfbridge.TransformJSONDocument(val)
	if err != nil {
		return resource.PropertyValue{}, err
	}

	// validate val here

	// if ok
	return result, nil
},
```

But that seems kind of cumbersome.

justinvp: @CyrusNajmabadi?

CyrusNajmabadi: Answered offline.

justinvp:
Feedback addressed (using simpler TypeScript types and just documenting that at least one property must be set). Cleaned up the commit and adding one to use the latest `pulumi-terraform` with the necessary changes.

justinvp:
Travis PR test failure:

```
[ umi/pulumi-aws/examples/bucket ] Invoke '/home/travis/.pulumi/bin/pulumi preview --non-interactive' failed: exit status 255

[ umi/pulumi-aws/examples/bucket ] Previewing update (moolumi/p-it-travis-job-bucket-08395507):

[ umi/pulumi-aws/examples/bucket ] 

[ umi/pulumi-aws/examples/bucket ]  +  pulumi:pulumi:Stack bucket-p-it-travis-job-bucket-08395507 create 

[ umi/pulumi-aws/examples/bucket ] @ Previewing update.......

[ umi/pulumi-aws/examples/bucket ]  +  pulumi:providers:aws prov create 

[ umi/pulumi-aws/examples/bucket ]     aws:s3:Bucket websiteBucket  error: could not validate provider configuration: 1 error occurred:

[ umi/pulumi-aws/examples/bucket ]     aws:s3:Bucket websiteBucket  1 error

[ umi/pulumi-aws/examples/bucket ]  

[ umi/pulumi-aws/examples/bucket ] Diagnostics:

[ umi/pulumi-aws/examples/bucket ]   aws:s3:Bucket (websiteBucket):

[ umi/pulumi-aws/examples/bucket ]     error: could not validate provider configuration: 1 error occurred:

[ umi/pulumi-aws/examples/bucket ]     	* : invalid or unknown key: env_region

[ umi/pulumi-aws/examples/bucket ]  

[ umi/pulumi-aws/examples/bucket ] Permalink: https://app.pulumi.com/moolumi/bucket/p-it-travis-job-bucket-08395507/previews/c95af275-a963-49b2-b85f-7c1e0f427f41
```

@CyrusNajmabadi, this looks related to the change you made in #614. Is this a bug or just a point-in-time thing that will be fixed-up subsequently?

CyrusNajmabadi:
@justinvp No.  you'll need to update any test you've created to make and pass along an explicit provider resource.  This is intentional so we can validate all our code works with non-default providers.  I can provide more info if you need it. However, you can also look at our other tests to see how this is done.

justinvp:
@stack72, this should be ready for a final review, FYI, if you'd like to take a look.

stack72:
This LGTM! Thanks for this change @justinvp - a lot of interaction here for you to have had to drive though so 🥇

PR #624, opened 2019-06-21
Update changelog.

CyrusNajmabadi:


ellismg:
I think that James and Paul have been using some tool to manage these updates and that does something slightly different for the formatting so it can be managed by a tool.

This seems fine if we really need to get the changes out today, but ideally we’d do this the right way and write down how this stuff is managed (maybe in the contributing file?)

CyrusNajmabadi:
@jen20 Any thoughts on the above?

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
+
+## 0.18.11 (2019-06-20)
+
```
lukehoban: I assume this needs to be `0.18.12`?

PR #655, opened 2019-07-11
Upgrade to v2.19.0 of the AWS Terraform Provider

stack72:


sdk/nodejs/apigateway/stage.ts
```diff
@@ -7,57 +7,6 @@ import * as utilities from "../utilities";
...
- *     settings: {
- *         loggingLevel: "INFO",
- *         metricsEnabled: true,
- *     },
- *     stageName: testStage.stageName,
- * });
- * ```
- *
- * > This content is derived from https://github.com/terraform-providers/terraform-provider-aws/blob/master/website/docs/r/api_gateway_stage.html.markdown.
- */
```
CyrusNajmabadi: why did we lose this?

PR #659, opened 2019-07-17
Add support for RepositoryCredentials in ecs ContainerDefinition

burnzy:
Added support for RepositoryCredentials in ECS container definitions based on https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_RepositoryCredentials.html


stack72:
Hi @burnzy 	

Thanks so much for the PR here - please can you share some details on how you tested this and the result? This will help me to get this merged

Thanks

Paul

stack72:
ping @burnzy :)

lukehoban:
@CyrusNajmabadi is this a newly added field of this interface in AWS?  Are there others we are missing?  What’s the best way to audit these?

burnzy:
Sorry for delay. Yes, this change is for supporting private repository credentials for fargate tasks in ecs. AWS added this feature in [September 2018](https://aws.amazon.com/blogs/compute/introducing-private-registry-authentication-support-for-aws-fargate/). The new field was added to the [container definition](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerDefinition.html).

I just updated the ContainerDefinition interface to support this, using the same format that the previous fields were added.

lukehoban:
I'm actually a little confused why this interface is even here.  It doesn't appear to be used by any code in this library.  Is it just used by code in `awsx`?

I assume we *should* really be adding an `AltType` to `containerDefinitions` on `aws.ecs.TaskDefinition` that allows it to be an array of these instead of a raw string?

As for adding to it - sounds like it is more or less locked in time from when it was first created in early 2018.  We should do a pass to update it fully.  

But this change by itself looks good.

burnzy:
@lukehoban Yes true, I noticed the issue while using `awsx.ecs.FargateTaskDefinition`

PR #665, opened 2019-07-21
Change iam PolicyDocument to accept Input parameters

stack72:
In order to allow the passing of arguments from other resource
outputs, we should change how PolicyDocument works to allow passing
an Input<string> rather than a string. This will help our users to
avoid any unnecessary interpolation of types

### DIFF
--- a/sdk/nodejs/iam/documents.ts
+++ b/sdk/nodejs/iam/documents.ts
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import {Input} from "@pulumi/pulumi";
+
 /**
  * You manage access in AWS by creating policies and attaching them to IAM identities or AWS resources. A policy is an
  * object in AWS that, when associated with an entity or resource, defines their permissions. AWS evaluates these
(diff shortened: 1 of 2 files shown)


sdk/nodejs/iam/documents.ts
```diff
@@ -44,9 +46,9 @@
  */
 export interface PolicyDocument {
     // The version of the policy language that you want to use. As a best practice, use the latest `2012-10-17` version.
-    Version: "2008-10-17" | "2012-10-17";
+    Version: Input<"2008-10-17" | "2012-10-17">;
     // An optional document ID.
-    Id?: string;
+    Id?: Input<string>;
     // One or more policy statements, describing the effect, principal, action, resource, and condition.
     Statement: PolicyStatement[];
```
CyrusNajmabadi: shoudl we also update this to `Input<Input<PolicyStatement>[]>`

sdk/nodejs/iam/documents.ts
```diff
@@ -68,14 +70,14 @@ export interface PolicyStatement {
...
     // Include a list of actions that the policy allows or denies.
-    Action?: string | string[];
+    Action?: Input<string> | Input<string>[];
     // Include a list of actions that are not covered by this policy.
-    NotAction?: string | string[];
+    NotAction?: Input<string> | Input<string>[];
 
     // A list of resources to which the actions apply.
-    Resource?: string | string[];
+    Resource?: Input<string> | Input<string>[];
```
CyrusNajmabadi: this one is crazier in that it should also be wrapped with a higher level `Input<...>`  (SAme with the guys above it).

sdk/nodejs/iam/documents.ts
```diff
@@ -68,14 +70,14 @@ export interface PolicyStatement {
     NotPrincipal?: Principal;
```
CyrusNajmabadi: these two shoudl likely be inputs as well.

stack72:
@CyrusNajmabadi requested changes made!

CyrusNajmabadi:
LGTM!  merge when you want!

PR #669, opened 2019-07-22
WIP - Expose the elasticloadbalancingv2 module through the elbv2 name.

CyrusNajmabadi:


sdk/python/pulumi_aws/__init__.py
```diff
@@ -24,4 +24,5 @@
 from .get_partition import *
 from .get_prefix_list import *
 from .get_region import *
+from .sdk/python/pulumi_aws/extra_imports.py import *
```
CyrusNajmabadi: @pgavlin @lukehoban  i don't think destFiles overlays work for python unfortunately.  will have to figure out something here. Or, we can just totally remove applicationloadbalancing and accept the breaking change.

CyrusNajmabadi:
Tabling this.  Have some definite issues that aren't clear how to deal with cleanly.

CyrusNajmabadi:
Closing out.  We accomplished this a different way.

PR #674, opened 2019-07-25
Update pulumi-terraform to fdafc00b3f

stack72:


stack72:
Going to self merge this - nothing in the SDK changed

PR #675, opened 2019-07-25
[WIP] - Rename aws sdk modules to match terraform equivalents.

CyrusNajmabadi:
Note: this is a WIP.  We need to inject aliases into the new resources so that someone can move transparently from the old ones to the new ones.  (there will also be a test for this).


CyrusNajmabadi:
Moving forward with this.  Works great in practice :)  Built off of the already reviewed pulumi/pulumi and pulumi/terraform bits.

PR #681, opened 2019-07-30
Add support for new Middle East Region in location constants

stack72:


sdk/nodejs/region.ts
```diff
@@ -27,6 +27,7 @@ export let USEast1Region: Region      = "us-east-1";
 export let USEast2Region: Region      = "us-east-2";
 export let USWest1Region: Region      = "us-west-1";
 export let USWest2Region: Region      = "us-west-2";
+export let MESouth1: Region           = "me-south-1";
```
mikhailshilkov: Consider keeping the alphabetic order of the regions

stack72: Done!

mikhailshilkov:
Do we also miss `me-south-1` `cn-north-1` `cn-northwest-1` `eu-north-1`?

stack72:
> Do we also miss `me-south-1` `cn-north-1` `cn-northwest-1` `eu-north-1`?

This is me-south-1 but I will add the others now! Good find :)

PR #692, opened 2019-08-07
Upgrade to v2.23.0 of the AWS Terraform Provider

stack72:


jen20:
Does the test parallelism still need to be reduced now that the other issues are fixed?

stack72:
@jen20 I am going to leave at 10 for a little bit and test that out on another PR to see if it's still a problem

PR #693, opened 2019-08-07
[WIP] Support for capturing secrets in CallbackFunction

lukehoban:
Secrets are captured as environment variables instead of being serialied into program text.  This ensures they are encrypted at rest, do not show up in the state file, and are decrypted and provided to the running Lambda function via environment variables function implicitly by the Lambda runtime infrastructure.

An alternative we are considering, which could be added as opt-in in the future, is to go a step further and create SecretsManager secrets for each captured secret, then inject code to decrypt these using SecretsManager at runtime inside the Lambda.  This involves adding non-trivial additional infrastructure, but would provide an even stronger level of protection (for example, the secrets would not be visible in the Lambda console, and would not be decrypted until/unless actually accessed within the Lambda function).

Part of https://github.com/pulumi/pulumi/issues/2718.

### DIFF
--- a/examples/callbackfunction-secrets/Pulumi.yaml
+++ b/examples/callbackfunction-secrets/Pulumi.yaml
@@ -1,0 +1,4 @@
+name: callbackfunction-secrets
+description: Capturing secrets in a CallbackFunction
+runtime: nodejs
+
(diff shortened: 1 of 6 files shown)


lukehoban:
Builds on https://github.com/pulumi/pulumi/pull/3050.

lukehoban:
This was addressed in #1282 and #1525. Closing this one out.

PR #695, opened 2019-08-09
Upgrade to pulumi-terraform 9db2fc93cd

stack72:


stack72:
Self merging this as we need to get this through - this unblocks all of the Python work and the issues we are seeing in master

PR #696, opened 2019-08-09
Example of what docs look like with synchronous data sources

CyrusNajmabadi:
Not actually for merge.  Just serves to show what https://github.com/pulumi/tf2pulumi/pull/107 ends up producing.

### DIFF
--- a/sdk/nodejs/ec2/getSubnetIds.ts
+++ b/sdk/nodejs/ec2/getSubnetIds.ts
@@ -17,15 +17,13 @@ import * as utilities from "../utilities";
  * import * as pulumi from "@pulumi/pulumi";
  * import * as aws from "@pulumi/aws";
  * 
- * const exampleSubnetIds = pulumi.output(aws.ec2.getSubnetIds({
+ * const exampleSubnetIds = aws.ec2.getSubnetIds({
  *     vpcId: var_vpc_id,
- * }));
+ * });
  * const exampleSubnet: Output<aws.ec2.GETSUBNETResult>[] = [];
  * for (let i = 0; i < exampleSubnetIds.apply(exampleSubnetIds => exampleSubnetIds.ids.length); i++) {
  *     exampleSubnet.push(aws.ec2.getSubnet);
- * %!(EXTRA string=exampleSubnetIds.apply(exampleSubnetIds => aws.ec2.getSubnet({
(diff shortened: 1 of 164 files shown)


sdk/nodejs/ec2/getSubnetIds.ts
```diff
@@ -17,15 +17,13 @@ import * as utilities from "../utilities";
...
  *     vpcId: var_vpc_id,
- * }));
+ * });
  * const exampleSubnet: Output<aws.ec2.GETSUBNETResult>[] = [];
  * for (let i = 0; i < exampleSubnetIds.apply(exampleSubnetIds => exampleSubnetIds.ids.length); i++) {
  *     exampleSubnet.push(aws.ec2.getSubnet);
- * %!(EXTRA string=exampleSubnetIds.apply(exampleSubnetIds => aws.ec2.getSubnet({
- *         id: exampleSubnetIds.ids[i],
- *     })))}
+ * }
```
CyrusNajmabadi: note: both the before/after here are somewhat borked (and i'll file issue on it).  The before was borked with that `%!(Extra...` bit.  The before/after is **still** borked in that:

1. we do `exampleSubnetIds.apply(exampleSubnetIds => exampleSubnetIds.ids.length)` instead of `exampleSubnetIds.ids.length`
2. we do `exampleSubnet.push(aws.ec2.getSubnet)` instead of `exampleSubnet.push(aws.ec2.getSubnet({ id: exampleSubnetIds.ids[i] })`
3. We don't do `for ... of` (though that would be *very* fancy).

sdk/nodejs/ec2/getSubnetIds.ts
```diff
@@ -17,15 +17,13 @@ import * as utilities from "../utilities";
  * import * as pulumi from "@pulumi/pulumi";
  * import * as aws from "@pulumi/aws";
  * 
- * const exampleSubnetIds = pulumi.output(aws.ec2.getSubnetIds({
+ * const exampleSubnetIds = aws.ec2.getSubnetIds({
  *     vpcId: var_vpc_id,
- * }));
+ * });
  * const exampleSubnet: Output<aws.ec2.GETSUBNETResult>[] = [];
```
CyrusNajmabadi: this line is also bad.  where did `GETSUBNETResult` come from instead of `GetSubnetResult`?

Issue #698, opened 2019-08-14
No support for t3a.* instance types


armenr:
Looks like the t3a.* instance types aren't supported?

## Code:
```
myProjVPC.publicSubnetIds.forEach((server, idx) => {
  const webHost = new aws.ec2.Instance(nameFactory(`web-${idx + 1}`), {
    ami: `${config.DEFAULT_AMI}`,
    instanceType: 't3a.medium',
    keyName: `${config.ENV_PEM_KEY}`,
    subnetId: myProjVPC.publicSubnets[idx].subnet.id,
    availabilityZone: myProjVPC.publicSubnets[idx].subnet.availabilityZone,
    vpcSecurityGroupIds: [sgImpactHubAccess.id, sgEC2Container.id, loadBalancerToInstance.id],

    userData: userDataPythonScript,
    tags: {
      Name: `${config.ENV_STAGE}` + '-' + `web-${idx + 1}`,
      Environment: `${config.ENV_STAGE}`,
      InfraVersion: `${config.INFRA_VERSION}`,
    },
  });
```

## Error
```
    error: Running program '/Users/armenr/Desktop/myProj-new/myProj-tessitura' failed with an unhandled exception:
    TSError: ⨯ Unable to compile TypeScript:
    pulumi/src/index.ts(213,3): error TS2322: Type '"t3a.medium"' is not assignable to type 'Input<InstanceType>'.

        at createTSError (/Users/armenr/Desktop/myProj-new/myProj-tessitura/node_modules/ts-node/src/index.ts:261:12)
        at getOutput (/Users/armenr/Desktop/myProj-new/myProj-tessitura/node_modules/ts-node/src/index.ts:367:40)
        at Object.compile (/Users/armenr/Desktop/myProj-new/myProj-tessitura/node_modules/ts-node/src/index.ts:558:11)
        at Module.m._compile (/Users/armenr/Desktop/myProj-new/myProj-tessitura/node_modules/ts-node/src/index.ts:439:43)
        at Module._extensions..js (internal/modules/cjs/loader.js:879:10)
        at Object.require.extensions.<computed> [as .ts] (/Users/armenr/Desktop/myProj-new/myProj-tessitura/node_modules/ts-node/src/index.ts:442:12)
        at Module.load (internal/modules/cjs/loader.js:731:32)
        at Function.Module._load (internal/modules/cjs/loader.js:644:12)
        at Module.require (internal/modules/cjs/loader.js:771:19)
[...]

PR #699, opened 2019-08-14
Add support for T3a instance types

stack72:
Fixes: #698

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-___NULL___
+* Add support for `t3a` instance types
 
 ---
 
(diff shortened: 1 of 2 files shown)


mikhailshilkov:
LGTM.

lukehoban:
I assume there are lots of other instances missing?  Should we do a more complete pass to bring this up to date?

In general - given that missing values in these enums  make it *seem* like the missing options are unsupported - seems like we need to put some system or process in place to keep them up to date on at least a ~monthly basis. Same applies to all other enums.

stack72:
@lukehoban yes, I will be doing this as soon as I have fixed up some of the things for 0.26

We need to investigate a better way of managing this

lukehoban:
I continue to believe we should automate or semi-automate generation of these enums from primary sources.  

For examples - instances are available here: https://github.com/aws/aws-sdk-go/blob/master/models/apis/ec2/2016-11-15/api-2.json#L15131

stack72:
@lukehoban I 100% agree with you!

PR #706, opened 2019-08-15
Upgrade to v2.24.0 of the upstream provider

jen20:
This is waiting for some other changes to land before being ready to merge.


stack72:
@jen20 	this LGTM BUT there is an issue with the changelog - you will need to update that from master as it looks like your code is missing a tonne of CHANGELOG entries

stack72:
I think there was something wrong with the branch here - superseded by #714

PR #707, opened 2019-08-15
Allow IAM document Conditions to be arrays

### LINKED ISSUE #705: IAM PolicyDocument ConditionArguments needs to allow Array of values in addition to scalar value
IAM and Resource policy allow for certain condition arguments that can evaluate multiple values in a single condition statement.

2 examples are aws:SourceIp like
```
"Condition": {
      "IpAddress": {
        "aws:SourceIp": [
          "203.0.113.0/24",
          "2001:DB8:1234:5678::/64"
        ]
      }
    }
```
And s3:prefix
```
"Condition": {
                "StringEquals": {
                    "s3:prefix": [
                        "",
                        "home/"
                    ],
                    "s3:delimiter": [
                        "/"
                    ]
                }
            }
```

To allow for this the ConditionArguments type here https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/iam/documents.ts#L115 would need to be expanded to allow for array values in addition to the current scalar support.

lukehoban:
Fixes #705.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 * Add support for `t3a` instance types
+* Allow IAM Policy Document `Condition`s to be arrays as well as scalars
 
 ---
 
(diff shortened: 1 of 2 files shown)


lukehoban:
Travis failures are unrelated - so going ahead and merging as-is.

PR #709, opened 2019-08-16
Upgrade to pulumi-terraform 14e049b09d

stack72:


sdk/nodejs/acm/certificateValidation.ts
```diff
@@ -27,10 +29,10 @@ import * as utilities from "../utilities";
  *     domainName: "example.com",
  *     validationMethod: "DNS",
  * });
- * const zone = pulumi.output(aws.route53.getZone({
+ * const zone = aws.route53.getZone({
```
justinvp: @pgavlin, are these changes expected? Are they from the newer `tf2pulumi`?

pgavlin: Yes.

PR #711, opened 2019-08-20
Addition of the work to test provider upgrades during tests

stack72:
Related to: https://github.com/pulumi/pulumi/issues/2757


examples/examples_test.go
```diff
@@ -57,33 +57,74 @@ func TestExamples(t *testing.T) {
 
 	shortTests := []integration.ProgramTestOptions{
 		baseJS.With(integration.ProgramTestOptions{Dir: path.Join(cwd, "minimal")}),
-		baseJS.With(integration.ProgramTestOptions{Dir: path.Join(cwd, "express")}),
+		baseJS.With(integration.ProgramTestOptions{
```
ellismg: Could/Should we add `RunUpdateTest: true` to `baseJS` and `basePython` so we don't have to utter this everywhere (and when we add new tests we get it "by default"?

stack72: I will test this out locally and then push another commit if it works out for us - the issue might be with the minimal test but we will see :) Going to merge for now and then come back to it but I love the suggestion

PR #730, opened 2019-09-03
Switch to GOPROXY to save problems with go module sources

stack72:


build/common.mk
```diff
@@ -180,8 +181,13 @@ all:: build install lint test_all
 
 ensure::
 	$(call STEP_MESSAGE)
-	@if [ -e 'Gopkg.toml' ]; then echo "dep ensure -v"; dep ensure -v; \
-		elif [ -e 'go.mod' ]; then echo "GO111MODULE=on go mod vendor"; GO111MODULE=on go mod vendor; gomod-doccopy -provider terraform-provider-$(PACK); fi
```
lukehoban: Is it safe to remove this from `common.mk` generally?  (Do we have any repos using `common.mk` that need this still?).

I believe goal remains to keep `common.mk` in-sync across all repors?  @ellismg ?

stack72:
Closed via #731

build/common.mk
ellismg: I believe that everything is now on go modules, so removing the dep related stuff seems reasonable.  We should push this everywhere, however.

PR #739, opened 2019-09-09
Allow ec2 instanceType to be passed as a string

stack72:
This continually trips our users up so we should allow both
string and the InstanceType constants

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 * Upgrade to 1.0 of pulumi SDK
+* Allow ec2 InstanceTypes to be passed as a string as well as the typed value
 
 ---
 
--- a/resources.go
+++ b/resources.go
@@ -769,7 +769,8 @@ func Provider() tfbridge.ProviderInfo {
 						AltTypes: []tokens.Type{awsType(iamMod, "InstanceProfile")},
 					},
 					"instance_type": {
-						Type: awsType(ec2Mod+"/instanceType", "InstanceType"),
+						Type:     "string",
+						AltTypes: []tokens.Type{awsType(ec2Mod+"/instanceType", "InstanceType")},
 					},
 				},
 			},
--- a/sdk/nodejs/ec2/instance.ts
+++ b/sdk/nodejs/ec2/instance.ts
@@ -152,7 +152,7 @@ export class Instance extends pulumi.CustomResource {
     /**
      * The type of instance to start. Updates to this field will trigger a stop/start of the EC2 instance.
      */
-    public readonly instanceType!: pulumi.Output<InstanceType>;
+    public readonly instanceType!: pulumi.Output<string>;
     public readonly ipv6AddressCount!: pulumi.Output<number>;
     /**
      * Specify one or more IPv6 addresses from the range of the subnet to associate with the primary network interface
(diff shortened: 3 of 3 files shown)


lukehoban:
We should decide on and the. Consistently implement our design guidelines here. 

In the past, we decided that:
1. AltTypes should always be alternatives, and should not change the underlying type
2. Except for enums where there is arguably no value is `Enum | string` since it is the exact same thing as `string`. 

I think we’ve decided that (2) is wrong, and that it leads to user pain and breaking changes. That seems reasonable to me. Cc also @cyrusnajmabadi and @pgavlin for thoughts. 

We should start adding notes on these design considerations to a wiki or doc in the Pulumi-Terraform repo. We should also make this change “everywhere” if we are going to make it at all.

Dominik-K:
Why not having a `ParseInstanceType()` function instead which also helps to fail fast if the string contains a wrong instance type? Just found out that you already have the [full instance types list](https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/ec2/instanceType.ts#L186) in code.

I really like that with Pulumi it's easier that fields gets stronger types or types at all.

cowwoc:
To be fair, this issue only tripped me up because I'm new to Typescript. I prefer more strong-typing rather than less. Had I known how to cast I would have just done so :)

cowwoc:
@Dominik-K A simple cast will work:

```
let MACHINE_TYPE = "t2.small";
const machineTypeAsString = process.env.MACHINE_TYPE as string;
if (machineTypeAsString)
	MACHINE_TYPE = <InstanceType>machineTypeAsString;
```

CyrusNajmabadi:
I'm really on the fence.  I actually like how strict TS is here and how it can catch real bugs.  It's really easy for users to force TS to be happy too (in the case where they truly have a string).  Is that not sufficient?  I'd prefer to stick with the status quo here a bit and hope that users can easily get through these TS warnings.

lukehoban:
@cyrusnajmabadi fwiw - my biggest concern with the status quo personally is that it makes it a breaking change to add these typings. Which means we probably can’t add anymore. And given we have very few, that means we basically can’t have this feature as a general thing. Or we have to decide this doesn’t count as a breaking change even though it will in practice require code changes to user code. 

What I’m unsure of is whether there is actually any value in `enum | string`. Like does this help intellisense at all?  Or is it truly just an inconvenient way of saying `string`?

CyrusNajmabadi:
> Like does this help intellisense at all? 

It doesn't seem to at all (afaict).  TS just collapses it down to `string`.  You don't even see the enum type in things like sighelp/quickinfo.  It basically gets (mostly) erased.

> Or we have to decide this doesn’t count as a breaking change even though it will in practice require code changes to user code.

That's the part I'm more amenable to.  I agree it's a break.  But one that is in the 'very minor' category.  Importantly, it's not a binary/runtime break.  Indeed, it's more of a source-warning versus a break.  I do like the idea of our APIs being able to be more precise in the future even if that has the chance of impacting someone.  

Importantly, i view this specific type of TS change as one that is easy to absorb.  i.e. if a user does run into a problem here a simple `<EnumType>` or `as EnumType` or `<any>` cast gets them into a good place immediately in the small handful of cases they'll run into.  Contrast this with, say, a change to Input/Output which has an extremely high chance of affecting people *everywhere*.

I strongly am behind the "be very careful about breaking changes" position we have.  However, for my personal bar, the "using a string-union type" goes into the acceptable category for me.

stack72:
Closed via #876

PR #749, opened 2019-09-19
Fix docs examples for CertificateValidation

danielrbradley:
Manually specify the record name when doing DNS validation, otherwise it will auto-generate the record name, instead of using the name required for the DNS validation.

### DIFF
--- a/sdk/nodejs/acm/certificateValidation.ts
+++ b/sdk/nodejs/acm/certificateValidation.ts
@@ -34,6 +34,7 @@ import * as utilities from "../utilities";
  *     privateZone: false,
  * });
  * const certValidation = new aws.route53.Record("certValidation", {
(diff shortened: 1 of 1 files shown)


stack72:
Hi @danielrbradley 	

Thanks for pointing this out - unfortunately, these files are autogenerated and we can't merge PRs that change these files

You are correct about the names being autogenerated - this is part of our implementation details in Pulumi - https://www.pulumi.com/docs/intro/concepts/programming-model/#autonaming

I hope you understand why this is the case

Paul

danielrbradley:
No worries @stack72, could you point me in the direction of where these files are generated from and I'll do a new PR based on that?

clstokes:
Related to https://github.com/pulumi/tf2pulumi/issues/117

PR #766, opened 2019-10-11
Examples: Change the examples to run as individual tests

stack72:


examples/examples_test.go
```diff
@@ -3,126 +3,215 @@
...
+			Dir: path.Join(getCwd(t), "minimal"),
+		})
 
-	cwd, err := os.Getwd()
-	if !assert.NoError(t, err, "expected a valid working directory: %v", err) {
-		return
-	}
+	t.Run(test.Dir, func(t *testing.T) {
+		integration.ProgramTest(t, &test)
+	})
```
lukehoban: We don’t actually need this `t.Run` anymore - right?  That causes this to be a child test, but we don’t need that as there is just one test?

stack72: Ah you are correct - sorry, my misunderstanding! Just pushed an update to this to remove those t.Runs

PR #776, opened 2019-10-21
Overriding default of Managed by Terraform entries

### LINKED ISSUE #775: The default description of rds/mysql is "Managed by Terraform"
The default description of rds/mysql is **Managed by Terraform**

stack72:
Fixes #775

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+* Update default `descriptions` to `Managed by Pulumi`. This may cause a diff for users.
 
 ---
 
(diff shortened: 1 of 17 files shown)


lukehoban:
This will cause diffs on all these resources. 

1. Are we sure none of these are ForceNew?
2. Are we okay with updates to a new version of this package causing updates to all of these resources even if not replacements?

stack72:
@lukehoban I went through the list and removed the items classed as "ForceNew". I think it would be better for us to accept that it *might* cause a user change. I think it's confusing for users right now - as noted in #775

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+* Update default `descriptions` to `Managed by Pulumi`. This may cause a diff for users.
```
lukehoban:
```suggestion
* Update default `descriptions` for several resources to `Managed by Pulumi`.
```

Issue #774, opened 2019-10-21
Update the default Lambda runtime to Node 10.x


mikhailshilkov:
Node.js 8.10 will reach EOL end of this year:

> The Node community has decided to end support for Node.js 8.x on December 31, 2019. From this date forward, Node.js 8.x will stop receiving bug fixes, security updates, and/or performance improvements. To ensure that your new and existing functions run on a supported and secure runtime, language runtimes that have reached their EOL are deprecated in AWS.
> For Node.js 8.x, there will be 2 stages to the runtime deprecation process:
> 1. Disable Function Create – Beginning January 6, 2020, customers will no longer be able to create functions using Node.js 8.10
> 2. Disable Function Update – Beginning February 3, 2020, customers will no longer be able to update functions using Node.js 8.10
> After this period, both function creation and updates will be disabled permanently. However, existing Node 8.x functions will still be available to process invocation events.

As per email communication ([this page](https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html) doesn't seem to reflect it yet).

`pulumi-aws` should now default to Node 10.x for Callback Functions.

Probably, the docs should reflect this too, e.g. [here](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/aws/lambda/#NodeJS8d10Runtime)

stack72:
@mikhailshilkov so the clarify the work you want done here:

1. Mark 8.10 as deprecated and add a message as to why
2. Set the default to be 10.x - we currently don't set a default in Pulumi - but I am presuming we can do that

mikhailshilkov:
@stack72 Yes, correct. Not sure if it's correct to deprecate it right now or closer to Jan 1, but the default (which we set [here](https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/lambda/lambdaMixins.ts#L320)) should be changed.

danielrbradley:
My suggestion would be to indicate the deprecation as early as possible to dissuade use of a runtime which will soon be reaching end of life. As of Jan 6 using the runtime will start causing deploy errors when trying to create new functions. Therefore, it would be good to give plenty of time to migrate ahead of that date to minimise disruption.

stack72:
I agree @danielrbradley - we should add the deprecation message to it now saying when it will be EOLed

stack72:
@CyrusNajmabadi / @lukehoban I'd like your thoughts on the defaulting to NodeJS 10.x please :)

[...]

PR #778, opened 2019-10-22
Add a deprecation message to NodeJS 8.x in Lambda Runtimes

stack72:
Related: #774

The NodeJS team have EOLed NodeJS 8.x on 31st December 2019. This
deprecation will warn our users that this is coming and that they
should move to a different runtime in advance

This commit also adds a deprecation to DotNetCore 1.0 which was
actually deprecated in AWS earlier in 2019

### DIFF
--- a/sdk/nodejs/lambda/runtimes.ts
+++ b/sdk/nodejs/lambda/runtimes.ts
@@ -21,32 +21,45 @@ export type Runtime =
     "dotnetcore2.1"  |
     "go1.x"          |
     "java8"          |
+    "java11"         |
     "ruby2.5"        |
     "nodejs4.3-edge" |
     "nodejs4.3"      |
     "nodejs6.10"     |
     "nodejs8.10"     |
     "nodejs10.x"     |
+    "nodejs12.x"     |
     "nodejs"         |
     "python2.7"      |
     "python3.6"      |
     "python3.7"      |
+    "python3.8"      |
     "provided"       ;
 
-export let DotnetCore1d0Runtime: Runtime = "dotnetcore1.0";
 export let DotnetCore2d1Runtime: Runtime = "dotnetcore2.1";
 export let Go1dxRuntime:         Runtime = "go1.x";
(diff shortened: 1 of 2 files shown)


stack72:
@lukehoban thoughts on adding this deprecation message now so that people can understand that this is EOLed soon?

jen20:
Do we want the general format of the deprecation messages to be clearer about what is required? IMO:

- the ones that are no longer supported should read something like "New lambda functions created using this runtime will fail" after the "No longer supported".

- for things which are planned to be deprecated, we should say something like "New lambdas created using this runtime will succeed until <date>, at which point they may start to fail".

Seem reasonable?

sdk/nodejs/lambda/runtimes.ts
```diff
@@ -21,32 +21,40 @@ export type Runtime =
...
+/** @deprecated No longer supported. New lambda functions created using this runtime will fail */export let NodeJSRuntime: Runtime = "nodejs";
+/** @deprecated No longer supported. New lambda functions created using this runtime will fail */export let NodeJS4d3EdgeRuntime: Runtime = "nodejs4.3-edge";
+/** @deprecated No longer supported. New lambda functions created using this runtime will fail */export let NodeJS4d3Runtime: Runtime = "nodejs4.3";
+/** @deprecated No longer supported. New lambda functions created using this runtime will fail */export let NodeJS6d10Runtime: Runtime = "nodejs6.10";
+/** @deprecated No longer supported. New lambda functions created using this runtime will fail */export let DotnetCore1d0Runtime: Runtime = "dotnetcore1.0";
+/** @deprecated No longer supported. New lambda functions created using this runtime will fail */export let DotnetCore2d0Runtime: Runtime = "dotnetcore2.0";
+/** @deprecated New lambdas created using this runtime will succeed until December 31, 2019, at which point they may start to fail */export let NodeJS8d10Runtime: Runtime = "nodejs8.10";
```
jen20: Missing a new line here?

Issue #793, opened 2019-11-13
Pulumi creates Glue database and table names that are incompatible with Athena


EvanBoyle:
Athena is commonly backed by Glue tables (hive metastore). Pulumi programs creating Glue tables and databases use `-` as the separator between the user defined name and the random number appended to the resource. Athena throws an error [if it sees any special characters in table/db names other than '_'](https://docs.aws.amazon.com/athena/latest/ug/tables-databases-columns-names.html#ate-table-database-and-column-names-allow-only-underscore-special-characters). This is not leading users into the pit of success, we should consider using the legal separator value here.

stack72:
@EvanBoyle is this due to our autonaming?

lukehoban:
Yes - sounds like we need to customize the autonaming for this resource.

stack72:
@lukehoban it just so happens I centralised the custom naming code yesterday 💃

EvanBoyle:
Thanks for looking into this @stack72!

EvanBoyle:
@stack72 [This doc](https://docs.aws.amazon.com/athena/latest/ug/tables-databases-columns-names.html) provides further details about how db and table naming impact querying. For the most straight forward user experience, I think we should conform to something along the lines of 

```
^[a-z]+(_[a-z]+)*$
```

Which will match `abc_def`, but not `_abc_`.  Can we convert the random number to a lowercase letter string: `tablename_123` -> `tablename_abc` ?

Issue #781, opened 2019-10-27
aws.appsync.Function defaults to generating an invalid function name


alexmcmanus:
Pulumi generates a `name` argument for most resources by appending a unique suffix to the resource name. E.g.
```ts
const lookupInputBusinessId = new aws.appsync.Function('lookupInputBusinessId', ...)
```
...will generate a function name like `lookupInputBusinessId-058765e`. Unfortunately, `-` is not a legal character for a function name. 

Pulumi -> Terraform -> AWS happily lets you create a function with an illegal name however. After creating it, the AWS console  won't let you edit the function, reporting that the name is illegal. And even more strangely, if you then try and modify the function and update via Pulumi, it leads to the following error:
```
BadRequestException: Invalid Function name given, must match pattern [_A-Za-z][_0-9A-Za-z]*
```

Note that appsync function names do not need to be unique - the `functionId` is used to reference the function and not the name. So I think that adding the unique suffix is unnecessary in this case. 

There is a simple workaround: explicitly specify the name in the Pulumi config:
```ts
const lookupInputBusinessId = new aws.appsync.Function('lookupInputBusinessId', {
  name: 'lookupInputBusinessId',
  ...
})
```

PR #802, opened 2019-11-22
Add custom name transformations for appsync_func and glue_

stack72:
Fixes: #793
Fixes: #781

### DIFF
--- a/resources.go
+++ b/resources.go
@@ -250,7 +250,20 @@ func Provider() tfbridge.ProviderInfo {
 					},
 				},
 			},
-			"aws_appsync_function":    {Tok: awsResource(appsyncMod, "Function")},
+			"aws_appsync_function": {
+				Tok: awsResource(appsyncMod, "Function"),
+				Fields: map[string]*tfbridge.SchemaInfo{
+					"name": {
+						Default: &tfbridge.DefaultInfo{
+							From: tfbridge.FromName(tfbridge.AutoNameOptions{
+								Separator: "-",
+								Maxlen:    255,
+								Randlen:   7,
+							}),
+						},
+					},
+				},
+			},
 			"aws_appsync_graphql_api": {Tok: awsResource(appsyncMod, "GraphQLApi")},
 			"aws_appsync_datasource":  {Tok: awsResource(appsyncMod, "DataSource")},
 			"aws_appsync_resolver":    {Tok: awsResource(appsyncMod, "Resolver")},
--- a/resources.go
+++ b/resources.go
@@ -16,6 +16,8 @@ package aws
 
 import (
 	"errors"
+	"fmt"
+	"math/rand"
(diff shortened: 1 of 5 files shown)


resources.go
```diff
@@ -2455,3 +2497,13 @@ func updateLegacyModuleNames(prov *tfbridge.ProviderInfo) {
 	updateModuleName("aws_ses_configuration_set", "ConfgurationSet", "ConfigurationSet",
 		sesMod, sesMod, nil)
 }
+
+func TransformWithRandomString(n int) string {
```
jen20: Does this need to be exported? Or if it is used in multiple places can it go in the bridge instead?

resources.go
```diff
@@ -1090,8 +1105,35 @@ func Provider() tfbridge.ProviderInfo {
...
-			"aws_glue_catalog_table":          {Tok: awsResource(glueMod, "CatalogTable")},
+			"aws_glue_catalog_database": {
+				Tok: awsResource(glueMod, "CatalogDatabase"),
+				Fields: map[string]*tfbridge.SchemaInfo{
+					"name": {
+						Default: &tfbridge.DefaultInfo{
+							From: tfbridge.FromName(tfbridge.AutoNameOptions{
+								Transform: func(name string) string {
+									newName := fmt.Sprintf("%s_%s", name, TransformWithRandomString(8))
+									return strings.ToLower(newName)
```
lukehoban: Why do we need this Transform?  Can't we parameterize the AutoNamer to do this without custom code here?  (OR is there a feature we should add there to handle this)?

EvanBoyle:
Looks good to me! Do we have automated tests for this sort of thing? Would be nice to execute that logic and vet it against `^[a-z]+(_[a-z]+)*$`

resources.go
```diff
@@ -250,7 +252,20 @@ func Provider() tfbridge.ProviderInfo {
...
 				},
 			},
-			"aws_appsync_function":    {Tok: awsResource(appsyncMod, "Function")},
+			"aws_appsync_function": {
+				Tok: awsResource(appsyncMod, "Function"),
+				Fields: map[string]*tfbridge.SchemaInfo{
+					"name": {
+						Default: &tfbridge.DefaultInfo{
+							From: tfbridge.FromName(tfbridge.AutoNameOptions{
+								Separator: "-",
```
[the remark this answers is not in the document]
stack72: D'oh! that should be an `_` - good spot!

resources.go
stack72: This transform stops adding any random numbers to the ID. This isn't something we have needed before. My thoughts are, if this needs moved to be a feature of it's own in the AutoNaming logic, then we will do so

resources.go
stack72: Fixed!

Issue #808, opened 2019-11-29
aws.iam.*PolicyAttachmentArgs inconsistencies


ncsibra:
Hi,

`aws.iam.RolePolicyAttachmentArgs` requires the role as a `pulumi.Input<string | Role>`, but `GroupPolicyAttachmentArgs`, `UserPolicyAttachmentArgs` and `PolicyAttachmentArgs` accepts only their respective class e.g.  `pulumi.Input<Group>`, `pulumi.Input<User>` and does not accept string, which makes hard to use them when those resources are managed outside pulumi, so we only have a name or id.
Especially because the `aws.iam.get*` functions does not return these types, but `Get*Results`, e.g. `GetGroupResults.`
Would be nice to make them similar like the `RolePolicyAttachmentArgs`.

PR #809, opened 2019-12-01
Allow use of string or Type in IAM PolicyAttachment resources

stack72:
Fixes: #808

### DIFF
--- a/sdk/nodejs/iam/groupPolicyAttachment.ts
+++ b/sdk/nodejs/iam/groupPolicyAttachment.ts
@@ -61,7 +61,7 @@ export class GroupPolicyAttachment extends pulumi.CustomResource {
     /**
      * The group the policy should be applied to
      */
-    public readonly group!: pulumi.Output<Group>;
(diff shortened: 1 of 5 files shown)


sdk/nodejs/iam/groupPolicyAttachment.ts
```diff
@@ -61,7 +61,7 @@ export class GroupPolicyAttachment extends pulumi.CustomResource {
     /**
      * The group the policy should be applied to
      */
-    public readonly group!: pulumi.Output<Group>;
+    public readonly group!: pulumi.Output<string>;
```
lukehoban: Note that although this is a typescript level breaking change - any code which no longer type checks would not have worked correctly as the true value was always a string.

stack72: I am so glad you mentioned this because I had no idea why it changed to string as the output - I just didn't want to bother you tonight with it :) Thanks for that! It really helps me

PR #810, opened 2019-12-02
Add a test for a C# example

mikhailshilkov:


mikhailshilkov:
Requires https://github.com/pulumi/pulumi/pull/3595 to build successfully

PR #815, opened 2019-12-04
Upgrade to v2.41.0 of the AWS Terraform Provider

stack72:


resources.go
```diff
@@ -1042,8 +1042,9 @@ func Provider() tfbridge.ProviderInfo {
 				DeleteBeforeReplace: true, // only 1 mount target per AZ.
 			},
 			// ECS for Kubernetes
-			"aws_eks_cluster":    {Tok: awsResource(eksMod, "Cluster")},
-			"aws_eks_node_group": {Tok: awsResource(eksMod, "NodeGroup")},
+			"aws_eks_cluster":         {Tok: awsResource(eksMod, "Cluster")},
+			"aws_eks_node_group":      {Tok: awsResource(eksMod, "NodeGroup")},
+			"aws_eks_fargate_profile": {Tok: awsResource(eksMod, "FargateProfile")},
```
lukehoban: Can we mark the `fargate_profile_name` property of this resource as automaker?  For some reason they used that instead of the normal `name`.

stack72: done! PR update on way

Issue #478, opened 2019-02-22
When getting an existing EKS cluster, there isn't a way to pull the kubeconfig


metral:
In the current eks object, we can use the `cluster` props: `certificateAuthority`, `name`, and `endpoint` to fill in the blanks in the kubeconfig file https://gist.github.com/b01eeecacccc3e284771463ed626af5e

we only do this right now in the eks object constructor, when we should actually have a `kubeconfig()` method on the object.

see community slack: https://pulumi-community.slack.com/archives/C84L4E3N1/p1550871454193500?thread_ts=1550862777.160400&cid=C84L4E3N1

stack72:
@metral I think this can be achieved using the import functionality right?

stack72:
ping @metral :)

metral:
@stack72 Apologies for the delay in response! IIUC how the import works, it takes an existing object not created in Pulumi and imports it to use. If that is correct, then this does not seem like it will work for the `kubeconfig` file as that is not stored in the cluster at all - it is generated by whatever means instantiated the cluster.

The proper fix for this is to expose a `kubeconfig()` method.

What use case are you trying to handle?

lukehoban:
We support this in the [`eks` package](https://www.pulumi.com/registry/packages/eks/), but not currently in the raw AWS provider.  It would be nice to expose that function as a general capability, even for clusters not created by that package.  But I expect if we were to do that - we would add it to that package.  This is tracked in https://github.com/pulumi/pulumi-eks/issues/11#issuecomment-430349711, so I'll close this as a dupe of that at this point.

PR #840, opened 2019-12-22
Add functionality to expose EKS KubeConfig on the Cluster resource

stack72:
Fixes: #478

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+* Expose EKS KubeConfig on EKS Cluster resource
 
 ---
 
--- a/examples/eks/.gitignore
+++ b/examples/eks/.gitignore
@@ -1,0 +1,2 @@
+/bin/
+/node_modules/
--- a/examples/eks/Pulumi.yaml
+++ b/examples/eks/Pulumi.yaml
@@ -1,0 +1,3 @@
+name: eks
+runtime: nodejs
+description: A Eks cluster that exposes the KubeConfig from a builtin function
--- a/examples/eks/index.ts
+++ b/examples/eks/index.ts
@@ -1,0 +1,38 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as aws from "@pulumi/aws";
+import * as awsx from "@pulumi/awsx";
+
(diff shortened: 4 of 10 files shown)


lukehoban:
I’m not sure we want to add one-off helpers like this in this library. We already have an EKS package with helpers for EKS. If we think this is valuable - I’d suggest we add it there. Or alternatively, contribute it to as a data source in the upstream provider to that it is available for all Pulumi languages. 

Thiughts?

stack72:
@lukehoban good points - I am more than happy to close this out - I am just trying to fix #478 

This functionality exists as part of the create in Eks already. I guess we can ask @metral if this is even needed any more :)

lukehoban:
Yeah - not clear why that issue was in this repo - I would have assumed we would address it in the EKS repo. But not 100% sure I follow what the issue is asking for - so perhaps need to confirm with @metral before making any change here.

stack72:
@lukehoban Azure, AWS and GCP all had the issue. Here's the same issue in GCP https://github.com/pulumi/pulumi-gcp/issues/97

metral:
Thanks for taking the lead and working on this @stack72.

IIRC https://github.com/pulumi/pulumi-aws/issues/478, https://github.com/pulumi/pulumi-azure/issues/186, and https://github.com/pulumi/pulumi-gcp/issues/97 were concerned with addressing the generation of kubeconfig files for cluster's that were **not** created using Pulumi.

The use case specifically is if a user created a cluster by some other means that was not Pulumi, they could `.get()` the cluster using it's name, and this would return the cluster object, but it would be void of a kubeconfig prop as we create natively expose for EKS, AKS, and GKE via Pulumi.

> Yeah - not clear why that issue was in this repo - I would have assumed we would address it in the EKS repo

Correct, this is addressed in the `pulumi/eks` repo, but it was opened here as the `.get()` method for the `aws.eks.Cluster` object lives in this repo:

https://github.com/pulumi/pulumi-aws/blob/4747717488d91a764f03015aba5b3a6d84b6f01b/sdk/nodejs/eks/cluster.ts#L108-L119

IIUC this PR is replicating what we already handle for Pulumi created clusters, so it's not needed here.

We can certainly talk through if we still want to support the ability to import clusters created outside of Pulumi, and how we can create parity to Pulumi created clusters with props like the `kubeconfig`.

PR #842, opened 2020-01-02
Stop autonaming `aws.route53.Record` `name` property

lukehoban:
This property is used as the domain name of the DNS entry - and should never be autonamed.

This is a breaking change, but any code that is broken by this is most likely not working as intended.

### DIFF
--- a/examples/route53/README.md
+++ b/examples/route53/README.md
@@ -1,3 +1,3 @@
-# examples/cloudwatch
+# examples/rout53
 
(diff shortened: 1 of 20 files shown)


examples/route53/README.md
```diff
@@ -1,3 +1,3 @@
-# examples/cloudwatch
+# examples/rout53
```
jen20:
```suggestion
# examples/route53
```

PR #853, opened 2020-01-18
Upgrade to v2.45.0 of the AWS Terraform Provider

stack72:


lukehoban:
Lots of test failures with “Error creating role” due to the role already existing. Is this a regression in the new upstream provider version?

PR #862, opened 2020-01-28
Merge pull request #862 from pulumi/stack72/bridge-v1.6.3

stack72:


.golangci.yml
```diff
@@ -4,8 +4,8 @@ linters:
     - deadcode
     - errcheck
     - goconst
-    - gofmt
     - golint
+    - gofmt
```
mikhailshilkov: Same question about the order

PR #874, opened 2020-02-11
Read all AWS env vars

### LINKED ISSUE #252: Cannot authenticate on AWS using switchrole

lukehoban:
For both programmatic access, and for pre-configuration, we must read in all the appropriate env vars.  We were previously failing to read many of the env vars for setting up programmatic access, and failing to read any of the env vars for pre-configuration checks.

Fixes #252


resources.go
```diff
@@ -251,6 +259,16 @@ func Provider() tfbridge.ProviderInfo {
 		Repository:  "https://github.com/pulumi/pulumi-aws",
 		Version:     version.Version,
 		Config: map[string]*tfbridge.SchemaInfo{
+			"access_key": {
+				Default: &tfbridge.DefaultInfo{
+					EnvVars: []string{"AWS_ACCESS_KEY_ID"},
+				},
+			},
+			"secret_key": {
+				Default: &tfbridge.DefaultInfo{
```
cleverguy25: Should AWS_PROFILE be somewhere here?

stack72: It is just below this comment :) Line 280

stack72:
@lukehoban just pushed a quick fix to the branch for the range statement

PR #875, opened 2020-02-12
Update to the latest pulumi{,-terraform-bridge}.

pgavlin:
These updates bring in schema-based NodeJS codegen. A few tweaks to
resources.go were necessary to ensure that all tokens used by this
package are in a consistent format, i.e. `aws:module/member:member`.


pgavlin:
> Do we really want to depend pulumi/pulumi@master?

In this case, yes: the latest stable release doesn't have the changes we need.

pgavlin:
Not sure what's causing the Python diffs. I'll investigate this afternoon.

stack72:
@pgavlin I think we can close this out as @mikhailshilkov already upgraded this provider to the new generator

lukehoban:
Closing out.  https://github.com/pulumi/pulumi-aws/pull/921 is a newer PR that updates these deps.

PR #876, opened 2020-02-13
A number of resources had no support for AltTypes

stack72:
The strategy we have is that a type should fall back to a primitive

an example of this would be for aws ec2 instancetypes, if we don't allow
strings then we will have to deploy the provider before we allow a user
to pick up the latest instance types

### DIFF
--- a/resources.go
+++ b/resources.go
@@ -345,8 +345,9 @@ func Provider() tfbridge.ProviderInfo {
 						Elem: &tfbridge.SchemaInfo{
 							Fields: map[string]*tfbridge.SchemaInfo{
 								"rest_api_id": {
-									Name: "restApi",
-									Type: awsTypeDefaultFile(apigatewayMod, "RestApi"),
+									Name:     "restApi",
+									Type:     "string",
+									AltTypes: []tokens.Type{awsTypeDefaultFile(apigatewayMod, "RestApi")},
 								},
 							},
 						},
@@ -357,8 +358,9 @@ func Provider() tfbridge.ProviderInfo {
 				Tok: awsResource(apigatewayMod, "Authorizer"),
 				Fields: map[string]*tfbridge.SchemaInfo{
 					"rest_api_id": {
-						Name: "restApi",
-						Type: awsTypeDefaultFile(apigatewayMod, "RestApi"),
+						Name:     "restApi",
+						Type:     "string",
+						AltTypes: []tokens.Type{awsTypeDefaultFile(apigatewayMod, "RestApi")},
 					},
 				},
 			},
@@ -366,8 +368,9 @@ func Provider() tfbridge.ProviderInfo {
 				Tok: awsResource(apigatewayMod, "BasePathMapping"),
 				Fields: map[string]*tfbridge.SchemaInfo{
 					"api_id": {
-						Name: "restApi",
-						Type: awsTypeDefaultFile(apigatewayMod, "RestApi"),
+						Name:     "restApi",
+						Type:     "string",
+						AltTypes: []tokens.Type{awsTypeDefaultFile(apigatewayMod, "RestApi")},
 					},
 				},
 			},
@@ -376,8 +379,9 @@ func Provider() tfbridge.ProviderInfo {
 				Tok: awsResource(apigatewayMod, "Deployment"),
 				Fields: map[string]*tfbridge.SchemaInfo{
 					"rest_api_id": {
-						Name: "restApi",
-						Type: awsTypeDefaultFile(apigatewayMod, "RestApi"),
+						Name:     "restApi",
+						Type:     "string",
+						AltTypes: []tokens.Type{awsTypeDefaultFile(apigatewayMod, "RestApi")},
 					},
 				},
 			},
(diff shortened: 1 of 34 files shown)


resources.go
```diff
@@ -884,7 +912,8 @@ func Provider() tfbridge.ProviderInfo {
 						AltTypes: []tokens.Type{awsTypeNoFile(iamMod, "InstanceProfile")},
 					},
 					"instance_type": {
-						Type: awsTypeDefaultFile(ec2Mod, "InstanceType"),
+						Type:     "string",
+						AltTypes: []tokens.Type{awsTypeDefaultFile(ec2Mod, "InstanceType")},
```
lukehoban: At least to date I believe our policy has been to intentionally use enums to narrow string types.  If we make the change here, the enum really has no value - it may as well just be `string`.

`string | "t2.micro"` is identical to `string`.  You won't get any type-checking, intellisense, etc. benefits.

That said, keeping this as is does have significant downsides if we don't have a robust way to keep these up to date, and honestly probably does cause more friction in practice than it's worth.

We should make a decision on this point separately from the other changes in this PR.  We should also make sure to write this decision down in resource modeling design guidelines.

cc @CyrusNajmabadi @pgavlin @EvanBoyle for input.

sdk/go/aws/alb/loadBalancer.go
```diff
@@ -202,9 +202,9 @@ type loadBalancerArgs struct {
 	// If true, the LB will be internal.
 	Internal *bool `pulumi:"internal"`
 	// The type of IP addresses used by the subnets for your load balancer. The possible values are `ipv4` and `dualstack`
-	IpAddressType *string `pulumi:"ipAddressType"`
+	IpAddressType interface{} `pulumi:"ipAddressType"`
 	// The type of load balancer to create. Possible values are `application` or `network`. The default value is `application`.
-	LoadBalancerType *string `pulumi:"loadBalancerType"`
+	LoadBalancerType interface{} `pulumi:"loadBalancerType"`
```
lukehoban: Is this expected?  Is it a breaking change?  Will we no longer be able to safely add AltTypes once Go is GA?

cc @pgavlin @EvanBoyle

resources.go
lukehoban: More background on this in https://github.com/pulumi/pulumi-aws/pull/739.

EvanBoyle: I don't have context around how these values flow through later on and what control we have over manipulation, but could we still get autocomplete if we did something along the lines off:

```ts
type  InstanceOverride = {
  override: string
}
type InstanceType = "t1micro" | "t2micro"; // ..... etc. 

type Instance = InstanceType | InstanceOverride; 

```

The goal would be to keep instanceType as up to date as possible for autocomplete, and to have instanceOverride as an escape hatch that doesn't interfere with string auotocompletion. Not sure if this would even work?

pgavlin:
> I'm also surprised there are not any changes to Python or .NET files here?

The Python and .NET code generators do not support AltTypes. I think it would be relatively straightforward to add support for this to these targets.

sdk/go/aws/alb/loadBalancer.go
pgavlin: > Is this expected? Is it a breaking change? 

This is expected. It is not a breaking change, as it only allows additional types for this property.

> Will we no longer be able to safely add AltTypes once Go is GA?

Not without a breaking change, no. However, fixing this s.t. the primitive type is used for AltTyped (`oneOf` in the Pulumi schema) properties would allow us to add proper AltTypes in Go w/o a breaking change.

lukehoban:
> The Python and .NET code generators do not support AltTypes

But these changes also change `Type` itself.  I suppose we not support that either in .NET and Python code generators?

> I think it would be relatively straightforward to add support for this to these targets.

Probably worth figuring out where we want to get on all of these ahead of GA so we are clear on what will be a breaking change as we look to being able to evolve these in the future.

> > Will we no longer be able to safely add AltTypes once Go is GA?
> Not without a breaking change, no. However, fixing this s.t. the primitive type is used for AltTyped (oneOf in the Pulumi schema) properties would allow us to add proper AltTypes in Go w/o a breaking change.

Just to make sure I understand - you are saying that as of today it is a breaking change to add AltTypes, but we could make it not a breaking change.  We possibly will need to do that so that we're not prevented from making any resource model changes post-GA.  Should we open an issue to track that?

pgavlin:
> But these changes also change Type itself. I suppose we not support that either in .NET and Python code generators?

Ah, interesting--I'll take a look.

> you are saying that as of today it is a breaking change to add AltTypes, but we could make it not a breaking change.

I am saying that as of today it is _not_ a breaking change to add AltTypes, as that strictly widens the set of values the generated code accepts (e.g. a property that used to be just `string` now accepts `interface{}` values). If we ship the widened types, however, and _then_ add AltType support to the code generator, the updated code would accept narrower types. It should always be a non-breaking change to go from a narrow type (i.e. a type w/o AltTypes) to a wider type (i.e. a type with AltTypes).

stack72:
@lukehoban I have changed this PR to not tough the enum types

Issue #890, opened 2020-03-05
AWS temporary credentials of custom aws.Provider resources are saved in Pulumi state, preventing stack deletion


nhinds:
The `aws.Provider` resource automatically grabs my `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN` environment variables and persists them in Pulumi's state as its inputs and outputs. For permanent credentials, it seems odd to store secrets from the environment like this (#884 tracks marking them as secret, at least), but for temporary credentials this prevents stacks from being deleted because Pulumi tries to use the persisted credentials when deleting the stack.

The behaviour seems to have changed in #874

In my use case, this makes it impossible to `pulumi destroy` an EKS cluster that was created using the [`eks.getRoleProvider()`](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/eks/#getRoleProvider) helper unless I run a dummy `pulumi up` to update the `aws.Provider` first.

Minimal reproduction:
```typescript
import * as aws from "@pulumi/aws";

const provider = new aws.Provider("provider", { region: "us-west-2" });
new aws.cloudwatch.LogGroup("logGroup", undefined, { provider: provider });
```

Assume a role on the CLI (or run `aws sts get-session-token` to get temporary credentials and export them into your environment), then run `pulumi up`. Wait for your original credentials to expire, then re-assume the role and run `pulumi destroy`.

More detailed steps:
1. Get temporary token and confirm it is valid
```
$ aws sts get-session-token --duration-seconds 900
{
    "Credentials": {
        "AccessKeyId": "redacted",
        "SecretAccessKey": "redacted",
        "SessionToken": "redacted",
        "Expiration": "2020-03-05T01:04:31Z"
    }
}
$ export AWS_ACCESS_KEY_ID=redacted AWS_SECRET_ACCESS_KEY=redacted AWS_SESSION_TOKEN=redacted
$ aws sts get-caller-identity
{
    "UserId": "redacted",
    "Account": "redacted",
    "Arn": "arn:aws:iam::redacted:user/redacted"
}
```
2. Create the stack with Pulumi
```
$ pulumi up
Previewing update (foo):
     Type                        Name                          Plan       
 +   pulumi:pulumi:Stack         aws-provider-credentials-foo  create     
 +   ├─ pulumi:providers:aws     provider                      create     
 +   └─ aws:cloudwatch:LogGroup  logGroup                      create     
 
[...]

Issue #814, opened 2019-12-03
CLI hangs at sts:GetCallerIdentity when temporary AWS credentials exist but are expired


austinbutler:
Where there are no AWS access keys in my shell environment, running `pulumi up`  pretty much immediately errors out with `error: unable to discover AWS AccessKeyID and/or SecretAccessKey`. But when AWS access keys exist but had expired, it just hangs (for at least several minutes) at trying `sts:GetCallerIdentity` from what I can tell from the debug output (which I have now lost 😨 ), however without debug on it just looks like it's doing normal planning forever. Ideally it quickly determines the credentials are invalid and informs the user.

Pulumi: v1.6.1
Pulumi AWS: v1.13.0

austinbutler:
Let the token expire overnight to get the output:

```
> $ pulumi preview -d
Enter your passphrase to unlock config/secrets
    (set PULUMI_CONFIG_PASSPHRASE to remember):
Previewing update (staging):

     Type                 Name                Plan       Info
 +   pulumi:pulumi:Stack  pulumi-staging  create     17 debugs
     └─ aws:s3:Bucket     newbucket              1 error

System Messages
  ^C received; cancelling. If you would like to terminate immediately, press ^C again.
  ^C received; terminating

Diagnostics:
  aws:s3:Bucket (newbucket):
    error: transport is closing

  pulumi:pulumi:Stack (pulumi-staging):
    debug: Registering resource: t=pulumi:pulumi:Stack, name=pulumi-staging, custom=false
    debug: RegisterResource RPC prepared: t=pulumi:pulumi:Stack, name=pulumi-staging
    debug: RegisterResource RPC finished: resource:pulumi-staging[pulumi:pulumi:Stack]; err: null, resp: urn:pulumi:staging::pulumi::pulumi:pulumi:Stack::pulumi-staging,,,,
    debug: Running program '/Users/abutler/Documents/pulumi' in pwd '/Users/abutler/Documents/pulumi' w/ args:
    debug: Registering resource: t=aws:s3/bucket:Bucket, name=newbucket, custom=true
    debug: RegisterResourceOutputs RPC prepared: urn=urn:pulumi:staging::pulumi::pulumi:pulumi:Stack::pulumi-staging
    debug: RegisterResource RPC prepared: t=aws:s3/bucket:Bucket, name=newbucket
[...]

PR #894, opened 2020-03-11
Avoid baking config from environment into checkpoint

lukehoban:
In https://github.com/pulumi/pulumi-aws/pull/874 we added config defaults from environment variables for four new configuration settings.  Thees config defaults are used in two places: (1) `aws.config` and (2) the defaults for `aws.Provider`.  For (1) these changes were a good thing, but for (2) they led to values from the environment getting baked into checkpoints that should not be.

It's not clear to me that we should be doing (2) at all - that is - I don't think `region` or `profile` should be picked up from the environment and baked into the checkpoint file either.  But for now we'll just revert the more recent change here which has led to the more significant immediate issue.

Part of https://github.com/pulumi/pulumi-aws/issues/814.

Fixes https://github.com/pulumi/pulumi-aws/issues/890.


lukehoban:
Going to go ahead and merge this to undo the regression - but happy to get any post-hoc feedback @pgavlin @stack72 .

PR #896, opened 2020-03-12
Fix WebServerGo test

lukehoban:


lukehoban:
Merging as we apparently don't even run this test in PR builds anyway - so will need to merge into master to verify whether this unblocks the test in Travis.

PR #899, opened 2020-03-17
Update to AWS v2.53.0

mikhailshilkov:


sdk/nodejs/eks/cluster.ts
```diff
@@ -139,7 +139,7 @@ export class Cluster extends pulumi.CustomResource {
     /**
      * Nested attribute containing `certificate-authority-data` for your cluster.
      */
-    public /*out*/ readonly certificateAuthority!: pulumi.Output<outputs.eks.ClusterCertificateAuthority>;
+    public /*out*/ readonly certificateAuthorities!: pulumi.Output<outputs.eks.ClusterCertificateAuthority[]>;
```
lukehoban: This is also a breaking change - and I think may even break `@pulumi/eks`.  Similar question on whether this is expected/necessary?

We may need to invest in the ability in `tfbridge` to project fields like this as both singular and plural to avoid this sort of change being a breaking change?

mikhailshilkov:
> There are a surprisingly high number of breaking changes around properties becoming plural 

That's probably because they changed the type from singles to arrays and then TF gen changes the name. We can override names but types are still a breaking change...

lukehoban:
> That's probably because they changed the type from singles to arrays

I'm a little confused actually.

Take `certificateAuthority` on `eks.Cluster`.  

I don't see any change to the upstream definition since the resource was first introduced: https://github.com/terraform-providers/terraform-provider-aws/blame/master/aws/resource_aws_eks_cluster.go#L47.

Why is our codegen changing?  How was our previous codegen ever correct?

mikhailshilkov:
Fascinating! I guess blame is lying then, because this file changed in https://github.com/terraform-providers/terraform-provider-aws/commit/e6489e61a37b1b27bcc1566e33b9ffb34f25b362#diff-dc67516ccb0c52b0f31f7f5ad55986b3

We rely on `MaxItems` to pluralize or not. I'll try to figure out why TF changed this.

Issue #900, opened 2020-03-18
Remediate gone MaxItems


mikhailshilkov:
TF removed a bunch of MaxItems annotations from the latest version of the AWS provider https://github.com/terraform-providers/terraform-provider-aws/commit/e6489e61a37b1b27bcc1566e33b9ffb34f25b362 which breaks our singular/plural rules and corresponding types. They were just clearing their lint warnings, so apparently they don’t need this annotation (see https://github.com/bflad/tfproviderlint/blob/ddad891cc11517c1a15c58687aaaef1667b5272b/passes/S031/README.md).

We should come up with a way to fix this for 2.53.0 and long-term.

lukehoban:
> a way to fix this for 2.53.0

For now, let's manually recover the existing experience by adding `MaxItemsOne: boolRef(true),` annotations in `resources.go`.  That will at least avoid a breaking change.

In fact, that might be all we can realistically do here?

PR #903, opened 2020-03-20
Setting MaxItems for upstream resources and datasources

stack72:
Related: #900

Please Note: There is work now required to the Rename funcs to allow
passing of a set of fields before this is 100% completled

This is required due to aws_elb and aws_lb


mikhailshilkov:
Would it make sense to merge these changes to #899 to make sure the actual diff is correct?

stack72:
@mikhailshilkov sure - I can do that :) I will cherry-pick it across

stack72:
Closed in favour of #899

PR #905, opened 2020-03-22
Update to use new module system

EvanBoyle:
Updates this provider to have a separate go.mod for `/sdk`, and to use the new modules from pulumi/pulumi. 

There's a local replace for the tfbridge repo that needs to be removed after https://github.com/pulumi/pulumi-terraform-bridge/pull/134 is merged.


stack72:
@EvanBoyle this works great! Thanks so much for this :) Merging now!

PR #910, opened 2020-03-26
Adapt package publishing for new repo layout

stack72:
This is required due to the new go mod module layout


stack72:
//cc @EvanBoyle

PR #921, opened 2020-04-03
Adopt latest pulumi and pulumi-terraform-bridge

lukehoban:
Pulls in latest codegen changes.


lukehoban:
@praneetloke There are a lot of whitespace changes that I think must be due to the examples extraction.  Were these expected?  Since these end up in language doc comments (and thus in IDEs), we probably do need to keep things cleaner.  So we may need to fix the root cause of those before merging this.

praneetloke:
> There are a lot of whitespace changes that I think must be due to the examples extraction. Were these expected?

Yes. These are from processing the [examples shortcode](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/docs.go#L160). I can take a look next week if we are ok to have these extra lines for a short period?

stack72:
Going to merge this so I can base the next provider update off it!

PR #924, opened 2020-04-03
IAM: Principals: Added ECS-tasks principal

Dominik-K:
Just found it this is missing in the list. Nice to see that there are constants for it 👍 

I refrained to add this also outside the `Principals` modules as this is deprecated since PR https://github.com/pulumi/pulumi-aws/pull/600.

I think this small addition might not worth to be in the changelog. If you think otherwise, please tell me.


stack72:
Thanks so much for this @Dominik-K 

This doesn't need a CHANGELOG entry - I will mark it with the correct tag in this PR

PR #926, opened 2020-04-03
IAM: PolicyStatement: document Action or NotAction required

Dominik-K:
When you forget the `Action` in a policy-statement, `pulumi preview` doesn't fail. I didn't find out (in reasonable time) where I could add the verification or linter inside the code. Therefore at least mentioning it in the documentation is a step forward.

**Questions** which arose for me as a quite new contributor (also [asked in #contribute Slack-channel](https://pulumi-community.slack.com/archives/CB81H6DG9/p1585937650007900)):

1. I've seen the `PolicyStatement` interface & others [are generated code](https://github.com/pulumi/pulumi-aws/pull/232/files?file-filters%5B%5D=.ts#diff-db03b6a8eac7dbf3938f52cfd3311a0aR31). In that PR https://github.com/pulumi/pulumi-aws/pull/232 it isn't mentioned how the code-generation was done. My guess it's generated from TF's [datasource `aws_iam_policy_document`](https://github.com/terraform-providers/terraform-provider-aws/blob/98b8b848ca94031b20c3e626c9d40484e3af80de/aws/data_source_aws_iam_policy_document.go). Is that correct? Is this procedure documented?
2. If it's based from these datasource: Would it make sense to add an [`ExactlyOneOf`](https://github.com/hashicorp/terraform-plugin-sdk/blob/master/helper/schema/schema.go#L227) there (currently it's just [`"actions": setOfString`](https://github.com/terraform-providers/terraform-provider-aws/blob/98b8b848ca94031b20c3e626c9d40484e3af80de/aws/data_source_aws_iam_policy_document.go#L56))? Or ...
3. Is there an equivalent setting for `ExactlyOneOf` in Typescript or Pulumi's API?

At the end I would like that at least `pulumi preview` would fail with a proper error-message on this. Even better a `pulumi lint` which directly hints the bug inline in the IDE :-)


leezen:
@Dominik-K Sorry for the super long delay in replying to this. I think this PR somehow slipped through the cracks. The file in question is handwritten, so you could add the documentation as you show. Are you interested in still making this docstring edit?

Dominik-K:
@leezen Sure. That short kind of documentation is better than not.

However, it would be nice if you could answer my questions in here (which might be found by other contributors, too).

leezen:
> I've seen the PolicyStatement interface & others are generated code. In that PR #232 it isn't mentioned how the code-generation was done. My guess it's generated from TF's datasource aws_iam_policy_document. Is that correct? Is this procedure documented?

The interfaces in `documents.ts` are not automatically generated and are hand authored. Generated code should include a warning at the top, e.g. https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/iam/getPolicyDocument.ts shows 

> Is there an equivalent setting for ExactlyOneOf in Typescript or Pulumi's API?

I don't think there's that exact equivalent in TypeScript. Probably we'd have to do something along the lines of:
`Policy = NotActionPolicy | ActionPolicy` where those then have just one of the fields.

leezen (approved):

PR #936, opened 2020-04-10
Upgrade to v2.57.0 of the AWS Terraform Provider

stack72:


sdk/dotnet/Provider.cs
```diff
@@ -80,29 +80,11 @@ public InputList<string> ForbiddenAccountIds
...
-
-        /// <summary>
-        /// Resource tag keys to ignore across all resources.
-        /// </summary>
-        public InputList<string> IgnoreTags
-        {
-            get => _ignoreTags ?? (_ignoreTags = new InputList<string>());
-            set => _ignoreTags = value;
-        }
+        public Input<Inputs.ProviderIgnoreTagsArgs>? IgnoreTags { get; set; }
```
mikhailshilkov, line 87: Hmmm... this looks like a breaking change?

lukehoban: See https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#2570-april-09-2020.

This is a breaking change in a "preview feature" - so they believe it is okay for a minor version bump.

stack72: awesome - merging then

PR #943, opened 2020-04-19
Upgrade to v2.58.0 of the AWS Terraform Provider

stack72:


CHANGELOG.md
```diff
@@ -2,7 +2,8 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+* Remove deprecated IAM Policy types and ec2 InstanceTypes
```
mikhailshilkov, line 5: Could you summarize the background for this item? I can see TS types gone - was it a manual action? Why now?

stack72: These were manually created constants that were moved into modules

so ec2.M3Medium is now ec2.InstanceTypes.M3Medium style

I suggested we do this for 2.0 and was told we should do it for our 2.1 release

PR #949, opened 2020-04-24
Upgrade to v2.59.0 of the AWS Terraform Provider

stack72:


sdk/dotnet/Ec2/Outputs/GetLaunchTemplateNetworkInterfaceResult.cs
```diff
@@ -13,7 +13,7 @@ namespace Pulumi.Aws.Ec2.Outputs
     [OutputType]
     public sealed class GetLaunchTemplateNetworkInterfaceResult
     {
-        public readonly bool AssociatePublicIpAddress;
+        public readonly string AssociatePublicIpAddress;
```
mikhailshilkov, line 16: That's a breaking change, I guess an intentional change on the TF side?

stack72: This is now fixed :)

PR #960, opened 2020-05-05
Upgrade to pulumi-terraform-bridge v2.3.0

stack72:


sdk/go.mod
```diff
@@ -4,5 +4,6 @@ go 1.13
 
 require (
 	github.com/pkg/errors v0.9.1
+	github.com/pulumi/pulumi/sdk v1.14.1
```
jaxxstorm: looks like this was pulled in -`pulumiManifest.go` ?

pgavlin: Yeah, fixed that 👍

PR #972, opened 2020-05-15
Upgrade to v2.62.0 of the AWS Terraform Provider

stack72:
Fixes: #971


mikhailshilkov:
That's a lot of breaking changes across languages. Basically, all map value types changed from untyped to strings. What did TF do this time?

PR #975, opened 2020-05-20
Add .NET Install instructions

stack72:


README.md
```diff
@@ -42,6 +42,12 @@ To use from Go, use `go get` to grab the latest version of the library
 
     $ go get github.com/pulumi/pulumi-aws/sdk/go/...
 
+### .NET
+
+To use from .NET, install using `nuget`:
```
justinvp, line 47: I realize it's a NuGet package, but should we just say the same command as we're showing below, like we do for `go get`?

```suggestion
To use from .NET, install using `dotnet add package`:
```

README.md
```diff
@@ -64,7 +70,7 @@ of this underlying capability.
 ## Reference
 
 For detailed reference documentation, please visit [the API docs](
-https://pulumi.io/reference/pkg/nodejs/@pulumi/aws/index.html).
+https://pulumi.io/reference/pkg/aws).
```
justinvp, line 73: Should we move to using the actual URL?

```suggestion
https://www.pulumi.com/docs/reference/pkg/aws/).
```

stack72: Happy to move to that :) Will update!

README.md
stack72: I followed the same example as TypeScript and Python here by mentioning the package manager - Go has no package manager - thus why it says Go get I reckon?

lukehoban:
While we’re in this file:
1. Should we update or remove the final section that points at a wiki page that says it’s out of date?
2. Is it worth linking to the cloud provider page in the docs as well as the API reference (the former includes lots of details on configuring the provider). Honestly wish these were one thing - but today they are not.

stack72:
> While we’re in this file:
> 
> 1. Should we update or remove the final section that points at a wiki page that says it’s out of date?
> 2. Is it worth linking to the cloud provider page in the docs as well as the API reference (the former includes lots of details on configuring the provider). Honestly wish these were one thing - but today they are not.

Happy to act on both :)

README.md
justinvp: When it's styled as code (with backticks) it makes me think it's a command like `npm`, `pip`, `yarn`, `go get`. If we were just saying the package manager, wouldn't we refer to these as NPM and PyPI?

stack72: Got it! Will update :)

PR #987, opened 2020-05-29
Upgrade to v2.64.0 of the AWS Terraform Provider

stack72:
Fixes: #986


provider/cmd/pulumi-resource-aws/schema.json
```diff
@@ -9696,11 +9696,8 @@
         },
         "aws:codebuild/ProjectSecondarySource:ProjectSecondarySource": {
             "properties": {
-                "auths": {
-                    "type": "array",
-                    "items": {
-                        "$ref": "#/types/aws:codebuild/ProjectSecondarySourceAuth:ProjectSecondarySourceAuth"
-                    },
+                "auth": {
```
stack72, line 9699: @mikhailshilkov the issue here is that the upstream provider implemented MaxItems: 1

This is technically a breaking change as it changes the parameter name for our users

https://github.com/terraform-providers/terraform-provider-aws/commit/dd9af8653cfc0bc6ecfeeb536946b891158ce69d

provider/cmd/pulumi-resource-aws/schema.json
```diff
@@ -9832,11 +9829,8 @@
         },
         "aws:codebuild/ProjectSource:ProjectSource": {
             "properties": {
-                "auths": {
-                    "type": "array",
-                    "items": {
-                        "$ref": "#/types/aws:codebuild/ProjectSourceAuth:ProjectSourceAuth"
-                    },
+                "auth": {
```
stack72, line 9832: @mikhailshilkov the issue here is that the upstream provider implemented MaxItems: 1

This is technically a breaking change as it changes the parameter name for our users

https://github.com/terraform-providers/terraform-provider-aws/commit/dd9af8653cfc0bc6ecfeeb536946b891158ce69d

stack72:
@mikhailshilkov I believe we are just going to have to document these are *Breaking* as if we allow our users to specify more than 1 item, TF will reject them

sdk/dotnet/ElasticSearch/GetDomain.cs
```diff
@@ -141,7 +141,7 @@ public sealed class GetDomainResult
         /// Status of a configuration change in the domain.
         /// * `snapshot_options` – Domain snapshot related options.
         /// </summary>
-        public readonly string Processing;
+        public readonly bool Processing;
```
mikhailshilkov, line 144: This is breaking too

sdk/dotnet/LightSail/Instance.cs
```diff
@@ -196,7 +196,7 @@ public partial class Instance : Pulumi.CustomResource
         public Output<string> PublicIpAddress { get; private set; } = null!;
 
         [Output("ramSize")]
-        public Output<int> RamSize { get; private set; } = null!;
+        public Output<double> RamSize { get; private set; } = null!;
```
mikhailshilkov, line 199: This is breaking too

provider/cmd/pulumi-resource-aws/schema.json
```diff
@@ -127464,7 +127537,7 @@
                         "description": "Domain in transit encryption related options.\n"
                     },
                     "processing": {
-                        "type": "string",
```
stack72, line 127467: breaking as @mikhailshilkov pointed out

provider/cmd/pulumi-resource-aws/schema.json
```diff
@@ -92979,7 +93052,7 @@
                         "type": "string"
                     },
                     "ramSize": {
-                        "type": "integer"
```
stack72, line 92982: breaking as @mikhailshilkov pointed out

provider/cmd/pulumi-resource-aws/schema.json
```diff
@@ -92862,7 +92935,7 @@
                     "type": "string"
                 },
                 "ramSize": {
-                    "type": "integer"
```
stack72, line 92865: breaking as @mikhailshilkov pointed out

lukehoban:
>  I believe we are just going to have to document these are Breaking as if we allow our users to specify more than 1 item, TF will reject them

We do have the option to override MaxItemsOne in our mapping. You still wouldn’t be able to pass multiple items - but I doubt you ever were able to pass multiple items in the first place (due to AWS restrictions that just weren’t yet manifest in the terraform provider).

My suggestion is we do that and maintain compatibility here.

I’ll add this specific case to the breaking change policy draft and share that for review today so we can discuss repeatable policy here.

sdk/dotnet/ElasticSearch/GetDomain.cs
lukehoban: This is type-breaking, but did the API work at all previously?  My guess is no. That the actual value that was returned the provider was a bool already?

That aspect of “was there a working program that no longer works” is a key ingredient in the breaking change policy proposal I am drafting. 

Another point is “does it impact <1% of users”. This feels definitely in that category.

sdk/dotnet/LightSail/Instance.cs
lukehoban: I’m struggling to have a principled answer on this one - but my feeling is this break is okay for a minor release - due to a combination of (1) rarely used resource (2) break only in .NET and Go and even then only in some cases (3) very easy to adjust code to work and(4) more “correct“ (presumably you are truly allowed to pass fractional ram sizes?

If the last point is not true, then it might make sense to just override this though.

provider/cmd/pulumi-resource-aws/schema.json
stack72: Fixed!

provider/cmd/pulumi-resource-aws/schema.json
stack72: Fixed!

PR #1001, opened 2020-06-11
update bridge to include go examples

EvanBoyle:
Adds 681 go examples 🎉 

Removes a handful of C# examples that had TODOs.


EvanBoyle:
Unsure what's causing the schema diffs in CI. Any ideas?

pgavlin:
> Unsure what's causing the schema diffs in CI. Any ideas?

This is probably related to https://github.com/pulumi/pulumi-terraform-bridge/issues/212. Try uninstalling all versions of the AWS plugin (`pulumi plugin rm resource aws`) and deleting it if it's on the path (`rm which pulumi-resource-aws`) and re-running the build.

EvanBoyle:
@pgavlin that did the trick, thanks!

Issue #1013, opened 2020-06-22
Pulumi panic on pulumi up


aj-garcia05:
```
    panic: interface conversion: interface {} is []interface {}, not *schema.Set
    goroutine 143 [running]:
    github.com/terraform-providers/terraform-provider-aws/aws.setAutoscalingTags(0xc000010a40, 0xc0010d3500, 0x17, 0x4749900)
        /home/travis/gopath/pkg/mod/github.com/pulumi/terraform-provider-aws@v1.38.1-0.20200616201417-ed844a7bd9d5/aws/autoscaling_tags.go:83 +0xe49
    github.com/terraform-providers/terraform-provider-aws/aws.resourceAwsAutoscalingGroupUpdate(0xc0010d3500, 0x4eacba0, 0xc00047af00, 0x0, 0x0)
        /home/travis/gopath/pkg/mod/github.com/pulumi/terraform-provider-aws@v1.38.1-0.20200616201417-ed844a7bd9d5/aws/resource_aws_autoscaling_group.go:939 +0x4b8
    github.com/hashicorp/terraform-plugin-sdk/helper/schema.(*Resource).Apply(0xc000484900, 0xc000931ae0, 0xc000d53da0, 0x4eacba0, 0xc00047af00, 0x1, 0x0, 0x0)
        /home/travis/gopath/pkg/mod/github.com/hashicorp/terraform-plugin-sdk@v1.13.1/helper/schema/resource.go:316 +0x263
    github.com/hashicorp/terraform-plugin-sdk/helper/schema.(*Provider).Apply(0xc00049e180, 0xc000a6d818, 0xc000931ae0, 0xc000d53da0, 0x0, 0x0, 0x0)
        /home/travis/gopath/pkg/mod/github.com/hashicorp/terraform-plugin-sdk@v1.13.1/helper/schema/provider.go:294 +0x99
    github.com/pulumi/pulumi-terraform-bridge/v2/pkg/tfbridge.(*Provider).Update(0xc000692b60, 0x64cff00, 0xc000fd85a0, 0xc000ecc2a0, 0xc000692b60, 0x4e29901, 0xc000f95a80)
```

Hello all, I encountered this error above. 

I am not sure what happened but this script worked last week witout any issues

stack72:
Hi @aj-garcia05 

[...]

Issue #997, opened 2020-06-10
autoscalingGroup panics when adding additional tag


jaxxstorm:
I'm trying to create an autoscaling group with multiple tags:

```
const autoScalingGroup = new aws.autoscaling.Group(`main`, {
    desiredCapacity: 1,
    maxSize: 1,
    minSize: 1,
    launchTemplate: {
        id: launchTemplate.id,
    },
    vpcZoneIdentifiers: [ "subnet-08acb799eefccb499" ],
    loadBalancers: ["elb-name"],
    tags: [{
        key: "ManagedBy",
        propagateAtLaunch: true,
        value: "pulumi",
    }, {
        key: "Name",
        propagateAtLaunch: true,
        value: "server"
    }],
})
```

When I run this, I get a nasty panic:

```
Diagnostics:
  pulumi:pulumi:Stack (asg-asg.production):
    panic: interface conversion: interface {} is []interface {}, not *schema.Set
    goroutine 35 [running]:
    github.com/terraform-providers/terraform-provider-aws/aws.setAutoscalingTags(0xc000136170, 0xc000e075e0, 0x17, 0x52d6000)
    	/home/travis/gopath/pkg/mod/github.com/pulumi/terraform-provider-aws@v1.38.1-0.20200605145502-fb12b583285e/aws/autoscaling_tags.go:83 +0xe49
    github.com/terraform-providers/terraform-provider-aws/aws.resourceAwsAutoscalingGroupUpdate(0xc000e075e0, 0x5a2b240, 0xc000132000, 0x0, 0x0)
    	/home/travis/gopath/pkg/mod/github.com/pulumi/terraform-provider-aws@v1.38.1-0.20200605145502-fb12b583285e/aws/resource_aws_autoscaling_group.go:945 +0x4b8
    github.com/hashicorp/terraform-plugin-sdk/helper/schema.(*Resource).Apply(0xc0001266c0, 0xc001036550, 0xc000f718c0, 0x5a2b240, 0xc000132000, 0x1, 0x0, 0x0)
    	/home/travis/gopath/pkg/mod/github.com/hashicorp/terraform-plugin-sdk@v1.13.1/helper/schema/resource.go:316 +0x263
    github.com/hashicorp/terraform-plugin-sdk/helper/schema.(*Provider).Apply(0xc000190c00, 0xc0003f7818, 0xc001036550, 0xc000f718c0, 0x0, 0x0, 0x0)
[...]

Issue #990, opened 2020-06-03
Latest update broke autoscaling groups with tags set


solidDoWant:
Due to this PR in the terraform provider (https://github.com/terraform-providers/terraform-provider-aws/pull/13515), autoscaling groups with tags set will fail due to the type change on line 83 (https://github.com/terraform-providers/terraform-provider-aws/pull/13515/files#diff-153d061c5aa12e0ddd425ff2728eb0bfL83) and 88 (https://github.com/terraform-providers/terraform-provider-aws/pull/13515/files#diff-153d061c5aa12e0ddd425ff2728eb0bfL88). It looks like the documentation was updated from "list" to "set" in all the SDK documentation but the underlying type was not updated.

Minimum code to reproduce (python)
```
from pulumi_aws import autoscaling, ec2, get_availability_zones, get_ami

ami = get_ami(
    most_recent=True,
    owners=[
        "amazon"
    ],
    filters=[
        {
            "name": "name",
            "values": ["amzn-ami-hvm-2018.03.0.20180811-x86_64-gp2"]
        }
    ]
)

launch_template = ec2.LaunchTemplate(
    "launch-template",
    instance_type="t3.medium",
    image_id=ami.id
)

autoscaling.Group(
    "as-group",
    max_size=1,
    min_size=1,
    launch_template={
        "id": launch_template.id,
        "version": launch_template.latest_version
    },
    availability_zones=get_availability_zones().names,
    tags_collection=[    # This is the important part
        {
            "key": "example",
            "value": "value",
            "propagate_at_launch": True
        }
    ]
)
```

Not sure if the root issue is here or the terraform bridge.

loopingz:
It started with version `2.7.0` and it still open.
To work around you can enforce the aws version to 2.6.1
`new aws.Provider(`provider-${region}`, { region: <any>region }, { version: "2.6.1" });`

PR #1016, opened 2020-06-24
Ensure autoscaling.Group.Tag and autoscaling.Group.Tags are correct schema type

stack72:
Fixes: #997
Fixes: #1013
Fixes: #990


stack72:
Test failure was due to too many buckets :/

Issue #1020, opened 2020-06-26
lambda vs. lambda_ Python modules


praneetloke:
It seems that there is now a `lambda_` module in the Python SDK. Moreover, the previous `lambda` module hasn't been receiving updates since May; likely due to codegen changes.

komalali:
Okay! So this is actually working as expected, lambda_ is the correct module name since `lambda` is a keyword. It looks like there was a regression a few months ago that was then corrected, but the lambda folder was a leftover build artifact.

PR #1021, opened 2020-06-26
Remove dead code

komalali:
The `lambda` folder seems to be a leftover build artifact from a regression. It is no longer needed since the correct module is called `lambda_`

Fixes: #1020


praneetloke:
Thanks for taking care of this, @komalali! I will defer approval to either @lukehoban or @stack72 on this one.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+* Remove unused code [(#1021)](https://github.com/pulumi/pulumi-aws/pull/1021)
```
praneetloke, line 5: I guess no one would have been able to use this module because of its name and hence we don't have to worry about listing this as a "breaking" change?

lukehoban: Relatedly - we only need to add CHANGELOG entries for things that have a user-facing impact.  In this case, I actually think we can safely leave out of the CHANGELOG entirely.

komalali: @praneetloke correct! @lukehoban yeah I was on the fence for this one, but that makes sense. removed it.

PR #1025, opened 2020-07-01
Add C5A Instance Type Family

con5cience:
Also minor adjustment to the ordering of strings in the instance type list.


stack72:
Thanks so much for this! LGTM

Issue #1028, opened 2020-07-03
Elasticache cluster name length restricted to 20 characters


satyaki1:
As per the latest update from AWS, the cluster name length restriction of elasticache is increased to 50 characters in place of 20.

FYI: https://aws.amazon.com/about-aws/whats-new/2019/08/elasticache_supports_50_chars_cluster_name/

But during Pulumi invocation of stack I'm getting following error -

> aws:elasticache:Cluster (wxyz-ind-1-satyaki): error: could not make instance of 'aws:elasticache/cluster:Cluster': name 'wxyz-ind-1-satyaki-' plus 7 random chars is longer than maximum length 20

---
Pulumi version is v2.5.0.
@pulumi/aws version is v2.6.0

PR #1030, opened 2020-07-03
Upgrade to v2.69.0 of the AWS Terraform Provider

stack72:
Fixes: #1029 
Fixes: #1028


sdk/dotnet/Ec2/CustomerGateway.cs
```diff
@@ -49,7 +49,7 @@ public partial class CustomerGateway : Pulumi.CustomResource
         /// The gateway's Border Gateway Protocol (BGP) Autonomous System Number (ASN).
         /// </summary>
         [Output("bgpAsn")]
-        public Output<int> BgpAsn { get; private set; } = null!;
```
mikhailshilkov, line 52: So, do we have our breaking changes policy in place yet? We need to apply it...

stack72: This is because of string to int right? This is actually a change here https://github.com/terraform-providers/terraform-provider-aws/pull/14030

It seems like TF was actually making this a strong incorrectly :/

I think we are going to have to accept this with a changelog entry as the api upstream (aws) requires an int

mikhailshilkov: Ah, let's add an explicit changelog and ship it then

stack72: I will add that at merge time to save waiting on another build for CHANGELOG

PR #1033, opened 2020-07-04
Add ability to specify aws.queue.QueuePolicy policy as policydocument

### LINKED ISSUE #1032: add PolicyDocument type to QueuePolicyArgs
Provide an option to use `PolicyDocument` instead of only json string to `aws.sqs.QueuePolicy`. This also greatly improves the type safety.

https://github.com/pulumi/pulumi-aws/blob/22a1d0fd9cc882a38d25306162be97e48c8dc4e7/sdk/nodejs/sqs/queuePolicy.ts#L136

above can be changed to
```typescript
readonly policy: pulumi.Input<string | PolicyDocument>;
```

This is also inline with the other IAM policies or s3 bucket policy and etc...

I would be willing to provide a PR for this small change if required.

koneru9999:
fixes #1032

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 _(none)_
+* Add the ability to specify aws.queue.QueuePolicy `policy` as iam.PolicyDocument
 
 ---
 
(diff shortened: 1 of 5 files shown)


koneru9999:
Since this is my first PR, I was unsure if auto generated code has to be included in the PR.

stack72:
Hi @koneru9999 

Thanks so much for this work - I just squashed this and fixed up the CHANGELOG conflict as well

This will be part of todays release for the provider

Paul

PR #1050, opened 2020-07-24
Discarding the master and slave language

bopopescu:
For diversity reasons, it would be nice to try to avoid 'master' and 'slave' terminology in this repository which can be associated to slavery. The master-slave terminology could be problematic for people in several countries which has the history of slavery like Romania, USA and many others. Thank you for considering the proposal. Let me know if any changes in the PR are needed, I would be happy to implement them.


stack72:
Hi @bopopescu 

Thank you so much for opening this PR - this is something that is important to us and we are actually discussing internally how we coordinate this approach across all of the 200+ repositories we have in the organisation + the updates in our documentation etc 

This PR isn't being discarded - I just want you to know that this is something that is ongoing for us

I hope you understand the little longer this is taking due to this coordination

Paul

Issue #1022, opened 2020-06-29
aws.acm.Certificate subjectAlternativeNames order is not consistent and replaces on different order


shellicar:
## Version
Windows - v2.5.0
@pulumi/pulumi 2.5.0
@pulumi/aws 2.11.0

## Description
I am creating a certificate with multiple (4) SAN entries.
I had noticed that I had to change the ordering of the subjectAlternativeNames value for the values to match pulumi's current state.
Otherwise on pulumi up, pulumi would detect changes and attempt to recreate the certificate, usually resulting in a situation that I need to manually intervene to fix (potentially due to a different route53.Record/acm.CertificateValidation issue).

This was a satisfactory workaround until I started to deploy to a new stack (environment).
In the other stack, the order of subjectAlternativeNames seems to be different again.
If I always use the same order (pre-sort), then pulumi wants to replace the Certificate on every update, as it thinks the value has changed.
I'm guessing this is due to how another API reports the SAN order.

leezen:
Unfortunately, this looks like a fairly deep issue upstream as discussed in https://github.com/terraform-providers/terraform-provider-aws/issues/8531

shellicar:
Thanks for the information and for the link.
For now I am working around the issue by using multiple wildacard certs with only a single SAN (the wildcard).

marns93:
https://github.com/terraform-providers/terraform-provider-aws/issues/8531 is fixed and will be released next week.

stack72:
This will be fixed and released in the v3.0.0 release

stack72:
Hi All

My most sincere apologies for taking so long to circle back here. I have just tested this with the latest pulumi-aws (v3.17.0) and at some point this has actually been fixed and I have verified that is the case.

```typescript
import * as pulumi from "@pulumi/pulumi";
import * as aws from "@pulumi/aws";
import * as awsx from "@pulumi/awsx";

const x = new aws.acm.Certificate("demo", {
domainName: "my.zone",
    validationMethod: "DNS",
    subjectAlternativeNames: [
        "one.my.zone",
        "two.my.zone",
        "three.my.zone",
    ]
});

export const options = x.domainValidationOptions;
```

pulumi up yields:

```
     Type                    Name             Status
 +   pulumi:pulumi:Stack     gh-aws-1022-dev  created
 +   └─ aws:acm:Certificate  demo             created

Outputs:
    options: [
        [0]: {
            domainName         : "three.my.zone"
[...]

PR #1056, opened 2020-07-31
Upgrade to v3.0.0 of the AWS Terraform Provider

stack72:
Fixes: #1055
Fixes: #1022


CHANGELOG.md
```diff
@@ -2,7 +2,14 @@ CHANGELOG
...
 ## HEAD (Unreleased)
-_(none)_
+* Upgrade to v3.0.0 of the AWS Terraform Provider
+  **Please Note:**
+  * A number of `deprecated` properties on resources have been removed. You can check the full list in the 
+  [upstream CHANGELOG](https://github.com/pulumi/terraform-provider-aws/blob/upstream-v3.0.0/CHANGELOG.md#300-july-31-2020)
+  * v3.0.0 introduces:
+    * Improved Credential Ordering
+    * Major changes to the `aws.acm.Certificate` to fix a number of issues regarding perpetual diffs of propertes
+    * Enabling drift detection of `loadBalancers` and `targetGroupArns` in `aws.autoscaling.Group`
```
lukehoban, line 12: Can we structure this to make more clear that the full set of changes are in the upstream CHANGELOG - but that some of the most notable changes are listed explicitly here?  Currently it reads like there are some deprecated properties removed and then three other changes - which I don't think is true?

sdk/nodejs/appmesh/virtualRouter.ts
```diff
@@ -28,7 +28,7 @@ import * as utilities from "../utilities";
  * import * as aws from "@pulumi/aws";
  *
  * const serviceb = new aws.appmesh.VirtualRouter("serviceb", {
- *     meshName: aws_appmesh_mesh_simple.id,
+ *     meshName: aws_appmesh_mesh.simple.id,
```
lukehoban, line 31: I know this isn't related to this PR - but is this "right" in terms of codegen?  This is a reference to an non-existant variable - so there's sort of no real notion of "right" - but the previous code looks more like what I would expect for a reference to an undefined variable.

cc @pgavlin @lblackstone

sdk/nodejs/appsync/resolver.ts
```diff
@@ -70,23 +64,29 @@ import * as utilities from "../utilities";
  *     $utils.appendError($ctx.result.body, $ctx.result.statusCode)
  * #end
  * `,
- *     type: "Query",
+ *     cachingConfig: {
+ *         cachingKeys: [
+ *             `$context.identity.sub`,
```
lukehoban, line 69: Is it intentional that we changed lots of string literals to use backticks?  Are we doing this when there is a literla `$` in the string now?  

cc @pgavlin @lblackstone

sdk/nodejs/appmesh/virtualRouter.ts
pgavlin: I believe that this was caused by the change from 0.11 to 0.12 syntax: https://github.com/terraform-providers/terraform-provider-aws/commit/d47e82f170c20e655a30557289ef3772c44e13a8#diff-0601a42442cec4be2db943fba3998311

With TF12, we render this into a traversal without trying to infer the kind of the missing reference. We could improve that and retain the original rendering by doing that inference in `tf2pulumi`.

sdk/nodejs/appsync/resolver.ts
pgavlin: Same answer re: the transition to TF12 syntax. I'm not 100% sure what's triggering this, but the `$`s are a good guess.

lukehoban:
> Is it possible to seperate those two things? It's hard to tell what the v3.0.0 changes are as there is a lot of noise (and a number of questions on whether the codegen changes are even "right") mixed in.

From @pgavlin’s other replies it sounds sa like the answer is “no” since these changes are coming from examples moving from TF0.11 to TF0.12. So will just need to review these together.

stack72:
Merged as 6c13f319a64f3fdf2c71cb3808715a12c2322f0d

PR #1088, opened 2020-08-23
Add Python examples that use input types

justinvp:
Modify the Python examples that specify nested inputs to specify them using the new input classes rather than dicts.

I copied the existing tests into a new dir with `-old` suffix without modification, so no need to review these. These tests use the "old" way of specifying nested inputs with dicts and remain to ensure this continues to work.

Also, some minor cleanup while making changes to these test files: switched to using `filepath.Join` consistently, rather than `path.Join`.


stack72 (approved):
<3

PR #1101, opened 2020-08-29
utils.ts causes a circular reference

JakeGinnivan:
An example of this is cloudwatch eventRuleMixin, which references lambda.ts, which imports utils.ts, which imports the root mixin file, which requires everything.

This causes a runtime error with classes.

I am unsure about the impacts of removing this, as far as I can tell it is imported from index.ts, meaning if you import `@pulumi/aws` the side effect of adding the proxied .sdk will still happen. But there may be other things I cannot spot

### DIFF
--- a/sdk/nodejs/utils.ts
+++ b/sdk/nodejs/utils.ts
@@ -12,14 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Directly import awsMixins.  We want to execute the code within it, but we don't want to import or
-// export any values from it.  Specifically, awsMixins adds a *getter* property (called "runtime")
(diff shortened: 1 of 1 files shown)


leezen:
@JakeGinnivan Apologies for the lack of reply on this. Do you have an example program this PR is seeking to address?

leezen:
Closing stale PR.

JakeGinnivan:
Ah sorry @leezen I missed your first comment.

The issue is TypeScript compilation time, this circular reference means that even if you import `@pulumi/aws/route53` or similar it ends up pulling in the root of the project. 

I made a similar fix to the AWS SDK. https://github.com/aws/aws-sdk-js/pull/3398

The issue is just the volume of code that the AWS and pulumi SDKs have (due to the AWS services surface area).

PR #1119, opened 2020-09-16
Upgrade to v3.2.0 of the AWS Terraform Provider

stack72:
Fixes: #1074


examples/examples_nodejs_test.go
```diff
@@ -173,9 +173,8 @@ func TestAccAlbNew(t *testing.T) {
 			// RunUpdateTest: true,
 			EditDirs: []integration.EditDir{
 				{
-					Dir:             "step2",
-					Additive:        true,
-					ExpectNoChanges: true,
```
leezen, line 178: Why does this change?

examples/examples_py_test.go
```diff
@@ -44,9 +44,8 @@ func TestAccAlbLegacyPy(t *testing.T) {
 					RunUpdateTest: true,
 					EditDirs: []integration.EditDir{
 						{
-							Dir:             "step2",
-							Additive:        true,
-							ExpectNoChanges: true,
```
leezen, line 49: Why does this change?

examples/examples_py_test.go
```diff
@@ -65,9 +64,8 @@ func TestAccAlbNewPy(t *testing.T) {
...
```
leezen, line 70: Why does this change?

sdk/nodejs/types/output.ts
```diff
@@ -6488,7 +6542,7 @@ export namespace ec2 {
 
     export interface GetLaunchTemplateNetworkInterface {
         associatePublicIpAddress?: boolean;
-        deleteOnTermination: boolean;
+        deleteOnTermination: string;
```
leezen, line 6545: Is this change expected?

sdk/go/aws/ec2/pulumiTypes.go
```diff
@@ -14632,8 +14689,8 @@ func (o GetLaunchTemplateMonitoringArrayOutput) Index(i pulumi.IntInput) GetLaun
 }
 
 type GetLaunchTemplateNetworkInterface struct {
-	AssociatePublicIpAddress *bool `pulumi:"associatePublicIpAddress"`
-	DeleteOnTermination      bool  `pulumi:"deleteOnTermination"`
+	AssociatePublicIpAddress *bool  `pulumi:"associatePublicIpAddress"`
+	DeleteOnTermination      string `pulumi:"deleteOnTermination"`
```
leezen, line 14693: (Also left a comment on the TS side)

examples/examples_nodejs_test.go
stack72: Subnets are now a computed field so changes outside of the RouteTable will cause diffs

examples/examples_py_test.go
stack72: Subnets are now a computed field so changes outside of the RouteTable will cause diffs

examples/examples_py_test.go
stack72: Subnets are now a computed field so changes outside of the RouteTable will cause diffs

sdk/nodejs/types/output.ts
stack72: This is not - we need to override this as its due to their "nullable bool" approach

sdk/go/aws/ec2/pulumiTypes.go
stack72: This is not - we need to override this as its due to their "nullable bool" approach

leezen (approved):

PR #1120, opened 2020-09-17
Upgrade to v3.3.0 of the AWS Terraform Provider

stack72:
Fixes: #1084


sdk/nodejs/ec2/subnet.ts
```diff
@@ -97,7 +97,7 @@ export class Subnet extends pulumi.CustomResource {
      * The IPv6 network range for the subnet,
      * in CIDR notation. The subnet size must use a /64 prefix length.
      */
-    public readonly ipv6CidrBlock!: pulumi.Output<string>;
+    public readonly ipv6CidrBlock!: pulumi.Output<string | undefined>;
```
leezen, line 100: This could be source breaking if someone uses this within an apply and doesn't account for undefined. Is there a way to avoid this change?

stack72: There is not - as this is an output we can't do anything about undefined params :/

leezen (approved):

Issue #1130, opened 2020-09-22
Schematize enum overlays


komalali:
We currently have enum overlays in nodejs, example: https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/region.ts

Schematize these types so that they can be generated for all supported languages.

leezen:
Specific request for `ManagedPolicies in #1147

PR #1151, opened 2020-10-07
Schematized enums

komalali:
Schematizing the enums in pulumi-aws and deprecating existing constants. 

### Remaining enums
  
- [x] autoscaling/notifcationType.ts
- [x] alb/ipAddressType.ts
- [x] alb/loadBalancerType.ts
- [x] applicationloadbalancing/ipAddressType.ts
- [x] applicationloadbalancing/loadBalancerType.ts
- [x] ec2/instanceType.ts
- [x] ec2/instancePlatform.ts
- [x] ec2/placementStrategy.ts
- [x] ec2/protocolType.ts
- [x] ec2/tenancy.ts
- [x] rds/engineMode.ts
- [x] rds/engineType.ts
- [x] rds/instanceType.ts
- [x] rds/storageType.ts
- [x] route53/recordType.ts
- [x] s3/cannedAcl.ts
- [x] ssm/parameterType.ts

### Outlier
[Service Principals](https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/iam/principals.ts) are arguably enums that don't fit into the currently allowed types for enums. Is there anything we could do there?

Resolves https://github.com/pulumi/pulumi-aws/issues/1130


provider/resources.go
```diff
@@ -2190,6 +2191,68 @@ func Provider() tfbridge.ProviderInfo {
...
+					{Value: "python3.8", Name: "Python3d8"},
+					{Value: "provided", Name: "Custom"},
+					{Value: "nodejs", Name: "NodeJS", DeprecationMessage: "No longer supported. New lambda functions created using this runtime will fail."},
+					{Value: "nodejs4.3-edge", Name: "NodeJS4d3Edge", DeprecationMessage: "No longer supported. New lambda functions created using this runtime will fail."},
+					{Value: "nodejs4.3", Name: "NodeJS4d3", DeprecationMessage: "No longer supported. New lambda functions created using this runtime will fail."},
+					{Value: "nodejs6.10", Name: "NodeJS6d10", DeprecationMessage: "No longer supported. New lambda functions created using this runtime will fail."},
+					{Value: "nodejs8.10", Name: "NodeJS8d10", DeprecationMessage: "No longer supported. New lambda functions created using this runtime will fail."},
+					{Value: "dotnetcore1.0", Name: "DotnetCore1d0", DeprecationMessage: "No longer supported. New lambda functions created using this runtime will fail."},
+					{Value: "dotnetcore2.0", Name: "DotnetCore2d0", DeprecationMessage: "No longer supported. New lambda functions created using this runtime will fail."},
```
justinvp, line 2252: Do we need to carry these deprecated values forward in the schema? Seems like a good opportunity to avoid adding new deprecated values to the new enums. Of course, we'd keep the old deprecated constants in the overlays.

justinvp: Another thing: Should we do a pass over the current list of non-deprecated values to see if they should be deprecated (and if so, deprecate the overlay constant, and not add them here)?

komalali: Good call, will do.

komalali: @justinvp we can't exclude them from the new enums if we still want to maintain the older constants. The reason is that codegen is now responsible for creating the union type that determines all possible values i.e. `type SomeType = "foo" | "bar`. If we exclude those values from the schema, we can't have constants of the enum type that declare values outside of the type i.e. `const BazSomeType: SomeType = "baz"`

justinvp: Ah, bummer.

komalali: I did find some new ones though and only added them to the schematized enums.

provider/resources.go
```diff
@@ -2190,6 +2191,664 @@ func Provider() tfbridge.ProviderInfo {
...
+					{Name: "CNAME", Value: "CNAME"},
+					{Name: "CAA", Value: "CAA"},
+					{Name: "MX", Value: "MX"},
+					{Name: "NAPTR", Value: "NAPTR"},
+					{Name: "NS", Value: "NS"},
+					{Name: "PTR", Value: "PTR"},
+					{Name: "SOA", Value: "SOA"},
+					{Name: "SPF", Value: "SPF"},
+					{Name: "SRV", Value: "SRV"},
+					{Name: "TXT", Value: "TXT"},
```
justinvp, line 2838: Just curious: It looks like these have the same names and values. I see `aws:ssm/ParameterType:ParameterType` below only has values without names. Is there a reason `aws:route53/RecordType:RecordType` needs the names to be specified?

komalali: Haha nope, just zoned out with all the copy-pasta. Will fix it up.

sdk/nodejs/ec2/placementStrategy.ts
```diff
@@ -19,23 +19,20 @@
...
+ * @deprecated Use PlacementStrategy.Spread instead.
  */
 export let SpreadStrategy: PlacementStrategy = "spread";
 
 /**
- * A `cluster` placement group is a logical grouping of instances within a single 
+ * A `cluster` placement group is a logical grouping of instances within a single
  * Availability Zone that benefit from low network latency, high network throughput.
+ *
+ * @deprecated Use PlacementStrategy.Spread instead.
```
justinvp, line 35: Copy/paste typo?

```suggestion
 * @deprecated Use PlacementStrategy.Cluster instead.
```

sdk/nodejs/ec2/instanceType.ts
```diff
@@ -21,341 +21,341 @@
 
 /* tslint:disable:max-line-length ordered-imports variable-name */
 
+import { InstanceType } from ".";
+
 export module InstanceTypes {
```
justinvp, line 26: Add top-level deprecation message?

```suggestion
/** @deprecated Use InstanceType instead. */
export module InstanceTypes {
```

sdk/nodejs/ec2/tenancy.ts
```diff
@@ -19,21 +19,11 @@
 //
 // These just give a better developer experience and are just sugared strings.
 
+import {Tenancy} from ".";
+
 export module Tenancies {
```
justinvp, line 24:
```suggestion
/** @deprecated Use Tenancy instead. */
export module Tenancies {
```

sdk/nodejs/ec2/tenancy.ts
```diff
@@ -19,21 +19,11 @@
...
-    "dedicated" ;
-
-/**
- * @deprecated Use the equivalent constant in module `Tenancies` instead
- */
-export const DefaultTenancy: Tenancy = "default";
-/**
- * @deprecated Use the equivalent constant in module `Tenancies` instead
- */
-export const DedicatedTenancy: Tenancy = "dedicated";
```
justinvp, line 38: Are we sure we want to delete these?

sdk/nodejs/iam/managedPolicies.ts
```diff
@@ -16,268 +16,532 @@
 import {ARN} from "../arn";
 
 export module ManagedPolicies {
+    /**@deprecated Use ManagedPolicy.AWSAccountActivityAccess instead. */
```
justinvp, line 19: Nit: add a space for these

```suggestion
    /** @deprecated Use ManagedPolicy.AWSAccountActivityAccess instead. */
```

sdk/nodejs/iam/managedPolicies.ts
```diff
@@ -16,268 +16,532 @@
 import {ARN} from "../arn";
 
 export module ManagedPolicies {
```
justinvp, line 18:
```suggestion
/** @deprecated Use ManagedPolicy instead. */
export module ManagedPolicies {
```

sdk/nodejs/rds/instanceType.ts
```diff
@@ -20,125 +20,125 @@
 // These give a better developer experience and are just sugared strings.
 // Based from https://aws.amazon.com/rds/instance-types/
 
+import {InstanceType} from ".";
+
 export module InstanceTypes {
```
justinvp, line 25:
```suggestion
/** @deprecated Use InstanceType instead */
export module InstanceTypes {
```

sdk/nodejs/region.ts
```diff
@@ -12,57 +12,53 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import {Region} from ".";
+
+/** @deprecated The AFSouth1Region constant is deprecated, use Region.AFSouth1 instead. */
```
justinvp, line 17: Nit: Simplify these messages to be consistent with messages elsewhere?

```suggestion
/** @deprecated Use Region.AFSouth1 instead. */
```

provider/resources.go
justinvp: Was thinking about this more, and I think we could still remove these from the schema (and new enum), and still maintain the old constants.

Instead of having these defined as follows (as they currently are):

```typescript
/** @deprecated No longer supported. New lambda functions created using this runtime will fail */
export let NodeJSRuntime: Runtime = "nodejs";
```

We change it to:

```typescript
/** @deprecated No longer supported. New lambda functions created using this runtime will fail */
export let NodeJSRuntime = "nodejs" as Runtime;
```

The `as Runtime` [type assertion](https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions) makes the TS compiler happy, even though "nodejs" is not really a valid value of the new `Runtime` enum.

I couldn't come up with any downsides.

sdk/nodejs/route53/recordType.ts
```diff
@@ -12,31 +12,31 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import {RecordType} from ".";
+
 export module RecordTypes {
```
justinvp, line 17:
```suggestion
/** @deprecated Use RecordType instead. */
export module RecordTypes {
```

sdk/nodejs/s3/bucket.ts
```diff
@@ -4,10 +4,11 @@
 import * as pulumi from "@pulumi/pulumi";
 import * as inputs from "../types/input";
 import * as outputs from "../types/output";
+import * as enums from "../types/enums";
 import * as utilities from "../utilities";
 
 import {PolicyDocument} from "../iam";
-import {CannedAcl} from "./index";
+import {CannedAcl, RoutingRule} from "./index";
```
justinvp, line 11: Interesting. It doesn't look like `RoutingRule` is used in this file. Wonder why it's getting pulled in now.

sdk/nodejs/types/enums/lambda/index.ts
```diff
@@ -0,0 +1,55 @@
...
+    Python2d7: "python2.7",
+    Python3d6: "python3.6",
+    Python3d7: "python3.7",
+    Python3d8: "python3.8",
+    Custom: "provided",
+    CustomAL2: "provided.al2",
+    /**
+     * @deprecated No longer supported. New lambda functions created using this runtime will fail.
+     */
+    NodeJS: "nodejs",
```
justinvp, line 25: I thought of a way that we could avoid adding these deprecated values to the new enums. I added a comment to the previous comment thread.

sdk/nodejs/autoscaling/metrics.ts
```diff
@@ -21,31 +21,24 @@
 //
 // These give a better developer experience and are just sugared strings.
 
+import {MetricsGranularity, Metric} from ".";
 
-import * as cloudwatch from "../cloudwatch";
-
+// @deprecated Use MetricsGranularity.OneMinute instead.
```
justinvp, line 26: These need to use JSDoc comments (with the `/` and two `**`, e.g. `/** @deprecated Use MetricsGranularity.OneMinute instead. */`)

sdk/nodejs/autoscaling/notificationType.ts
```diff
@@ -19,16 +19,15 @@
 //
 // These give a better developer experience and are just sugared strings.
 
+import {NotificationType} from ".";
+
+// @deprecated Use NotificationType.InstanceLaunch instead.
```
justinvp, line 24: These need to use JSDoc style comments

sdk/nodejs/arn.ts
```diff
@@ -12,8 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import {ManagedPolicy} from "./iam";
 /**
  * An ARN is an Amazon Resource Name, and uniquely identifies a region globally across all accounts and regions.
  */
-export type ARN = string;
+export type ARN = string | ManagedPolicy;
```
justinvp, line 19: I didn't look into the details here. What was the reason for this change?

justinvp (approved):
Some comments, but LGTM otherwise!

> Service Principals are arguably enums that don't fit into the currently allowed types for enums. Is there anything we could do there?

Hmm, guessing there isn't anything we can really do here right now. I think it's OK. There are other types we're adding in the overlays around documents for TypeScript that doesn't fit into the enum pattern. We'll probably always have a small number of overlays that add ergonomic changes like this per language (if it makes sense for a language).

examples/examples_nodejs_test.go
```diff
@@ -322,7 +332,7 @@ func TestAccServerlessRaw(t *testing.T) {
 	test := getJSBaseOptions(t).
 		With(integration.ProgramTestOptions{
 			Dir:           filepath.Join(getCwd(t), "serverless-raw"),
-			RunUpdateTest: true,
+			RunUpdateTest: false,
```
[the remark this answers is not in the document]
komalali: I updated the test to use the new enums for `ManagedPolicy`. According to @stack72 if this is set to `true` it tests against the published pulumi-aws first and since these enums aren't released yet the test would fail. Can set this back once we get these out.

sdk/nodejs/arn.ts
komalali: ARN is a "relaxed enum". The existing [ManagedPolicies](https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/iam/managedPolicies.ts) have ARN as their type. Schematizing those as ManagedPolicy means that the type is `ManagedPolicy` instead of `ARN` so `ARN` can now be `ManagedPolicy` or any odd string.

sdk/nodejs/s3/bucket.ts
komalali: Yeah I don't know why this is happening but it's unrelated to the enums. I have another PR where I am just updating the bridge version and seeing the same change. Will clear that up before I merge this.

sdk/nodejs/alb/ipAddressType.ts
```diff
@@ -12,9 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import {IpAddressType} from ".";
+
+/** @deprecated Use IpAddressType.Ipv4 instead. */
```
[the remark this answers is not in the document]
komalali: I'd prefer to keep the `@deprecated` to clearly signal to the user that there are new constants without them having to specifically look for them.

examples/lambda-layer-new/index.ts
```diff
@@ -0,0 +1,24 @@
...
+
+let layer = new aws.lambda.LayerVersion("my-layer-version", {
+    code: new asset.AssetArchive({
+        "index.js": new asset.StringAsset(
+            "exports.handler = (e, c, cb) => cb(null, {statusCode: 200, body: 'Hello, world!'});",
+        ),
+    }),
+    compatibleRuntimes: [
+        aws.lambda.Runtime.NodeJS10dX,
+        aws.lambda.Runtime.NodeJS12dX,
```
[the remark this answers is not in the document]
komalali: Yeah. I wanted to have one test where we use old and new just to make sure they both continue to work. I can just change it to be one test where we just use the new ones if there's a strong preference.

sdk/nodejs/autoscaling/metrics.ts
komalali: Thanks for catching these

provider/resources.go
komalali: Kinda feels like cheating 😅

komalali:
It looks like there is no `lint` or `lint-sdk` jobs in our workflows but they are required checks for merging. Someone who is the admin of the repo probably needs to remove those checks from the settings if they are no longer being run.

stack72:
@komalali we had to remove these recently as they were timing out :/

I can merge this PR for you

PR #1163, opened 2020-10-10
Upgrade to pulumi-terraform-bridge v2.11.0

stack72:


sdk/nodejs/s3/bucket.ts
```diff
@@ -7,7 +7,7 @@ import * as outputs from "../types/output";
 import * as utilities from "../utilities";
 
 import {PolicyDocument} from "../iam";
-import {CannedAcl} from "./index";
+import {CannedAcl, RoutingRule} from "./index";
```
komalali, line 10: This is strange. It doesn't look like `RoutingRule` is used in this file at all.

stack72: 🤷 

This is jsut what was generated locally for me

komalali: Yeah I know I'm seeing it too in the enums PR. It's not a big deal really, mostly just curious why it's happening.

Interesting though that your machine didn't have whatever issues mine is having with regressions related to other providers. But glad it's working for you at least.

stack72: Welcome to the world of provider generation :D

komalali (approved):

PR #1166, opened 2020-10-12
Remove explicit deprecation of existing constants.

komalali:


stack72:
Curious as to why we are removing the deprecation? I'm sure the idea will be to remove these in the future right?

stack72 (approved):

PR #1169, opened 2020-10-13
Fix typo in CrawlerS3Target documentation for path field

Otanikotani:
Hello!
This is a fix for a typo I found when writing my infrastructure code for AWS Glue. Documentation for `path` field changed, correcting the description of the target to crawl (should be S3 bucket, not DynamoDB table)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
+* Fix typo in CrawlerS3Target documentation for path field, correcting the description of the target to crawl (should be S3 bucket, not DynamoDB table)
 * Upgrade to v3.10.0 of the AWS Terraform Provider
 * Disable autonaming on `directoryservice.Directory` as the `name` needs to be a fully qualified name for the directory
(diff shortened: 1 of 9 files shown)


Otanikotani:
Hi guys, I am not sure why the build fails. It says that the working directory is not clean on a build and some files are changed:
```
error: working tree is not clean, aborting!
HEAD detached at pull/1169/merge
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   provider/cmd/pulumi-resource-aws/schema.json
	modified:   sdk/python/pulumi_aws/glue/_inputs.py
	modified:   sdk/python/pulumi_aws/glue/outputs.py
```

Those are the files I changed, but I am not very sure why build fails though. Do I need to do something extra except commits and updating a changelog file?

leezen:
Thanks for pointing this out and submitting a PR. This looks like a bug in our generator and not something you'd typically fix at the schema level (since the schema is generated as well). This looks like another instance of https://github.com/pulumi/docs/issues/3895 where the schema picks up the incorrect description for properties that share the same name across supporting types. We'll want to fix this as part of docs generation and will track in the linked issue.

Issue #1174, opened 2020-10-16
AWS Lex bot, intent, and slottypes can not use autogenerated names with hyphens


displague:
The generated name is not acceptable for use in AWS Lex when using the order_flower_bot example at https://www.pulumi.com/docs/reference/pkg/aws/lex/bot/:

```
Diagnostics:
  aws:lex:Bot (order_flower_bot):
    error: aws:lex/bot:Bot resource 'order_flower_bot' has a problem: expected value of name to match regular expression "^([A-Za-z]_?)+$", got order_flower_bot-d885833
```

This also applies to `intents` and `slottypes` (I haven't checked other resources).

mikhailshilkov:
Should we keep auto naming but remove dashes? (`Separator: ""`)

PR #1186, opened 2020-10-24
Changes the autonaming of Lex resources and adds some missing enum values

stack72:
Fixes: #1174
Fixes: #1187 
Fixes: #1190

Disables the autonaming of Lex resources and adds some missing enum values


mikhailshilkov:
Should we keep auto naming but remove dashes? (`Separator: ""`)

stack72:
> Should we keep auto naming but remove dashes? (`Separator: ""`)

We'd also need to remove the digits as it doesn't support that. I am happy to write a custom function for this if you feel it's a better thing to do but this feels a lot of overhead

mikhailshilkov:
> if you feel it's a better thing

It's about the same consistency thing that popped up on Slack recently, the goal would be not to make users guess if autonaming is applied. Otherwise, I don't feel too strong about that.

jaxxstorm (approved):

Issue #1210, opened 2020-11-10
Autonaming for MSK (Kafka) clusters


leezen:
`clusterName` isn't mapped to be autonamed for MSK clusters

PR #1211, opened 2020-11-11
Enabling the autonaming of MSK Clusters

stack72:
Fixes: #1210


komalali (approved):
Ooh nice, we actually need to do the same with cluster_identifier for aws_rds_cluster, I'll be pushing that up soon.

stack72:
Thanks @komalali

PR #1217, opened 2020-11-15
Ensure autonamed FIFO Queue has ".fifo" suffix

### LINKED ISSUE #967: Support autonaming of SQS fifo queues
SQS FIFO queue names must have a ".fifo" suffix. This does not play well will auto-naming because the random hex will be tacked on at the end yielding an invalid name. I suggest that the ".fifo" suffix should be added by pulumi when autonaming a fifo queue.

Here is some example C# code:

```csharp

    // fails on AWS because name must end with .fifo
    var q1 = new Queue("myqueue", new QueueArgs {
        FifoQueue = true
    });

    //fails in pulumi with the message:
    //only alphanumeric characters and hyphens allowed in "name"
    var q2 = new Queue("myqueue.fifo", new QueueArgs {
        FifoQueue = true
   });

   //explicitly assigning a name works but does not use autonaming 
   var q3 = new Queue("myqueue" new QueueArgs {
        FifoQueue = true,
        Name = "myqueue.fifo"
    });
```

lukehoban:
Fixes #967.


stack72 (approved):
This is going to roll out the import docs for AWS which I was going to start on Monday as long as you are on with that, then I am too

lukehoban:
> This is going to roll out the import docs for AWS which I was going to start on Monday as long as you are on with that, then I am too

Do need to wait to merge https://github.com/pulumi/pulumi-terraform-bridge/pull/297 first, so that we can update this PR to a `master` commit of the bridge dependency.  Also not sure if you are okay with picking up a new bridge in the middle of the import docs work.  Happy for this to wait a couple days.

stack72:
So the import docs work is at a point now we can ship it and iterate it as we see fit

Ideally we'd merge the bridge work and then cut a tag but for the purposes of this fix, we can do it this way and I will update all repos to the new tag at the start of the week - this just gets a fix over the line at the same time :)

stack72:
Superceded by #1257

PR #1225, opened 2020-11-19
Upgrade to v3.16.0 of the AWS Terraform Provider

stack72:
Fixes: #1223


provider/resources.go
```diff
@@ -2292,6 +2297,42 @@ func Provider() tfbridge.ProviderInfo {
...
+					"logging_configuration": {
+						CSharpName: "LoggingConfig",
+					},
+				},
+			},
+			"aws_networkfirewall_rule_group": {
+				Tok: awsResource(networkFirewallMod, "RuleGroup"),
+				Fields: map[string]*tfbridge.SchemaInfo{
+					"rule_group": {
+						CSharpName: "RuleGroupConfiguration",
```
mikhailshilkov, line 2332: I don't think we have a guideline about the naming here. I would stay less creative and appended `Value` to every conflicting name in C#. Not sure how strong you are about `Configuration`.

mikhailshilkov (approved):
LGTM with one optional suggestion

provider/resources.go
stack72: So I added Configuration as it's a config block... usually I am much less aggressive with these names

PR #1246, opened 2020-12-03
Update enum types

komalali:
This PR updates the enum types to use the new generated types and use them as relaxed enums with a fallback to string.

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -974,7 +974,8 @@ func Provider() tfbridge.ProviderInfo {
 						AltTypes: []tokens.Type{awsTypeDefaultFile(iamMod, "InstanceProfile")},
 					},
 					"instance_type": {
-						Type: awsTypeDefaultFile(ec2Mod, "InstanceType"),
+						Type:     "string",
+						AltTypes: []tokens.Type{awsType(ec2Mod, "InstanceType", "InstanceType")},
 					},
 					"instance_state": {
 						CSharpName: "State",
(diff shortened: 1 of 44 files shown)


provider/resources.go
```diff
@@ -561,7 +561,7 @@ func Provider() tfbridge.ProviderInfo {
 					},
 					"metrics_granularity": {
 						Type:     "string",
-						AltTypes: []tokens.Type{awsType(autoscalingMod, "metrics", "MetricsGranularity")},
+						AltTypes: []tokens.Type{awsType(autoscalingMod, "MetricsGranularity", "MetricsGranularity")},
```
stack72, line 564: How is this manifested to the end user (if at all)?

stack72 (approved):
This is great - would it be possible to add a test that uses an enum?

provider/resources.go
komalali: It doesn't result in any impactful user-facing changes, just this: https://github.com/pulumi/pulumi-aws/blob/master/examples/lambda-layer-new/index.ts#L17-L18

komalali:
@stack72 There are a [couple](https://github.com/pulumi/pulumi-aws/blob/master/examples/lambda-layer-new/index.ts#L17-L18) [examples](https://github.com/pulumi/pulumi-aws/blob/master/examples/serverless-raw/index.ts#L44) that use the TS enums, but I'm going to make another PR with examples updated to use it in each language.

stack72:
Yeah this was just a follow up not a blocker :) merge away!

PR #1263, opened 2020-12-14
update to v3.21.0 of the terraform provider

jaxxstorm:
This change also disables the tests dedicated hosts until #1264 is figured out


stack72:
> This change also disables the dedicated hosts until #1264 is figured out

I believe you mean the tests for the dedicated hosts - not the actual resources themselves right?

jaxxstorm:
Yep, fixed, sorry for the confusion

komalali (approved):

examples/examples_nodejs_test.go
```diff
@@ -24,6 +25,7 @@ func TestAccDedicatedHosts(t *testing.T) {
 
 	integration.ProgramTest(t, &test)
 }
+*/
```
komalali, line 28: Is this meant to be excluded? If yes, could we add a note for why?

jaxxstorm: done!

PR #1272, opened 2020-12-19
Add ECS Managed policies

ahanoff:
AWS is planning to deprecate https://docs.aws.amazon.com/AmazonECS/latest/developerguide/managed-policy-transition.html
and pulumi-awsx is using this policy by default 
https://github.com/pulumi/pulumi-awsx/blob/2506065d78d21f88fea54d28f240c8e27d2fae3c/nodejs/awsx/ecs/taskDefinition.ts#L152

so to be able replace it in pulumi-awsx, I added missing ECS managed policies here first


ahanoff:
it was added already in ManagedPolicy, I was confused by ManagedPolicies

PR #1273, opened 2020-12-20
Add ECS managed policies

ahanoff:
AWS is planning to deprecate AmazonEC2ContainerServiceFullAccess policy https://docs.aws.amazon.com/AmazonECS/latest/developerguide/managed-policy-transition.html
and pulumi-awsx is using this policy by default
https://github.com/pulumi/pulumi-awsx/blob/2506065d78d21f88fea54d28f240c8e27d2fae3c/nodejs/awsx/ecs/taskDefinition.ts#L152

so to be able replace it in pulumi-awsx, I added missing ECS managed policies here first


ahanoff:
> Run aws-actions/configure-aws-credentials@v1
Error: Credentials could not be loaded, please check your action inputs: Could not load credentials from any providers

@komalali tests are failing because of error above. :thinking: could you please check?

sdk/python/pulumi_aws/iam/_enums.py
```diff
@@ -151,6 +151,9 @@ class ManagedPolicy(str, Enum):
     AMAZON_EC2_SPOT_FLEET_AUTOSCALE_ROLE = "arn:aws:iam::aws:policy/service-role/AmazonEC2SpotFleetAutoscaleRole"
     AMAZON_EC2_SPOT_FLEET_ROLE = "arn:aws:iam::aws:policy/service-role/AmazonEC2SpotFleetRole"
     AMAZON_EC2_SPOT_FLEET_TAGGING_ROLE = "arn:aws:iam::aws:policy/service-role/AmazonEC2SpotFleetTaggingRole"
+    AMAZON_EC_S_FULL_ACCESS = "arn:aws:iam::aws:policy/AmazonECS_FullAccess"
```
ahanoff, line 154: > AMAZON_EC_S_FULL_ACCESS

@komalali another concern is this name for python enum. Is it expected?

komalali: Yeah, this is an unfortunate side-effect of the camel case to snake case algorithm.  I updated this enum name to `AmazonECSFullAccess` which creates the desired python enum.

komalali:
Updated and opened https://github.com/pulumi/pulumi-aws/compare/pr/1237 with an improvement.

PR #1277, opened 2020-12-21
Upgrade to v3.22.0 of the AWS Terraform Provider

stack72:


provider/resources.go
```diff
@@ -2387,7 +2392,24 @@ func Provider() tfbridge.ProviderInfo {
 			"aws_signer_signing_profile_permission": {Tok: awsResource(signerMod, "SigningProfilePermission")},
 
 			// ServerlessRepository
-			"aws_serverlessrepository_stack": {Tok: awsResource(serverlessRepositoryMod, "Stack")},
+			"aws_serverlessapplicationrepository_cloudformation_stack": {
+				Tok: awsResource(serverlessRepositoryMod, "CloudformationStack"),
```
mikhailshilkov, line 2396: Should it be `CloudFormationStack` as per branding?

mikhailshilkov (approved):
One naming question, LGTM otherwise

provider/resources.go
stack72: Great catch! thanks for that - fixing up now

Issue #1275, opened 2020-12-21
Two lambda managed polices being deprecated


pierskarsenbarg:
Received an email this morning from AWS about two lambda policies being deprecated and replaced by (what they describe as) "scoped down policies":

* AWSLambdaFullAccess - being replaced by [AWSLambda_FullAccess](https://console.aws.amazon.com/iam/home#/policies/arn:aws:iam::aws:policy/AWSLambda_FullAccess)
* AWSLambdaReadOnlyAccess - being replaced by [AWSLambda_ReadOnlyAccess](https://console.aws.amazon.com/iam/home#/policies/arn:aws:iam::aws:policy/AWSLambda_ReadOnlyAccess)

Looks like nothing will change to resources that are already using these policies but they can't be attached to resources going forward.

Email text:

```
Hello,

AWS Lambda provides managed policies that grant access to Lambda management API actions, and in some cases, access to other services used to develop and manage Lambda resources. AWS routinely performs audits to ensure that managed policies are appropriately scoped for their intended purposes.

As a result of a recent audit, Lambda will deprecate two managed policies called "AWSLambdaFullAccess" and "AWSLambdaReadOnlyAccess" on January 25, 2021.

The old policies will be marked as deprecated with a warning icon next to them in the "Policies" list within the IAM console. Following deprecation, these policies will have following characteristics:

1. They will continue to work for all currently attached users, groups, and roles. Nothing will break.
2. They cannot be attached to any new users, groups, or roles. If you detach one of the policies from a current entity, you cannot reattach it.
3. After you detach the policies from all current entities, the policies will no longer be visible and can no longer be used in any way.

Lambda has introduced two new scoped down policies called "AWSLambda_FullAccess" [1] and "AWSLambda_ReadOnlyAccess" [2]. We recommend you use the new policies to grant entities access to Lambda. Additionally, you can create your own policies for granting additional permissions or for more fine-grained control [3].

If you have any questions, contact AWS Support [4].

[1] https://console.aws.amazon.com/iam/home#/policies/arn:aws:iam::aws:policy/AWSLambda_FullAccess
[2] https://console.aws.amazon.com/iam/home#/policies/arn:aws:iam::aws:policy/AWSLambda_ReadOnlyAccess
[3] https://docs.aws.amazon.com/lambda/latest/dg/access-control-identity-based.html
[4] https://aws.amazon.com/support

Sincerely,
Amazon Web Services
```

PR #1303, opened 2021-01-14
Update aws.lambda.CallbackFunction to create a replica of AWSLambdaFullAccess

stack72:
Fixes: #1275

This will ensure that `aws.sqs.QueueEventSubscription` and
`aws.kinesis.StreamEventSubscription` continue to work now that
AWSLambdaFullAccess has been deprecated.

This will manifest, to existing users, as a policy removal and a new
policy attachment


sdk/python/pulumi_aws/iam/_enums.py
```diff
@@ -79,9 +79,11 @@ class ManagedPolicy(str, Enum):
     AWS_LAMBDA_ENI_MANAGEMENT_ACCESS = "arn:aws:iam::aws:policy/service-role/AWSLambdaENIManagementAccess"
     AWS_LAMBDA_EXECUTE = "arn:aws:iam::aws:policy/AWSLambdaExecute"
     AWS_LAMBDA_FULL_ACCESS = "arn:aws:iam::aws:policy/AWSLambdaFullAccess"
+    AWS_LAMBDA_FULL_ACCESS = "arn:aws:iam::aws:policy/AWSLambda_FullAccess"
     AWS_LAMBDA_INVOCATION_DYNAMO_DB = "arn:aws:iam::aws:policy/AWSLambdaInvocation-DynamoDB"
     AWS_LAMBDA_KINESIS_EXECUTION_ROLE = "arn:aws:iam::aws:policy/service-role/AWSLambdaKinesisExecutionRole"
     AWS_LAMBDA_READ_ONLY_ACCESS = "arn:aws:iam::aws:policy/AWSLambdaReadOnlyAccess"
+    AWS_LAMBDA_READ_ONLY_ACCESS = "arn:aws:iam::aws:policy/AWSLambda_ReadOnlyAccess"
```
komalali, line 86: Hmm... the python representation of the two enums is coming out identical. This won't work... not totally sure what we should do here.

sdk/nodejs/lambda/lambdaMixins.ts
```diff
@@ -281,7 +281,7 @@ export class CallbackFunction<E, R> extends LambdaFunction {
 
             if (!args.policies) {
                 // Provides wide access to "serverless" services (Dynamo, S3, etc.)
```
lukehoban, line 283: This comment is not longer true.  The new policy doesn't provide either of these.  This will be a significant breaking change for users of this (in particular in AWSX and Cloud packages).

sdk/nodejs/kinesis/kinesisMixins.ts
```diff
@@ -151,7 +151,7 @@ function createFunctionFromEventHandler(
     if (handler instanceof Function) {
         return new lambda.CallbackFunction(name, {
             callback: handler,
-            policies: [iam.ManagedPolicy.AWSLambdaFullAccess, iam.ManagedPolicy.AWSLambdaKinesisExecutionRole]
+            policies: [iam.ManagedPolicy.AWSLambdaKinesisExecutionRole]
```
leezen, line 154: This looks breaking since the role will have fewer permissions than previously. I think this should create a copy of `AWSLambdaFullAccess`, then attach that new policy plus the `AWSLambdaKinesisExecutionRole` policy here.

sdk/nodejs/lambda/lambdaMixins.ts
```diff
@@ -254,6 +254,8 @@ export function createFunctionFromEventHandler<E, R>(
  * dependencies) into a form that can be used by AWS Lambda.  See
  * https://www.pulumi.com/docs/tutorials/aws/serializing-functions/ for additional
  * details on this process.
+ * This CallbackFunction will automatically create an IAM Policy that replicates the existing AWSLambdaFullAccess
+ * if no IAM Role is specified
```
leezen, line 258:
```suggestion
 * If no IAM Role is specified, CallbackFunction will automatically create an IAM Policy that replicates the now-deprecated AWSLambdaFullAccess managed policy. We do this to avoid breaking existing users who have come to rely on that set of permissions. In the next major version, we will remove this functionality and rely on the  `AWSLambda_FullAccess` managed policy.
```

sdk/nodejs/lambda/lambdaMixins.ts
```diff
@@ -279,19 +281,87 @@ export class CallbackFunction<E, R> extends LambdaFunction {
...
-                // RolePolicyAttachment objects don't have a physical identity, and create/deletes are processed
-                // structurally based on the `role` and `policyArn`.  So we need to make sure our Pulumi name matches the
-                // structural identity by using a name that includes the role name and policyArn.
-                const attachment = new iam.RolePolicyAttachment(`${name}-${utils.sha1hash(policy)}`, {
-                    role: role,
-                    policyArn: policy,
-                }, opts);
+            // we are going to create a copy of the "arn:aws:iam::aws:policy/AWSLambdaFullAccess" policy to ensure
+            // our mixins continue to work as expected
+            const lambdaFullAccessCopy = new iam.Policy(`${name}-LambdaFullAccess`, {
```
leezen, line 286: This is going to create the policy no matter what. Instead, this creation should be within the conditional that previously existed `if (!args.policies)`. So this should only be created and attached if no policies are provided to the `CallbackFunction`.

sdk/nodejs/sqs/sqsMixins.ts
```diff
@@ -104,7 +104,7 @@ function createFunctionFromEventHandler(
     if (handler instanceof Function) {
         return new lambda.CallbackFunction(name, {
             callback: handler,
-            policies: [iam.ManagedPolicy.AWSLambdaFullAccess, iam.ManagedPolicy.AmazonSQSFullAccess]
+            policies: [iam.ManagedPolicy.AmazonSQSFullAccess]
```
leezen, line 107: Similar to my comment about Kinesis. So I think we should clean this up by encapsulating the full copy of the policy somewhere common to refer to, then refer to it here, in the Kinesis mixin, and in the Lambda mixin.

sdk/nodejs/kinesis/kinesisMixins.ts
stack72: No, the use of the LambdaFullAccess happens inside the CallbackFunction and each of SQS and Kinesis calls

leezen: Right, but my point is that `CallbackFunction` should not indiscriminately create the `AWSLambdaFullAccess` policy. It should only be created when `policies` is not specified (as it currently works in the package).

CHANGELOG.md
```diff
@@ -2,7 +2,20 @@ CHANGELOG
...
+* Update auto-naming rules for RDS Cluster `clusterIdentifier` to follow specifications.
+* Add support for `LambdaFullAccess` as a replacement for `AWSLambdaFullAccess` IAM policy.
+* Add support for `LambdaReadOnlyAccess` as a replacement for `AWSLambdaReadOnlyAccess` IAM policy.
+* Update `aws.lambda.CallbackFunction` to accept a new parameter `needsServicesFullAccess`. If this parameter
+  is set or if no iam.Policies are set to the CallbackFunction then we will create an IAM Policy that replicates
+  the now deprecated `AWSLambdaFullAccess` policy.
+  
+
+
+to create an IAM Policy that replicates `AWSLambdaFullAccess` to ensure
```
leezen, line 16: Did something get cut by accident?

sdk/nodejs/iam/managedPolicies.ts
```diff
@@ -153,7 +153,7 @@ export module ManagedPolicies {
     export const AWSLambdaENIManagementAccess: ARN = "arn:aws:iam::aws:policy/service-role/AWSLambdaENIManagementAccess";
     /** Use ManagedPolicy.AWSLambdaExecute instead. */
     export const AWSLambdaExecute: ARN = "arn:aws:iam::aws:policy/AWSLambdaExecute";
-    /** Use ManagedPolicy.AWSLambdaFullAccess instead. */
+    /** @deprecated No longer supported. New IAM Roles created using this policy will fail. Use ManagedPolicy.AWSLambda_FullAccess instead. */
```
leezen, line 156:
```suggestion
    /** @deprecated No longer supported. New IAM Roles created using this policy will fail. Use ManagedPolicy.LambdaFullAccess instead. */
```

leezen: I also noticed we don't do the same thing for `AWSLambdaReadOnlyAccess` which should also be updated a few lines down in this file.

sdk/nodejs/lambda/lambdaMixins.ts
```diff
@@ -216,6 +216,29 @@ export interface CallbackFunctionArgs<E, R> extends BaseCallbackFunctionArgs {
...
+     * - iot
+     * - kinesis
+     * - lambda
+     * - logs
+     * - s3
+     * - sns
+     * - sqs
+     * - xray
+     */
+    needsServicesFullAccess?: boolean;
```
leezen, line 241: I can't say I'm a huge fan of this name, but I also don't know if I have any better suggestions. One thought would be to name this more closely to the original intent, so something like `includeDeprecatedLambdaFullAccessPolicy` to be more descriptive?

leezen (approved):
Generally looks good. Thanks for picking this up. If you could please make sure to cleanup the `CHANGELOG` and consider the boolean name.

sdk/nodejs/lambda/lambdaMixins.ts
stack72: Happy to make this change!

Issue #1321, opened 2021-01-23
AnalyticsApplicationOutput does not implement pulumi.Output (missing Apply method)


jslauthor:
Hello all. I'm receiving an error when creating a stream with the Kinesis library. This happened around the `v2.18.2` SDK release (it was working prior) when running `pulumi up` or `go get`:

```
github.com/pulumi/pulumi-aws/sdk/v3/go/aws/kinesis
../../go/pkg/mod/github.com/pulumi/pulumi-aws/sdk/v3@v3.25.0/go/aws/kinesis/analyticsApplication.go:274:43: impossible type assertion:
	AnalyticsApplicationOutput does not implement pulumi.Output (missing Apply method)
../../go/pkg/mod/github.com/pulumi/pulumi-aws/sdk/v3@v3.25.0/go/aws/kinesis/pulumiTypes.go:2204:6: AnalyticsApplicationOutput redeclared in this block
	previous declaration at ../../go/pkg/mod/github.com/pulumi/pulumi-aws/sdk/v3@v3.25.0/go/aws/kinesis/analyticsApplication.go:356:6
```
and here's the stream code:

```go
  stream, err := kinesis.NewStream(ctx, "aws.chat.cmd.rooms.0", &kinesis.StreamArgs{
	  RetentionPeriod: pulumi.Int(24),
	  ShardCount: pulumi.Int(1),
	  ShardLevelMetrics: pulumi.StringArray{
		  pulumi.String("IncomingBytes"),
		  pulumi.String("OutgoingBytes"),
	  },
	  Tags: pulumi.StringMap{
			  "Environment": pulumi.String("staging"),
	  },
  })
  if err != nil {
	  return err
  }
```

I've tried downgrading to `2.18.0`, but I received the same error. I've and tried `go clean --modcache && go get` to no avail. Any ideas?

mikhailshilkov:
@EvanBoyle could you triage this one?

EvanBoyle:
related to https://github.com/pulumi/pulumi/issues/6170 and https://github.com/pulumi/pulumi/pull/6128

This is a schematype colliding with generated types (codegen error). @viveklak can you take a look?

EvanBoyle:
@jslauthor this is a compilation error in the aws go SDK due to type collision. Was this working for you previously? If so, can you try downgrading the version of the AWS SDK you're using?

jslauthor:
[...]

PR #1335, opened 2021-01-31
Bump pulumi/pulumi to pickup codegen changes

viveklak:
Fixes https://github.com/pulumi/pulumi-aws/issues/1321


stack72:
@viveklak is this a breaking change?

viveklak:
> @viveklak is this a breaking change?

Depends.. The original breaking change was introduced in https://github.com/pulumi/pulumi-aws/pull/1224 (v3.13.0). This "fixes" it by renaming a schema type which did exist before 3.13.0.

viveklak:
@stack72 I have verified the change shouldn't result in any resource recreation and fleshed out the changelog with more details.

stack72 (approved):

PR #1356, opened 2021-02-12
Upgrade to v3.28.0 of the AWS Terraform Provider

stack72:
Fixes: #1355


provider/resources.go
```diff
@@ -668,6 +669,8 @@ func Provider() tfbridge.ProviderInfo {
 			"aws_cloudfront_public_key":             {Tok: awsResource(cloudfrontMod, "PublicKey")},
 			"aws_cloudfront_origin_access_identity": {Tok: awsResource(cloudfrontMod, "OriginAccessIdentity")},
 			"aws_cloudfront_origin_request_policy":  {Tok: awsResource(cloudfrontMod, "OriginRequestPolicy")},
+			"aws_cloudfront_cache_policy":           {Tok: awsResource(cloudfrontMod, "CachePolicy")},
+			"aws_cloudfront_realtime_log_config":    {Tok: awsResource(cloudfrontMod, "RealTimeLogConfig")},
```
mikhailshilkov, line 673: CF spells it as `RealtimeLogConfig` - I suggest we use that capitalization

mikhailshilkov (commented):
One suggestion - LGTM otherwise

provider/resources.go
stack72: Good catch! Updated

mikhailshilkov (approved):

PR #1357, opened 2021-02-15
Upgrade to v2.19.0 of the pulumi-terraform-bridge

stack72:


stack72:
Self merging as largely mechanical

Issue #1352, opened 2021-02-11
Some top-level EC2 functions are in the aws namespace (TypeScript)


tenwit:
Some functions built from data sources in Terraform's EC2 package have been put in pulumi-aws' top-level namespace.
I'm not au fait with all the Terraform modules and data sources, but I see at least three: Terraform's `aws_ami`, `aws_ami_ids` and `aws_eip` have become Pulumi's `aws.getAmi`, `aws.getAmiIds` and `aws.getElasticIp`. I expect to see these functions in the ec2 namespace: `aws.ec2.getAmi`, `aws.ec2.getAmiIds` and `aws.ec2.getElasticIp`.

There are only 9 top-level data sources in the Terraform AWS module, but currently 15 functions show up in VSCode when I view the IntelliSense pop-up after typing `aws.get`. So there may be  a few more errant functions to track down..

pgavlin:
I think that we could duplicate the function definitions into the `ec2` namespace (removing them from the top-level namespace would be a breaking change that we don't want to take). I'm not entirely sure that we have support for that in `pulumi-terraform-bridge`.

PR #1365, opened 2021-02-24
Move some datasources from top level package to ec2, autoscaling and s3 packages

stack72:
Fixes: #1352


mikhailshilkov (approved):
LGTM but we need to fix the failing test by specifying the full namespace

PR #1393, opened 2021-03-11
Upgrade to v2.20.0 of the pulumi-terraform-bridge

stack72:


mikhailshilkov (approved):
LGTM except for the disappeared examples, let's follow up to see why they are gone.

PR #1417, opened 2021-03-26
Upgrade to v3.34.0 of the AWS Terraform Provider

stack72:
Fixes: #1416


sdk/dotnet/Iam/ServerCertificate.cs
```diff
@@ -229,12 +247,6 @@ public static ServerCertificate Get(string name, Input<string> id, ServerCertifi
 
     public sealed class ServerCertificateArgs : Pulumi.ResourceArgs
     {
-        /// <summary>
-        /// The Amazon Resource Name (ARN) specifying the server certificate.
-        /// </summary>
-        [Input("arn")]
-        public Input<string>? Arn { get; set; }
```
mikhailshilkov, line 236: What happened here?

stack72: An ARN should never be an input argument - this is expected but I am not sure why it was ever there in the first place

mikhailshilkov (approved):

PR #1426, opened 2021-04-01
Correction to security group warning

sibuthomasmathew:
The warning message thrown when using the python sdk instance
resource property `security_group`, recommended arguments that
neither worked nor matched with the documentation.


komalali:
As discussed separately, the documentation for our providers (along with the code) is auto-generated, so any changes to the SDK code will be overwritten in a future update. Those doc comments come directly from the upstream provider, so it’s difficult to change them for python (which insists on snake case) specifically. We’ve done our best to align with python practices by converting to snake case where possible but comments are definitely difficult since they are just text 🙂

PR #1465, opened 2021-04-26
Add missing enum for policy

aamir-locus:


stack72 (commented):
hey @aamir-locus 

Thank you so much for adding this - these enums are actually autogenerated from here https://github.com/pulumi/pulumi-aws/blob/master/provider/resources.go#L3248 

I will take this code and retrofit it to the other sdks

stack72:
Hi @aamir-locus 

As you can see, I moved this into #1466 and made sure you got the attribution for the work here - I am going to close this out. Thanks so much for the work here!

Paul

PR #1494, opened 2021-05-18
add app sync principal for nodejs sdk

marksmithson:
Adds Principal for AWS AppSync to nodejs SDK.

This is required when creating a DataSource for AppSync.

https://docs.aws.amazon.com/appsync/latest/devguide/attaching-a-data-source.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,7 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
-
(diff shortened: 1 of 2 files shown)


stack72:
Hi @marksmithson 

Thanks so much for the work here - I am going to bring this into another PR that I am currently working on. We changed how enums are added in pulumi to be via provider/resources.go to ensure they are available for all languages

I will make sure and attribute your contribution and link the commit here shortly

Paul

stack72 (approved):

PR #1514, opened 2021-06-02
Upgrade to v3.43.0 of the AWS Terraform Provider

stack72:
Fixes: #1513﻿
Fixes: #1515


stack72:
@mikhailshilkov this is an expected "breaking" change to the data source. The data source originally had an input for logs when it should only have been an output

provider/resources.go
```diff
@@ -2122,6 +2130,16 @@ func Provider() tfbridge.ProviderInfo {
 				Tok: awsResource(servicecatalogMod, "ProductPortfolioAssociation"),
 			},
 			"aws_servicecatalog_service_action": {Tok: awsResource(servicecatalogMod, "ServiceAction")},
+			"aws_servicecatalog_budget_resource_association": {
+				Tok: awsResource(servicecatalogMod, "BudgetResourceApplication"),
```
mikhailshilkov, line 2134: Should it be
```suggestion
				Tok: awsResource(servicecatalogMod, "BudgetResourceAssociation"),
```
?

stack72: yes, great catch! Fixing now

mikhailshilkov (approved):
LGTM. Thank you for cleaning up all the examples!

Issue #1519, opened 2021-06-07
Python SDK uses duplicate argument in `servicecatalog/tag_option_resource_association.py`


csssuf:
The Python SDK has a function with a duplicate argument (`resource_name`) in `servicecatalog/tag_option_resource_association.py`:

https://github.com/pulumi/pulumi-aws/blob/e17984805f03c02f7b0b6433c585e3b71b7846d8/sdk/python/pulumi_aws/servicecatalog/tag_option_resource_association.py#L270-L278

This is an unrecoverable error for mypy, which means any projects depending on pulumi-aws can no longer run mypy against pulumi-aws 4.7.0.

## Steps to reproduce
1. `mypy sdk/python`

Expected:
mypy passes.

Actual:
```
sdk/python/pulumi_aws/servicecatalog/tag_option_resource_association.py:277: error: Duplicate argument 'resource_name' in function definition
Found 1 error in 1 file (errors prevented further checking)
```

stack72:
Hi @csssuf 

I will get this fixed up and a new release cut!

Sorry this happened

Paul

lukehoban:
@stack72 Are we running `mypy` on the generated SDKs in our provider repos?  If not, perhaps that's worth adding to catch this earlier?

stack72:
we are not - that is definitely something I will add here!

tlinhart:
Hi guys, what's the status of this issue, please? We are using https://github.com/tlinhart/pulumi-aws-tags extensively for autotagging our AWS resources and the library crashes due to this bug prohibiting us from provisioning the resources during CI/CD.

PR #1520, opened 2021-06-07
[breaking] Remap `aws.serviceCatalog.TagOptionResourceAssociation` `resourceName` to be `associatedResourceName`

stack72:
Fixes: #1519

`resourceName` is a reserved name and this is causing a break in
our python sdk

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,9 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
-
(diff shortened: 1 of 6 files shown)


leezen (approved):

lukehoban:
An alternative would be to fix the Python code-generator to map a `resourceName` property to a name that doesn't conflict - like `resource_name_`. That is probably a better fix here - as it addresses the problem generally and doesn't impact non-Python users.

PR #1525, opened 2021-06-10
Warn when a secret is captured by an `aws.lambda.CallbackFunction`

### LINKED ISSUE #1524: Warn when a secret is captured by a aws.lambda.CallbackFunction
In https://github.com/pulumi/pulumi-aws/pull/1282 we added the ability for CallbackFunctions which captured a secret value to turn the entire captured function body into a Pulumi secret so that the sensitive data would be encrypted in Pulumi state.  

However, the sensitive values in this case would still be captured into the text of the Lambda function visibile inside AWS.  Although this is standard and expected behaviour for Pulumi secrets generally, it is also an easy way to "accidentally" leak a secret into a context that may not be trusted.  

We should consider warning when this happens, and using the opportunity to suggest one of the alternatives, ideally SecretsManager, but even environment variables would be "better".

lukehoban:
Fixes #1524

### DIFF
--- a/sdk/nodejs/lambda/lambdaMixins.ts
+++ b/sdk/nodejs/lambda/lambdaMixins.ts
@@ -340,10 +340,10 @@ export class CallbackFunction<E, R> extends LambdaFunction {
         (<any>code).isSecret = closure.then(c => {
             if (c.containsSecrets) {
                 pulumi.log.warn(`A secret value was captured and serialized into the body of the Lambda Function '${name}'. ` +
-                                `This value will be stored as an encyrpted Pulumi secret, but may be available in plain text ` +
+                                `This value will be stored as an encrypted Pulumi secret, but may be available in plain text ` +
                                 `inside the AWS deployment package. You can use 'pulumi.unsecret' to convert the value to a non-secret ` +
                                 `value if this is not a sensitive value, or else use Secrets Manager or environment variables to pass the ` +
-                                `senstive data to your function.`, this);
+                                `sensitive data to your function.`, this);
                 return true;
             }
             return false;
--- a/sdk/nodejs/lambda/lambdaMixins.ts
+++ b/sdk/nodejs/lambda/lambdaMixins.ts
@@ -337,7 +337,17 @@ export class CallbackFunction<E, R> extends LambdaFunction {
             closure, serializedFileNameNoExtension, args.codePathOptions);
 
         const code = pulumi.output(new pulumi.asset.AssetArchive(codePaths));
-        (<any>code).isSecret = closure.then(c => c.containsSecrets);
+        (<any>code).isSecret = closure.then(c => {
+            if (c.containsSecrets) {
+                pulumi.log.warn(`A secret value was captured and serialized into the body of the Lambda Function '${name}'. ` +
+                                `This value will be stored as an encyrpted Pulumi secret, but may be available in plain text ` +
+                                `inside the AWS deployment package.  You can use 'pulumi.unsecret' to convert the value to a non-secret ` +
+                                `value if this is not a sensitive value, or else use Secrets Manager or environment variables to pass the` +
+                                `senstive data to your function.`)
+                return true;
+            }
+            return false;
+        });
 
         // Copy over all option values into the function args.  Then overwrite anything we care
         // about with our own values.  This ensures that clients can pass future supported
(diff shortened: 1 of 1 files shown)


sdk/nodejs/lambda/lambdaMixins.ts
```diff
@@ -337,7 +337,17 @@ export class CallbackFunction<E, R> extends LambdaFunction {
             closure, serializedFileNameNoExtension, args.codePathOptions);
 
         const code = pulumi.output(new pulumi.asset.AssetArchive(codePaths));
-        (<any>code).isSecret = closure.then(c => c.containsSecrets);
+        (<any>code).isSecret = closure.then(c => {
+            if (c.containsSecrets) {
+                pulumi.log.warn(`A secret value was captured and serialized into the body of the Lambda Function '${name}'. ` +
+                                `This value will be stored as an encyrpted Pulumi secret, but may be available in plain text ` +
+                                `inside the AWS deployment package.  You can use 'pulumi.unsecret' to convert the value to a non-secret ` +
```
cnunciato, line 344:
```suggestion
                                `inside the AWS deployment package. You can use 'pulumi.unsecret' to convert the value to a non-secret ` +
```

sdk/nodejs/lambda/lambdaMixins.ts
```diff
@@ -337,7 +337,17 @@ export class CallbackFunction<E, R> extends LambdaFunction {
...
+                                `value if this is not a sensitive value, or else use Secrets Manager or environment variables to pass the` +
```
cnunciato, line 345: We just need one space here, I think:

```suggestion
                                `value if this is not a sensitive value, or else use Secrets Manager or environment variables to pass the ` +
```

For reference here's how this renders out:

image

cnunciato (commented):
Couple super minor things I noticed when trying this out.

cnunciato (approved):

joeduffy:
My only comment on warnings/errors like this is that, when confronted with similar things in Pulumi, I always wish we had a stack trace. It's often very difficult to pinpoint where exactly the resource/etc was being created, especially since it can be buried in layers of libraries. Not sure if there's a "systemic" way to fix that but wanted to mention in case others agree.

lukehoban:
> My only comment on warnings/errors like this is that, when confronted with similar things in Pulumi, I always wish we had a stack trace. It's often very difficult to pinpoint where exactly the resource/etc was being created, especially since it can be buried in layers of libraries. Not sure if there's a "systemic" way to fix that but wanted to mention in case others agree.

Agreed with this.  I've updated to at least log the warning against the impacted resource.  Ideally we'd also include a "stack trace" or more likely a line of code, for where the secret was captured - but that would require a quite significant new investment in the closure serialization implementation.  On balance, I'm inclined to take this change without doing that now - and consider it in the future if this is an area of feedback. But I'm open to holding off on making any change here for now, and waiting until we can invest in the new closure serialization machinery if anyone feels that is preferable.

sdk/nodejs/lambda/lambdaMixins.ts
```diff
@@ -337,7 +337,17 @@ export class CallbackFunction<E, R> extends LambdaFunction {
...
 
         const code = pulumi.output(new pulumi.asset.AssetArchive(codePaths));
-        (<any>code).isSecret = closure.then(c => c.containsSecrets);
+        (<any>code).isSecret = closure.then(c => {
+            if (c.containsSecrets) {
+                pulumi.log.warn(`A secret value was captured and serialized into the body of the Lambda Function '${name}'. ` +
+                                `This value will be stored as an encyrpted Pulumi secret, but may be available in plain text ` +
+                                `inside the AWS deployment package. You can use 'pulumi.unsecret' to convert the value to a non-secret ` +
+                                `value if this is not a sensitive value, or else use Secrets Manager or environment variables to pass the ` +
+                                `senstive data to your function.`, this);
```
cnunciato, line 346:
```suggestion
                                `sensitive data to your function.`, this);
```

sdk/nodejs/lambda/lambdaMixins.ts
```diff
@@ -337,7 +337,17 @@ export class CallbackFunction<E, R> extends LambdaFunction {
             closure, serializedFileNameNoExtension, args.codePathOptions);
 
         const code = pulumi.output(new pulumi.asset.AssetArchive(codePaths));
-        (<any>code).isSecret = closure.then(c => c.containsSecrets);
+        (<any>code).isSecret = closure.then(c => {
+            if (c.containsSecrets) {
+                pulumi.log.warn(`A secret value was captured and serialized into the body of the Lambda Function '${name}'. ` +
+                                `This value will be stored as an encyrpted Pulumi secret, but may be available in plain text ` +
```
cnunciato, line 343:
```suggestion
                                `This value will be stored as an encrypted Pulumi secret, but may be available in plain text ` +
```

cnunciato (commented):
Apologies for not noticing these earlier!

cnunciato:
Are we thinking we'll merge this?

Zaid-Ajaj (approved):

PR #1545, opened 2021-06-24
Changed `egress` to `ingress`

alexbowers:
There is also the description field on `SecurityGroupIngress` which says: `Description of this egress rule.`. But i was unable to find where this came from.

https://www.pulumi.com/docs/reference/pkg/aws/ec2/securitygroup


leezen:
@alexbowers Thanks for noticing the typo and submitting a PR! These come from the upstream docs and should be fixed there: https://github.com/pulumi/terraform-provider-aws/blob/upstream-v3.47.0/website/docs/r/security_group.html.markdown -- you could submit a PR against our branch or I would suggest submitting against the upstream provider (https://github.com/hashicorp/terraform-provider-aws) and we can pick up the change from there.

PR #1554, opened 2021-06-29
[NodeJS] Add environmentFile to container definition type

leezen:
Part of https://github.com/pulumi/pulumi-awsx/issues/682


thenovelnomad:
Shouldn't the type of the EnvironmentFile.value parameter be wrapped by `Input<>`?

PR #1568, opened 2021-07-09
Update pulumi deps.

pgavlin:


provider/resources.go
```diff
@@ -3556,7 +3556,7 @@ func Provider() tfbridge.ProviderInfo {
 				ObjectTypeSpec: schema.ObjectTypeSpec{
 					Type: "string",
 				},
-				Enum: []*schema.EnumValueSpec{
```
t0yv0, line 3559: I'd like these changes on master please so it's easier to compile pulumi-aws against local pulumi :)

sdk/go/aws/apigateway/stage.go
```diff
@@ -156,7 +156,7 @@ type stageState struct {
 	// The identifier of a client certificate for the stage.
 	ClientCertificateId *string `pulumi:"clientCertificateId"`
 	// The ID of the deployment that the stage points to
-	Deployment *string `pulumi:"deployment"`
+	Deployment interface{} `pulumi:"deployment"`
```
t0yv0, line 159: This looks a bit concerning, less precise type or is intentional?

pgavlin: This is intentional--the original type was incorrect with respect to the actual types of the state inputs.

sdk/dotnet/Outputs/ProviderAssumeRole.cs
```diff
@@ -1,53 +0,0 @@
-// *** WARNING: this file was generated by the Pulumi Terraform Bridge (tfgen) Tool. ***
-// *** Do not edit by hand unless you're certain you know what you are doing! ***
-
-using System;
-using System.Collections.Generic;
-using System.Collections.Immutable;
-using System.Threading.Tasks;
-using Pulumi.Serialization;
-
-namespace Pulumi.Aws.Outputs
```
t0yv0, line 10: Interestingly enough this was not referenced anywhere so a safe deletion?

sdk/dotnet/S3/Inputs/BucketWebsiteGetArgs.cs
```diff
@@ -35,7 +35,7 @@ public sealed class BucketWebsiteGetArgs : Pulumi.ResourceArgs
         /// describing redirect behavior and when redirects are applied.
         /// </summary>
         [Input("routingRules")]
-        public InputUnion<string, ImmutableArray<string>>? RoutingRules { get; set; }
+        public InputUnion<string, ImmutableArray<Input<string>>>? RoutingRules { get; set; }
```
t0yv0, line 38: We can probably live with this. But the current .NET SDK  can't equate/autoconvert between these two things as in:

```
        ImmutableArray<string> a;
        ImmutableArray<Input<string>> b;

        b = a;

        // MyStack.cs(15,13): error CS0029: Cannot implicitly convert type 'System.Collections.Immutable.ImmutableArray<string>' to 'System.Collections.Immutable.ImmutableArray<Pulumi.Input<string>>' [/Users/anton/tmp/play-cs-pulumi/play-cs-pulumi.csproj]
```

t0yv0 (approved):
Looking good to me: Go inputs landed, added docs for functions, some unused? output types removed. Added a few comments on some changes that are not obviously bw-compatible.

stack72 (approved):
Not going to block this but there's an outstanding question on the new c# class numbers

komalali:
@stack72 can you open an issue to track the C# class numbers that you're referring to please so we can prioritize with all the other issues?

PR #1637, opened 2021-09-22
Upgrade to pulumi-terraform-bridge v3.8.0

stack72:


komalali (approved):

EvanBoyle (approved):
The Go types appear to be back in their original form!

Can we enable a `go build` of the SDK folder in this repo like we do with some of our native providers now that this issue is fixed?

stack72:
> The Go types appear to be back in their original form!
> 
> 
> 
> Can we enable a `go build` of the SDK folder in this repo like we do with some of our native providers now that this issue is fixed? 

It's already in the PR 🤣

PR #1640, opened 2021-09-24
Upgrade to v3.60.0 of the AWS Terraform Provider

stack72:
Fixes: #1638﻿


provider/resources.go
```diff
@@ -1455,6 +1465,7 @@ func Provider() tfbridge.ProviderInfo {
 			"aws_fsx_lustre_file_system":  {Tok: awsResource(fsxMod, "LustreFileSystem")},
 			"aws_fsx_windows_file_system": {Tok: awsResource(fsxMod, "WindowsFileSystem")},
 			"aws_fsx_backup":              {Tok: awsResource(fsxMod, "Backup")},
+			"aws_fsx_ontap_file_system":   {Tok: awsResource(fsxMod, "OnTapFilesystem")},
```
mikhailshilkov, line 1468: Should it be `OntapFileSystem` : similar to `WindowsFileSystem` above and I think "ONTAP" is a single acronym or brand name (I'm able to find `OntapVolumeConfiguration` in AWS C# SDK?

mikhailshilkov (commented):
One naming suggestion, LGTM otherwise

provider/resources.go
stack72: You are correct here! I will update it - good suggestion

stack72: Updated now @mikhailshilkov

mikhailshilkov (approved):

PR #1647, opened 2021-10-04
Upgrade to v3.61.0 of the AWS Terraform Provider

stack72:
Fixes: #1646﻿


provider/resources.go
```diff
@@ -700,6 +701,8 @@ func Provider() tfbridge.ProviderInfo {
 			"aws_chime_voice_connector_streaming":   {Tok: awsResource(chimeMod, "VoiceConnectorStreaming")},
 			// Cloud9
 			"aws_cloud9_environment_ec2": {Tok: awsResource(cloud9Mod, "EnvironmentEC2")},
+			//Cloud Control API
```
mikhailshilkov, line 704: Uber-nit :)
```suggestion
			// Cloud Control API
```

provider/resources.go
```diff
@@ -1269,8 +1272,12 @@ func Provider() tfbridge.ProviderInfo {
...
 				Tok: awsResource(ec2Mod, "LocalGatewayRouteTableVpcAssociation"),
 			},
-			"aws_ec2_tag":                       {Tok: awsResource(ec2Mod, "Tag")},
-			"aws_dedicated_host":                {Tok: awsResource(ec2Mod, "DedicatedHost")},
+			"aws_ec2_tag": {Tok: awsResource(ec2Mod, "Tag")},
+			"aws_dedicated_host": {
+				Tok:                awsResource(ec2Mod, "DedicatedHost"),
+				DeprecationMessage: "This resource has been deprecated in favour of `aws.ec2.Ec2DedicatedHost`",
+			},
+			"aws_ec2_host":                      {Tok: awsResource(ec2Mod, "Ec2DedicatedHost")},
```
mikhailshilkov, line 1280: Why not just `Host`?

stack72: We have decided to be able to use the DedicatedHost name here as the APIs are identical moving between versions of the resource

mikhailshilkov: That's the best outcome then 👍

PR #1702, opened 2021-11-10
Export constant with nodejs 14.x for lambda runtimes configuration

rapzo:
[Runtimes](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html)

The string `"nodejs14.x"` works fine tho.


stack72:
Hi @rapzo 

Sorry for the time taken to get back to you here - I am going to close this out as the enums changed here are now deprecated and you can use `aws.lambda.NodeJS14dXRuntime`

Paul

PR #1711, opened 2021-11-18
Update examples to use Pulumi 3.17.0

jkodroff:


jkodroff:
Closing: Not on the critical path, and the repo has diverged significantly since the time this PR was created. We can come back to this later.

PR #1725, opened 2021-12-02
Upgrade to v3.68.0 of the AWS Terraform Provider

stack72:
Fixes: #1724﻿


sdk/dotnet/Iot/ThingGroup.cs
```diff
@@ -179,14 +179,6 @@ public InputMap<string> Tags
             set => _tags = value;
         }
 
-        [Input("tagsAll")]
```
jkodroff, line 182: Do we need to add this to the changelog?

jkodroff (approved):
Approved, but see comment RE `tags_all`.

sdk/dotnet/Iot/ThingGroup.cs
stack72: I am going to say I don't believe so - it was never actually something you could set as it was overwritten..

Issue #1678, opened 2021-10-27
ECS Cluster "name" argument wrong description


SharpEdgeMarshall:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

The `name` argument has a wrong description: https://github.com/pulumi/pulumi-aws/blob/master/sdk/python/pulumi_aws/ecs/cluster.py#L29 

It was changed by d7d3d36b7324981b15c27397b80ee7159552de01

Expected:  "The name of the cluster (up to 255 letters, numbers, hyphens, and underscores)"
Actual:  "Name of the setting to manage. Valid values: `containerInsights`."

lukehoban:
This appears to be another instance of https://github.com/pulumi/pulumi-terraform-bridge/issues/529.

PR #1730, opened 2021-12-04
fix labelling of ecs cluster names

con5cience:
The same comments were used to describe the names of cluster settings as the names of the ECS clusters themselves. This PR fixes that as requested in https://github.com/pulumi/pulumi-aws/issues/1678 .


thomas11:
Hi @con5cience, thank you for your contribution and apologies for the very late response. Unfortunately, the files your PR touches are auto-generated, so we cannot accept manual changes to them.

Issue #1791, opened 2022-01-27
AWS m6a instances not supported


clamo:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

The Pulumi doesn't support the latest m6a family instances.
m6a.large
m6a.xlarge
m6a.2xlarge
m6a.4xlarge
m6a.8xlarge
m6a.12xlarge
m6a.16xlarge
m6a.24xlarge
m6a.32xlarge
m6a.48xlarge

### Steps to reproduce

1. instanceType: "m6a.large"
2. Confirm that the IDE throws an error

Expected: 
Pulumi should've handled AWS m6a instance family.

Actual: 
Pulumi doesn't handle AWS m6a instance family.

SharpEdgeMarshall:
Also m6i instance types are not supported

clamo:
> Also m6i instance types are not supported

Fix for this particular issue was merged.
I'd write up a new issue about m6i then we can go from there.

PR #1792, opened 2022-01-27
AWS m6a instances support #1791

clamo:
This PR is regarding issue #1791 - AWS m6a instances not supported


jkodroff (approved):
Thanks so much for the contribution!

stack72:
@jkodroff we need to build the schema to pick up these changes or we will have a failed build in main line

jkodroff (changes requested):
@clamo Per Paul's comment above, I need to change my review to request changes. Could you run the following commands and resubmit your PR?

* `make tfgen`
* `make build_sdks`

This will allow the build to pass. Note that these commands may take a while to run - on the order of 10 minutes.

jkodroff (approved):

PR #1838, opened 2022-02-26
Add support for m6i ec2 instances

RafalSumislawski:


stack72:
Hank you so much for this @RafalSumislawski! Works a treat :)

PR #1849, opened 2022-03-06
Improve code quality on sample code

dmelo:
Given only one variable is being used on Output.apply, it can use bucket_arn.apply, instead of Output.all.apply.


guineveresaenger:
Hi @dmelo - thank you so much, that does indeed look confusingly verbose.

Unfortunately, as these files are generated via codegen, they'd be overwritten on the next provider update.

We would welcome a contribution to the `/examples/` folder in this repo with the improved language, or one in [Pulumi Examples](https://github.com/pulumi/examples).

PR #1859, opened 2022-03-16
Upgrade to v4.0.0 of the AWS Terraform Provider

stack72:
Please Note: In the upstream provider, the AWS S3 Bucket resource has changed.
While we appreciate the work that has has been done there, we want to ensure that
our Pulumi users continue to be able to work with the existing S3 bucket resource
without any problems. We will publish a guide on how to migrate from `aws.s3.Bucket`
to `aws.s3.BucketV2` to help with this effort.

A number of other breaking changes have been made **outside of the S3 Bucket resource**.
You can find more about these changes in the [upstream provider](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/version-4-upgrade).


provider/resources.go
```diff
@@ -2435,8 +2407,20 @@ func Provider() tfbridge.ProviderInfo {
 			"aws_shield_protection_health_check_association": {Tok: awsResource(shieldMod, "ProtectionHealthCheckAssociation")},
 
 			// S3
-			"aws_s3_account_public_access_block": {Tok: awsResource(s3Mod, "AccountPublicAccessBlock")},
-			"aws_s3_bucket": {
+			"aws_s3_account_public_access_block":                 {Tok: awsResource(s3Mod, "AccountPublicAccessBlock")},
+			"aws_s3_bucket":                                      {Tok: awsResource(s3Mod, "BucketV2")},
```
jkodroff, line 2411: We should add a comment here explaining the scope of our decision, why, and when we can un-alias (v6.0), plus a link to the design doc.

provider/resources.go
```diff
@@ -4854,7 +4838,7 @@ func Provider() tfbridge.ProviderInfo {
 	prov.SetAutonaming(255, "-")
 
 	// Add a CSharp-specific override for aws_s3_bucket.bucket.
-	prov.Resources["aws_s3_bucket"].Fields["bucket"].CSharpName = "BucketName"
+	prov.Resources["aws_s3_bucket_legacy"].Fields["bucket"].CSharpName = "BucketName"
```
jkodroff, line 4841: Comment above should explain why we are doing this... and now that I think of it, why aren't we using `CSharpName` inline instead?

jkodroff (approved):
Approved with comments.

stack72:
@jkodroff I will add comments to the resources.go file

PR #1867, opened 2022-03-19
Add import comment to spot instance request as is standard in other files

chivandikwa:
Add import comment to spot instance request as is standard in other files.


guineveresaenger:
Hi @chivandikwa - thank you for opening this pull request. ❤️ 

Unfortunately, we cannot accept your change as these files are autogenerated and would be overwritten at the next release.

PR #1873, opened 2022-03-22
fix: 🐛 resolve getPolicyDocumentOutput nil dereference (#1872)

### LINKED ISSUE #1872: GetPolicyDocumentOutput panics instead of returning error
## Hello!

- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

This is a problematic line:
https://github.com/pulumi/pulumi-aws/blob/f73d82bf022793a85cdd5bf7ac55638b64740cb9/sdk/go/aws/iam/getPolicyDocument.go#L141

(expanded below)

```
func GetPolicyDocumentOutput(ctx *pulumi.Context, args GetPolicyDocumentOutputArgs, opts ...pulumi.InvokeOption) GetPolicyDocumentResultOutput {
	return pulumi.ToOutputWithContext(context.Background(), args).
		ApplyT(func(v interface{}) (GetPolicyDocumentResult, error) {
			args := v.(GetPolicyDocumentArgs)
			r, err := GetPolicyDocument(ctx, &args, opts...)
			return *r, err
		}).(GetPolicyDocumentResultOutput)
}
```

if `GetPolicyDocument` returns `nil, <error here>`, then `*r` will result in a panic

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1726ecf]
```

### Steps to reproduce

```go
_ = iam.GetPolicyDocumentOutput(ctx, iam.GetPolicyDocumentOutputArgs{
	Statements: iam.GetPolicyDocumentStatementArray{
		iam.GetPolicyDocumentStatementArgs{
			Actions: pulumi.StringArray{
				pulumi.String("sts:AssumeRoleWithWebIdentity"),
			},
			Conditions: iam.GetPolicyDocumentStatementConditionArray{
				iam.GetPolicyDocumentStatementConditionArgs{
					Test: pulumi.String("StringLike"),
					Values: pulumi.StringArray{
						pulumi.Sprintf("repo:%s/%s:*", "mytestorg", "mytestrepo"),
					},
					Variable: pulumi.String("token.actions.githubusercontent.com:sub"),
				},
			},
			Principals: iam.GetPolicyDocumentStatementPrincipalArray{
				iam.GetPolicyDocumentStatementPrincipalArgs{
					Identifiers: pulumi.StringArray{
						pulumi.String("arn:aws:iam::12345678:oidc-provider/token.actions.githubusercontent.com"),
					},
					Type: pulumi.String("Federated"),
				},
			},
		},
	},
})
```

```
    panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x17240af]
    goroutine 23 [running]:
    github.com/pulumi/pulumi-aws/sdk/v4/go/aws/iam.GetPolicyDocumentOutput.func1({0x18fe240?, 0xc0005a4380})
    	/Users/wmcnamee/.asdf/installs/golang/1.18/packages/pkg/mod/github.com/pulumi/pulumi-aws/sdk/v4@v4.38.1/go/aws/iam/getPolicyDocument.go:356 +0x18f
    reflect.Value.call({0x181b940?, 0xc000580480?, 0x100ea45?}, {0x197129f, 0x4}, {0xc00050e018, 0x1, 0xc0005b8240?})
    	/Users/wmcnamee/.asdf/installs/golang/1.18/go/src/reflect/value.go:556 +0x845
```

Note that the line numbers don't match up with the source code (in this repo), but here's a screenshot from the vendored code:

image

Expected: to get a human readable error
Actual: panic and nil pointer exception

ghostsquad:
resolves #1872

### DIFF
--- a/sdk/go/aws/iam/getPolicyDocument.go
+++ b/sdk/go/aws/iam/getPolicyDocument.go
@@ -138,6 +138,12 @@ func GetPolicyDocumentOutput(ctx *pulumi.Context, args GetPolicyDocumentOutputAr
 		ApplyT(func(v interface{}) (GetPolicyDocumentResult, error) {
 			args := v.(GetPolicyDocumentArgs)
 			r, err := GetPolicyDocument(ctx, &args, opts...)
+			if err != nil {
+				return nil, err
+			}
+			if r == nil {
+				return nil, fmt.Errorf("expected either result or error to be nil, not both")
+			}
 			return *r, err
 		}).(GetPolicyDocumentResultOutput)
 }


ghostsquad:
After figuring out how to attach a debugger, I can further confirm that this is the issue.

image

stack72:
Hi @ghostsquad 

What version of Pulumi are you using? I believe we fixed this in Pulumi https://github.com/pulumi/pulumi/pull/9067 in 3.25.1

Paul

guineveresaenger:
hi @ghostsquad! Thank you for your pull request.

The SDKs in this repository are created via codegen, and so this change would be overwritten on the next patch release.

I'll close this PR for now but it looks like @stack72 is pointing you to a possible resolution?

ghostsquad:
I'm not sure I understand @stack72. I'm using the latest release of this repository, which is 4.38.1

Don't know what this has to do with the Pulumi CLI. The error doesn't occur in the CLI.

stack72:
Hi @ghostsquad 

The engine is what controls this behaviour and the engine is a dependency of the CLI. We fixed the engine in the PR I linked you to an it was released as 3.25.1

The error may manifest itself in this code base but it's actually being returned from the engine 

Paul

ghostsquad:
For reference:
```shell
at 17:19:33 ❯ pulumi version
v3.26.1
```

But I still don't understand what voodoo you are referring to. The code that is running can be traced by go, the debugger, my IDE to this package, which is imported in my `main.go`:

```go
import "github.com/pulumi/pulumi-aws/sdk/v4/go/aws/iam"
```

Additionally, my binary is built _without_ help from the pulumi CLI via:

```yaml
runtime:
  name: go
  options:
    binary: app.sh
```

app.sh
```shell
if [ "${DEBUG:-}" == "1" ]; then
  go build -o debugapp -gcflags "all=-N -l" .
  dlv --listen=:2345 --headless=true --api-version=2 --accept-multiclient exec ./debugapp
else
  go build -o app .
  ./app
fi
```

go.mod
```
github.com/pulumi/pulumi-aws/sdk/v4 v4.38.1
```

Thank you for taking the time to explain to me what's going on.

ghostsquad:
If you are referring to the specific error that I found in the debugger, that's not what I'm concerned about. That was by design via this in the config

```yaml
pulumi:disable-default-providers:
- '*'
```

The problem is that this error was never surfaced. I got a nil pointer exception, and without debugging, it was not possible to retrieve the error message.

ghostsquad:
I updated https://github.com/pulumi/pulumi-aws/pull/1873#issuecomment-1075481409 to point out that `r = nil` here. Which is why `*r` panics.

Issue #1906, opened 2022-04-13
aws.Provider ignores $AWS_PROFILE on `pulumi refresh`


jre21:
### What happened?

When using a named AWS provider (i.e., calling new aws.Provider() within source code and then passing the result into other pulumi resources), the provider resource stores the current AWS profile into stack state even if it wasn't specified as part of the provider configuration.  Subsequent calls to `pulumi refresh` attempt to fetch credentials from the profile in stack state rather than the `AWS_PROFILE` environment variable, which can cause authentication errors if colleagues use different names for their AWS profiles.

### Steps to reproduce

1. Create a pulumi stack with a named AWS provider, such as
```
const provider = new aws.Provider("provider", {region: "us-east-1"});

const bucket = new aws.s3.Bucket("my-bucket", {}, {provider: provider});
```

2. Run `AWS_PROFILE=test1 pulumi up`

3. Run `AWS_PROFILE=test2 pulumi refresh`

### Expected Behavior

1. Because the AWS profile was not specified as part of the provider configuration, the profile should be stored in stack state.
2. The refresh should use the `test2` AWS profile when querying the live state of the AWS bucket.

### Actual Behavior

1. Pulumi's diff output indicates `profile: test1` as part of the recorded stack state for the `provider` resource.
2. The refresh is performed with the `test1` profile, ignoring the `AWS_PROFILE` environment variable.

### Versions used

Previously-filed issues (#1470, #1585) indicate that the undesired behavior was present in pulumi-aws 4.14.0 and earlier.  From my own tests, version 4.20.0 produced the expected behavior, while versions 4.21.1 and later (my latest tested version is 4.36.0) have reverted to the undesired behavior (the profile is stored in state and environment variables are ignored on refresh).

### Additional context

A colleague of mine provided some additional context on this in https://github.com/pulumi/pulumi-aws/issues/1585#issuecomment-1093541296.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

jre21:
The general principle I'm advocating for here is that any AWS provider configuration which was read from the environment (rather than the provider resource's arguments) at stack creation should continue to be read from the environment (rather than stack state) in all further interactions.

lukehoban:
[...]

PR #1909, opened 2022-04-15
Remove storing AWS_PROFILE in state

stack72:
Fixes: #1906

When this is stored in state, a second run of Pulumi with a different
PROFILE name will fail

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -311,11 +311,6 @@ func Provider() tfbridge.ProviderInfo {
 					EnvVars: []string{"AWS_REGION", "AWS_DEFAULT_REGION"},
 				},
 			},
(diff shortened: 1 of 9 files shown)


lukehoban (approved):
LGTM.

This is technically a breaking change, but narrow enough that I think we should take it, given the fairly serious problems this can cause in it's current state.

We should though add a CHANGELOG entry, note that this can be breaking change for users who are using `aws.config.profile` today, and that if users *are* relying on that returning values of profile from the environment, they can update their code to use `aws.config.profile || process.env.AWS_PROFILE` (or equivalent for other languags).

alexbowers:
Yeah, this broke our code

Issue #1958, opened 2022-05-21
RDS ClusterInstance example does not work


omercnet:
### What happened?

Tried implementing the simple example from File: [themes/default/content/registry/packages/aws/api-docs/rds/clusterinstance/_index.md](https://www.pulumi.com/registry/packages/aws/api-docs/rds/clusterinstance/)

### Steps to reproduce

Just run the example 

### Expected Behavior

it should work

### Actual Behavior

it doesn't work :)

```
Type 'Output<string | undefined>' is not assignable to type 'Input<EngineType> | undefined'.
  Type 'OutputInstance<string | undefined>' is not assignable to type 'Input<EngineType> | undefined'.
    Type 'OutputInstance<string | undefined>' is not assignable to type 'OutputInstance<EngineType>'.
      Type 'string | undefined' is not assignable to type 'EngineType'.
        Type 'undefined' is not assignable to type 'EngineType'.ts(2322)
clusterInstance.d.ts(416, 5): The expected type comes from property 'engine' which is declared here on type 'ClusterInstanceArgs'
```

### Versions used

NAME                              VERSION
@pulumi/datadog                   4.8.0
@tsconfig/node16                  1.0.2
@pulumi/tls                       4.4.0
@typescript-eslint/parser         5.25.0
typescript                        4.6.4
@pulumi/eks                       0.40.0
@pulumi/pulumi                    3.32.1
@typescript-eslint/eslint-plugin  5.25.0
@pulumi/awsx                      0.40.0
@pulumi/gcp                       6.23.0
@pulumi/kubernetes                3.19.1
@pulumi/kubernetesx               0.1.6
@pulumi/random                    4.6.0
@types/node                       17.0.33

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

guineveresaenger:
Thank you for filing this bug! It looks like there's an underlying type mismatch. The clusterInstance is expecting an `EngineType`, which the Cluster is outputting as a `string`.

We'll take a look as soon as we can.

PR #1974, opened 2022-05-27
Implement AltTypes for RDS cluster.engine field

guineveresaenger:
Fixes https://github.com/pulumi/pulumi-aws/issues/1958.

TL;DR: the `cluster.engine` field on `rds.Cluster` did not accept a string as an input, as per community bug. This PR allows for both enum and string types.


guineveresaenger:
Hm, I am running into some Typescript oddness; locally testing this it's a slightly different type mismatch error:

```
new aws.rds.ClusterInstance(`clusterName`, {
        identifier: `aurora-cluster-demo-${range.value}`,
        clusterIdentifier: _default.id,
        instanceClass: "db.r4.large",
        engine: _default.engine.apply,
        engineVersion: _default.engineVersion,

    })
```

`engine: _default.engine.apply` is still broken

```
TS2322: Type 'Output<string | undefined>' is not assignable to type 'Input<string> | undefined'.   
Type 'OutputInstance<string | undefined>' is not assignable to type 'Input<string> | undefined'.     
Type 'OutputInstance<string | undefined>' is not assignable to type 'OutputInstance<string>'.       
Type 'string | undefined' is not assignable to type 'string'.         
Type 'undefined' is not assignable to type 'string'.
```

but forcing the undefined type to be a string will work.

engine: _default.engine.apply(eng => eng!),

Would appreciate some advice!

PR #1976, opened 2022-05-30
[internal] Upgrading to Go 1.18.x

stack72:


AaronFriel:
@stack72 pushed a couple improvements. At first I tried switching to `//go:embed` directives, but the OOM happened well after. I ran `make tfgen` on my laptop and I saw how much memory `go build` was using due to 8 compilation processes, so I tweaked that next. The build parallelism is likely 2 (GOMAXPROCS, 2 core runner) on CI, this pushes it down to 1.

PR #1999, opened 2022-06-08
Fix OOM during build by using embed directive to halve memory usage.

AaronFriel:
Memory usage of the `go build` step of `make provider` without this change on go1.18 is between 7500 and 8500MiB across three runs.

With this branch, memory usage is reduced by >60% to around 3300MiB.


provider/cmd/pulumi-resource-aws/generate.go
```diff
@@ -12,13 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build ignore
 // +build ignore
```
guineveresaenger, line 16: do we want to remove the old syntax here?

AaronFriel: Looks like the new syntax is supported in 1.16 and later, I'll fix it up and remove +build

provider/cmd/pulumi-resource-aws/main.go
```diff
@@ -17,11 +17,16 @@
 package main
 
 import (
+	_ "embed"
```
jkodroff, line 20: Curious why this is necessary. I'm sure it is - I just don't understand this line since I thought (perhaps incorrectly) that other `go` commands do not require an import.

jkodroff:
@AaronFriel FYI, we need to ensure that the "dirty schema check" (a GH Action step that runs `make tfgen build_sdks` and then does a `git status` and fails if any changes are present in the tree) is not affected by this change. (`make provider` calls the `make tfgen` target.)

stack72:
I am combining this PR with the GHA upgrade so we need to abandon this

AaronFriel:
@jkodroff it should not cause a dirty tree, this PR creates a separate git-ignored file "schema-embed.json".

stack72:
This has been merged into master on the other PR :) and the build is green!

Issue #1995, opened 2022-06-06
pulumi-aws 5.7.0 seems unable to validate AWS credentials for AWS Profile only


nnewton:
### What happened?

Updating to pulumi-aws 5.7.0 caused stack to be unable to find the access key/secret key for aws. These keys are working via AWS cli and work on pulumi-aws 5.6.0 (downgrading fixes the issue). 

### Steps to reproduce

$ export AWS_PROFILE=PROFILENAME
$ pulumi preview --diff
Results in: 
error: unable to validate AWS AccessKeyID and/or SecretAccessKey - see https://pulumi.io/install/aws.html for details on configuration

### Expected Behavior

Validating aws credentials based on AWS_PROFILE

### Actual Behavior

It does seem to find the credentials, when run with --debug I see this:
```Diagnostics:
  pulumi:pulumi:Stack (aws-test):
    debug: AWS Auth provider used: "SharedCredentialsProvider"
    debug: Retrieved credentials from "SharedConfigCredentials: /Users/nnewton/.aws/credentials"
    debug: Trying to get account information via iam:GetUser
    debug: [aws-sdk-go-v2] Request
    debug: POST / HTTP/1.1
    debug: Host: iam.amazonaws.com
    debug: User-Agent: APN/1.0 Pulumi/1.0 Pulumi/1.0 Pulumi-Aws/0.11+compatible (+https://www.pulumi.com) aws-sdk-go-v2/1.16.2 os/macos lang/go/1.17.9 md/GOOS/darwin md/GOARCH/arm64 api/iam/1.18.0
    debug: Content-Length: 33
    debug: Amz-Sdk-Invocation-Id: ec89da52-f37b-4985-9dd8-e6d533763942
    debug: Amz-Sdk-Request: attempt=1; max=25
    debug: Content-Type: application/x-www-form-urlencoded
    debug: X-Amz-Date: 20220606T224320Z
    debug: Accept-Encoding: gzip
    debug:
    debug: Action=GetUser&Version=2010-05-08
    debug: [aws-sdk-go-v2] Response
    debug: HTTP/1.1 200
    debug: Content-Length: 527
    debug: Content-Type: text/xml
    debug: Date: Mon, 06 Jun 2022 22:43:20 GMT
    debug: X-Amzn-Requestid: 906e4294-f0bc-40a9-9aa7-e66dc4dc6fdc
    debug:
[...]

PR #2004, opened 2022-06-10
Ensure we validate credentials correctly

stack72:
Fixes: #1995

**Currently the creds validation is broken:**
[asciicast](https://asciinema.org/a/AgIZGKgVnFEejf2wKIegTECrm)

New code:

**With AWS_PROFILE:**
[asciicast](https://asciinema.org/a/qSdDAZCtgnDWiCBH2MBG5cQ8I)

**With default creds:**
[asciicast](https://asciinema.org/a/A4v7yF9VHq1Mr0hayfkabLNBm)

**With Environment Vars for AWS_ACCESS_KEY_ID && AWS_SECRET_ACCESS_KEY**
[asciicast](https://asciinema.org/a/qB50n4onFLYQAwARCGhwVQVK7)

**Using an AWS AssumeRole:**
[asciicast](https://asciinema.org/a/fvfbhZDzwYvWEM5NI72iIQzVC)

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -15,6 +15,7 @@
 package provider
 
 import (
+	"context"
 	"fmt"
 	"math/rand"
 	"os"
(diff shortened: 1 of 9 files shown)


mdop-wh:
How do I run pulumi with the aws provider from an ec2 instance using an instance role? This no longer works now

christophermaier:
We're hitting the exact same thing as @mdop-wh is (literally word-for-word 😅)

Reading the code in this PR, I'm surprised to see that the default value of `aws:skipCredentialsValidation` before was `true`... I'm not clear why having this value set to `false` is failing in AWS, particularly since we use the STS service to get credentials to use in the first place. Is there some specific IAM permission (or some other configuration) that is needed for validation to occur that I'm missing?

PR #2028, opened 2022-06-30
Configure Dependabot

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>


stack72:
Hi @gliptak 

Thanks so much for the work here - we actually centrally manage our GHA for all of our providers so I am going to close this out

Paul

PR #2040, opened 2022-07-06
[internal] Adding support for Java SDK to the pulumi-aws repo

stack72:


guineveresaenger (approved):
Do we need to add Java examples here?

Issue #1968, opened 2022-05-26
prompt user for `aws login`


dixler:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details
Idk if this is reasonable to implement, but it would be nice to improve this flow
```
kdixler@kdixler-20xw004aus ~/Documents/aws-python % pulumi up
Previewing update (dev)

View Live: https://app.pulumi.com/dixler/aws-python/dev/previews/a96bbfbb-02d0-4d09-bf91-9c83d8ed1c48

     Type                 Name            Plan       Info
 +   pulumi:pulumi:Stack  aws-python-dev  create     
     └─ aws:s3:Bucket     my-bucket                  1 error
 
Diagnostics:
  aws:s3:Bucket (my-bucket):
    error: unable to discover AWS AccessKeyID and/or SecretAccessKey - see https://pulumi.io/install/aws.html for details on configuration
 

255 kdixler@kdixler-20xw004aus ~/Documents/aws-python % aws sso login
Attempting to automatically open the SSO authorization page in your default browser.
If the browser does not open or you wish to use a different device to authorize this request, open the following URL:

https://device.sso.us-west-2.amazonaws.com/

Then enter the code:

<redacted>
Successully logged into Start URL: <redacted>
kdixler@kdixler-20xw004aus ~/Documents/aws-python % pulumi up
Previewing update (dev)

View Live: <redacted>

     Type                 Name            Plan       
 +   pulumi:pulumi:Stack  aws-python-dev  create     
 +   └─ aws:s3:Bucket     my-bucket       create     
 
Resources:
    + 2 to create

Do you want to perform this update?  [Use arrows to move, enter to select, type to filter]
  yes
> no
  details
...

```

stack72:
@dixler even by popping this open, it won't allow us to pick up the credentials. The credentials for the provider need to either be a profile, shared credentials file, env vars or pulumi config variants

I am going to close this out as a by-design as we are restricted to passing the configuration to the underlying provider in this way

dixler:
Apparently this error message is more common than forgetting to set up a required config value(which I do all the time).

dixler:
> Which "required config value" ?

like whenever a user has a

```typescript
const foo = config.require("my-key")
```

inside of their program and they make a new stack, but their new stack doesn't have that value configured.

stack72:
[...]

PR #2051, opened 2022-07-13
improve error message to contain command to configure AWS credentials

dixler:
part of #1968 

redirecting input from the provider to the CLI and back over GRP seems like a lot of work at the moment.

I was trying to optimize for power users double-clicking to pick up the entire command to copy and paste more effortlessly.

the output of the current PR looks as such
```
kdixler@kdixler-20xw004aus ~/Documents/aws-python % pulumi up
Previewing update (dev)

View Live: https://app.pulumi.com/dixler/aws-python/dev/previews/6ddacf00-ff5b-4d70-ac17-9f7191c173f6

     Type                 Name            Plan       Info
 +   pulumi:pulumi:Stack  aws-python-dev  create     
     └─ aws:s3:Bucket     my-bucket                  1 error
 
Diagnostics:
  aws:s3:Bucket (my-bucket):
    error: unable to validate AWS credentials - see https://pulumi.io/install/aws.html for details on configuration
    log in to AWS using the AWS CLI:
        aws configure
    log in to AWS using the AWS CLI(with SSO):
        aws sso login
 

255 kdixler@kdixler-20xw004aus ~/Documents/aws-python %         aws configure

AWS Access Key ID [None]: ^C
130 kdixler@kdixler-20xw004aus ~/Documents/aws-python %         aws sso login

Attempting to automatically open the SSO authorization page in your default browser.
If the browser does not open or you wish to use a different device to authorize this request, open the following URL:

https://device.sso.us-west-2.amazonaws.com/

Then enter the code:
```

would definitely appreciate feedback on the language of this command.

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -345,10 +345,11 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
 	config.SharedConfigFiles = []string{configPath}
 
 	if _, err := awsbase.GetAwsConfig(context.Background(), config); err != nil {
-		return fmt.Errorf("unable to validate AWS credentials - see https://pulumi.io/install/aws.html for details on configuration\n" +
-			"if using the AWS CLI:\n" +
+		return fmt.Errorf("unable to validate AWS credentials " +
+			"- see https://pulumi.io/install/aws.html for details on configuration\n" +
+			"log in to AWS using the AWS CLI:\n" +
 			"    aws configure\n" +
-			"if using AWS CLI(with SSO):\n" +
+			"log in to AWS using the AWS CLI(with SSO):\n" +
 			"    aws sso login\n")
 	}
 
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -345,8 +345,11 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
 	config.SharedConfigFiles = []string{configPath}
 
 	if _, err := awsbase.GetAwsConfig(context.Background(), config); err != nil {
-		return fmt.Errorf("unable to validate AWS credentials " +
-			"- see https://pulumi.io/install/aws.html for details on configuration")
+		return fmt.Errorf("unable to validate AWS credentials - see https://pulumi.io/install/aws.html for details on configuration\n" +
+			"if using the AWS CLI:\n" +
+			"    aws configure\n" +
+			"if using AWS CLI(with SSO):\n" +
+			"    aws sso login\n")
 	}
 
 	return nil
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -347,10 +347,11 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
 	if _, err := awsbase.GetAwsConfig(context.Background(), config); err != nil {
 
 		return fmt.Errorf("unable to validate AWS credentials. Make sure you have: \n\n" +
-			" \t • Set the AWS_SECRET_ACCESS_KEY \n" +
-			" \t • Set the AWS_ACCESS_KEY_ID \n" +
-			" \t • Set your AWS region using `pulumi config set aws:region <region>`, e.g `us-west-2` \n\n" +
-			"See https://pulumi.io/install/aws.html for details on configuration")
+			" \t • Configured your AWS access key \n" +
+			" \t • Configured your AWS_ access key ID \n" +
+			" \t • Set your AWS region using `pulumi config set aws:region <region>`, e.g `us-west-2` \n" +
+			" \t You can set these via `aws configure`. \n\n" +
+			"See https://pulumi.io/install/aws.html for further details on configuration")
 	}
 
 	return nil
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -347,11 +347,9 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
 	if _, err := awsbase.GetAwsConfig(context.Background(), config); err != nil {
 
 		return fmt.Errorf("unable to validate AWS credentials. Make sure you have: \n\n" +
-			" \t • Configured your AWS access key \n" +
-			" \t • Configured your AWS_ access key ID \n" +
-			" \t • Set your AWS region using `pulumi config set aws:region <region>`, e.g `us-west-2` \n" +
-			" \t You can set these via `aws configure`. \n\n" +
-			"See https://pulumi.io/install/aws.html for further details on configuration")
+			" \t • Set your AWS region, e.g. `pulumi config set aws:region us-west-2` \n" +
+			" \t • Configured your AWS credentials as per https://pulumi.io/install/aws.html \n" +
+			" \t You can also set these via cli using `aws configure`. \n\n")
 	}
 
 	return nil
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -345,12 +345,12 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
 	config.SharedConfigFiles = []string{configPath}
 
 	if _, err := awsbase.GetAwsConfig(context.Background(), config); err != nil {
-		return fmt.Errorf("unable to validate AWS credentials " +
-			"- see https://pulumi.io/install/aws.html for details on configuration\n" +
-			"log in to AWS using the AWS CLI:\n" +
-			"    aws configure\n" +
-			"log in to AWS using the AWS CLI(with SSO):\n" +
-			"    aws sso login\n")
+
+		return fmt.Errorf("unable to validate AWS credentials. Make sure you have: \n\n" +
+			" \t • Set the AWS_SECRET_ACCESS_KEY \n" +
+			" \t • Set the AWS_ACCESS_KEY_ID \n" +
+			" \t • Set your AWS region using `pulumi config set aws:region <region>`, e.g `us-west-2` \n\n" +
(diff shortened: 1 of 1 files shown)


dixler:
> You can verify you have valid AWS credentials by using
> aws sts get-caller-identity

@jaxxstorm I was running through this and I got confused myself with the return value of `aws sts get-caller-identity` since it still returns a value despite not having valid creds. I think an advanced user (such as yourself) might think of running that on their own.

```
kdixler@kdixler-20xw004aus ~/Documents/aws-python % pulumi up
Previewing update (dev)

View Live: https://app.pulumi.com/dixler/aws-python/dev/previews/d7a6ee17-6da8-4524-948b-5e264fdc6851

     Type                 Name            Plan       Info
 +   pulumi:pulumi:Stack  aws-python-dev  create     
     └─ aws:s3:Bucket     my-bucket                  1 error
 
Diagnostics:
  aws:s3:Bucket (my-bucket):
    error: unable to validate AWS credentials - see https://pulumi.io/install/aws.html for details on configuration
    log in to AWS using the AWS CLI:
        aws configure
        aws sso login
    check your AWS IAM Role:
        aws sts get-caller-identity
 

255 kdixler@kdixler-20xw004aus ~/Documents/aws-python %         aws configure

AWS Access Key ID [None]: ^C
130 kdixler@kdixler-20xw004aus ~/Documents/aws-python %         aws sso login

Attempting to automatically open the SSO authorization page in your default browser.
If the browser does not open or you wish to use a different device to authorize this request, open the following URL:

https://device.sso.us-west-2.amazonaws.com/

Then enter the code:

^C
130 kdixler@kdixler-20xw004aus ~/Documents/aws-python %         aws sts get-caller-identity

{
    "UserId": "<REDACTED>:kyle@pulumi.com",
    "Account": "<REDACTED>",
    "Arn": "arn:aws:sts::<REDACTED>:assumed-role/AWSReservedSSO_AdministratorAccess_<SALT>/kyle@pulumi.com"
}
kdixler@kdixler-20xw004aus ~/Documents/aws-python % pulumi up
Previewing update (dev)

View Live: https://app.pulumi.com/dixler/aws-python/dev/previews/a581bd02-d0cf-404d-8881-766cb9f03ab6

     Type                 Name            Plan       Info
 +   pulumi:pulumi:Stack  aws-python-dev  create     
     └─ aws:s3:Bucket     my-bucket                  1 error
 
Diagnostics:
  aws:s3:Bucket (my-bucket):
    error: unable to validate AWS credentials - see https://pulumi.io/install/aws.html for details on configuration
    log in to AWS using the AWS CLI:
        aws configure
        aws sso login
    check your AWS IAM Role:
        aws sts get-caller-identity
 

```

stack72 (commented):
I don't think this is strictly true - for example I didn't have to do that today - I just set the env vars from my 1Password. What about people who use profiles or iam roles not static credentials

Why is this any more clear than linking to the installation configuration page?

jaxxstorm:
Another question: why don't we return the error here? We only return the message, but not the actual error itself.

Why can't we return the error, parse it and then tell the user how to fix the problem based on the returned error?

dixler:
> I don't think this is strictly true - for example I didn't have to do that today - I just set the env vars from my 1Password. What about people who use profiles or iam roles not static credentials
> 
> Why is this any more clear than linking to the installation configuration page?

This PR is about adding extra clarity in the error message on a way to recover in a very likely situation that could be intimidating for a **new potential pulumi user**. The population of users you describe that are comfortable using complex ways of interacting with their cloud environments should be uninhibited by this addition. If you believe that they would be confused by this addition, could you please provide phrasing that might make it less confusing for them while still supporting new users.

### Context

Users from the usability study tend to skip prose and just tend to look at only anything inside of a code fence. The documentation recommends setting up AWS credentials in environment variables inside of a code fence with prose mentioning that it respects AWS configuration. 

I want to clarify this in the docs, but we're currently redesigning them and that may take some time and this is where anyone with this issue will run into no matter how they run into it.

This issue may still continue because CLI users tend to reuse the same commands/patterns over time. Because of this, there may be some users in the wild that would benefit from being made aware that the AWS CLI setup commands are compatible. Maybe they'll discover it on their own, but attention is a really scarce resource and this is a very minor change. We can't guarantee that they click that link(for context, we lose 25% of users before they even click past the landing page for the cloud examples) so putting help the fewest clicks away is more desirable.

provider/resources.go
```diff
@@ -346,7 +346,11 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
 
 	if _, err := awsbase.GetAwsConfig(context.Background(), config); err != nil {
 		return fmt.Errorf("unable to validate AWS credentials " +
-			"- see https://pulumi.io/install/aws.html for details on configuration")
+			"- see https://pulumi.io/install/aws.html for details on configuration\n" +
+			"log in to AWS using the AWS CLI:\n" +
```
guineveresaenger, line 350: Can we capitalize "Log in..."?

provider/resources.go
```diff
@@ -346,7 +346,11 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
...
+			"    aws configure\n" +
+			"log in to AWS using the AWS CLI(with SSO):\n" +
```
guineveresaenger, line 352: capitalize, same as above

provider/resources.go
```diff
@@ -346,7 +346,11 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
 
 	if _, err := awsbase.GetAwsConfig(context.Background(), config); err != nil {
 		return fmt.Errorf("unable to validate AWS credentials " +
```
guineveresaenger, line 348: Perhaps I'm overthinking this but "validate" almost makes it sound like we're interfacing with a cert authority or something. Any objections to "Invalid credentials"?

In other news, I googled this error message, and it appears that both AWS and Azure use this wording, and I kinda hate it but I may be alone in my opinion.

guineveresaenger (commented):
I really support this PR, @stack72, for all the reasons @dixler brought up.

LGTM but holding explicit approval awaiting response from Paul. <3

provider/resources.go
dixler: maybe I could add a `$` to make it more clear that it is a shell command

provider/resources.go
```diff
@@ -345,8 +345,12 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
 	config.SharedConfigFiles = []string{configPath}
 
 	if _, err := awsbase.GetAwsConfig(context.Background(), config); err != nil {
-		return fmt.Errorf("unable to validate AWS credentials " +
-			"- see https://pulumi.io/install/aws.html for details on configuration")
+
+		return fmt.Errorf("unable to validate AWS credentials. Make sure you have: \n\n" +
+			" \t • Set the AWS_SECRET_ACCESS_KEY \n" +
+			" \t • Set the AWS_ACCESS_KEY_ID \n" +
+			" \t • Set your AWS region using `pulumi config set aws:region <region>`, e.g `us-west-2` \n\n" +
```
dixler, line 352: @guineveresaenger I'd like to make sure to convey that we integrate with the configuration that a user might already have in their .aws directory. I think suggesting `aws configure` would help imply that we support `aws sso login`

guineveresaenger: Fair point - I made the checklist more generic and suggested `aws configure`. WDYT?

provider/resources.go
```diff
@@ -345,8 +345,13 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
...
 
 	if _, err := awsbase.GetAwsConfig(context.Background(), config); err != nil {
-		return fmt.Errorf("unable to validate AWS credentials " +
-			"- see https://pulumi.io/install/aws.html for details on configuration")
+
+		return fmt.Errorf("unable to validate AWS credentials. Make sure you have: \n\n" +
+			" \t • Configured your AWS access key \n" +
+			" \t • Configured your AWS_ access key ID \n" +
+			" \t • Set your AWS region using `pulumi config set aws:region <region>`, e.g `us-west-2` \n" +
+			" \t You can set these via `aws configure`. \n\n" +
```
dixler, line 353: Maybe "You can also set these..."

provider/resources.go
```diff
@@ -345,8 +345,13 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
 	config.SharedConfigFiles = []string{configPath}
 
 	if _, err := awsbase.GetAwsConfig(context.Background(), config); err != nil {
-		return fmt.Errorf("unable to validate AWS credentials " +
-			"- see https://pulumi.io/install/aws.html for details on configuration")
+
+		return fmt.Errorf("unable to validate AWS credentials. Make sure you have: \n\n" +
+			" \t • Configured your AWS access key \n" +
+			" \t • Configured your AWS_ access key ID \n" +
```
dixler, line 351: I think the environment variables in the previous commit were more clear.

dixler (commented):
I know that `aws configure` doesn't set environment variables, but it configures the environment and I think anyone using AWS would understand that we're implying that if the environment is configured to use an AWS SDK, pulumi is configured too.

Obviously need other opinions on this.

provider/resources.go
stack72: Folks, the user may have already used `aws configure` and are unsure of what we expect of the user. I don't think `aws configure` is correct herre

provider/resources.go
guineveresaenger: It's not how everyone sets up their creds, though, and in the docs, we do not recommend this as the best option.

stack72 (approved):

Issue #2081, opened 2022-08-10
API Gateway integrations incorrectly updated when target changes (AWS Classic)


ewucc:
### What happened?

I have a service deployed using pulumi as a lambda fronted with an API Gateway. I want to convert the lambda to use ECS and fargate instead. I update the pulumi files to change the lambda to now be an ECS Task/Service, spin up an ECS cluster, and create a restAPI integration to use HTTP_PROXY, VPC_LINK, and NLB url instead of the initial AWS_PROXY and event handler url.

I expect the Lambda along with any Lambda related resources like permissions and cloudwatch log groups to be deleted, while the ECS related services like task, service, vpc link, NLB, and any security groups, permissions to be created. I also expect the restAPI aws.apigateway.Integration to be updated to instead of using the lambda, use the VPC_LINK and nlb. 

Instead what happens is all of the expected resources are created, but the aws.apigateway.integrations are instead destroyed without recreating new integrations using the vpc link. At this point, I can navigate to the AWS console and manually update the method to create the integration. Doing so makes everything function correctly. 

### Steps to reproduce

Start off with a restAPI configured with resource and method. Create an integration for lambda and deploy your service

```
new aws.apigateway.Integration(`${methodPath}-${r.verb.toUpperCase()}`, {
                    restApi: restAPI.id,
                    resourceId: parentResource == null ? restAPI.rootResourceId : parentResource.id,
                    httpMethod: method.httpMethod,
                    integrationHttpMethod: "POST",
                    type: "AWS_PROXY",
                    uri: lambda.invokeArn
                })
```

After that is completed, set up pulumi code to create an ECS cluster and it's associated resources. Create a new integration 
for VPC link instead

```
new aws.apigateway.Integration(`${methodPath}-${r.verb.toUpperCase()}`, {
                    restApi: restAPI.id,
                    resourceId: parentResource == null ? restAPI.rootResourceId : parentResource.id,
                    httpMethod: method.httpMethod,
                    integrationHttpMethod: method.httpMethod,
                    type: "HTTP_PROXY",
                    connectionType: "VPC_LINK",
                    connectionId: vpcLink.id,
                    uri: pulumi.interpolate`http://${nlb.loadBalancer.dnsName}${r.path}`
                })
```
 
 Deploying this will result in all the resources being created but the integration being deleted. Pulumi up will complete and show that it was updated
[...]

PR #2089, opened 2022-08-12
apigatewayIntegration: Ensure deleteBeforeReplace is defaults to true

guineveresaenger:
Fixes https://github.com/pulumi/pulumi-aws/issues/2081.


guineveresaenger:
Note I have not verified this on an example locally yet due to slow builds but I've verified this behavior works on a previous resource.

stack72 (approved):
Thank you!!

PR #2116, opened 2022-08-29
adding i4i instances

kuperiu:


guineveresaenger:
Hi @kuperiu - unfortunately we cannot accept this PR, as the file you're changing is autogenerated and your changes would revert on the next update. Support for these instance types is dependent on upstream dependencies at this time.

Issue #2108, opened 2022-08-23
display all misconfiguration in the same error message


dixler:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details
As I was complaining, I thought of a solution.

Can we take all of the checks for provider configuration, and display them at the same time instead of failing iteratively.
### Desired State
if none are set:
```
    error: unable to configure AWS provider. Make sure you have all of the following configured:
    
     	 [ ] Set your AWS region, e.g. `pulumi config set aws:region us-west-2`
     	 [ ] Configured your AWS credentials(https://pulumi.io/install/aws.html)
```
if one is set, but not the other:
```
    error: unable to configure AWS provider. Make sure you have all of the following configured:
    
     	 [x] Set your AWS region, e.g. `pulumi config set aws:region us-west-2`
     	 [ ] Configured your AWS credentials(https://pulumi.io/install/aws.html)
```

### Current State:
```
    error: unable to find a AWS Region config. Set the Region by using:
    
     	 • `pulumi config set aws:region us-west-2` or
     	 • An AWS Profile or
     	 • Environment variables:`AWS_REGION` or `AWS_DEFAULT`
```
after resolving this, my creds were not configured and I received this error message
```
    error: unable to validate AWS credentials. Make sure you have:
    
     	 • Set your AWS region, e.g. `pulumi config set aws:region us-west-2`
     	 • Configured your AWS credentials as per https://pulumi.io/install/aws.html
     	 You can also set these via cli using `aws configure`.
```

### Motivation
I was using pulumi-aws and the error message confused me as there is an error message/check to see that the region is set.
```
    error: unable to find a AWS Region config. Set the Region by using:
    
     	 • `pulumi config set aws:region us-west-2` or
     	 • An AWS Profile or
     	 • Environment variables:`AWS_REGION` or `AWS_DEFAULT`
```
after resolving this, my creds were not configured and I received this error message
```
    error: unable to validate AWS credentials. Make sure you have:
    
     	 • Set your AWS region, e.g. `pulumi config set aws:region us-west-2`
     	 • Configured your AWS credentials as per https://pulumi.io/install/aws.html
     	 You can also set these via cli using `aws configure`.
```

can we remove the line suggesting
```
     	 • Set your AWS region, e.g. `pulumi config set aws:region us-west-2`
```
[...]

Issue #2105, opened 2022-08-23
Regression when migrating to SharedCredentialsFiles field


yarinm:
### What happened?

When we are initializing our Provider we've been using `SharedCredentialsFile`. We wanted to upgrade from 4.9 -> 5.11.0 and we've noticed that the field is deprecated and we should move to use `SharedCredentialsFiles` when we tried to use this field it seems that it's being completely ignored and not used at all.

Errors we're seeing are:

```
  aws:sqs:Queue (diskanalyzer-scans-dlq):
    error: unable to validate AWS credentials. Make sure you have:

         • Set your AWS region, e.g. `pulumi config set aws:region us-west-2`
         • Configured your AWS credentials as per https://pulumi.io/install/aws.html
         You can also set these via cli using `aws configure`.
```

When we use the old field things seem to work fine.

### Steps to reproduce

Use `SharedCredentialsFile`

### Expected Behavior

AWS provider should use appropriate provider file

### Actual Behavior

Seems that it's not being used at all

### Output of `pulumi about`

```
❯ pulumi about
CLI
Version      3.26.1
Go Version   go1.17.8
Go Compiler  gc

Host
OS       darwin
Version  12.5.1
Arch     arm64

Backend
Name     Yarins-MacBook-Pro-4.local
URL      file:///Users/yarin/.pulumi-state
User     yarin
```

### Additional context

Using automation api to invoke pulumi

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

stack72:
Hi @yarinm 

My apologies for this - you can skip this credentials validation check by using `pulumi config set aws:skipCredentialsValidation true`

I will ensure we get this working as expected now

Paul

yarinm:
Hey @stack72 even when we set `skipCredentialsValidation` it fails, it doesn't seem to use the new field for credential files

Issue #2104, opened 2022-08-19
Authentication check does not respect AWS Profile settings for `region`


guineveresaenger:
### What happened?

I am trying to provision an AWS resource musing only an AWS profile:

Code:
```
import * as pulumi from "@pulumi/pulumi";
import * as aws from "@pulumi/aws";

const bucket = new aws.s3.Bucket("b", {
    acl: "private",
    tags: {
        Environment: "Dev",
        Name: "My bucket",
    },
});
```

Auth config and console settings/output:

```
@guin:aws-bucket:owl: aws configure
AWS Access Key ID [****************]: 
AWS Secret Access Key [****************: 
Default region name [us-west-2]: 
Default output format [None]: 
@guin:aws-bucket:owl: pulumi up
Previewing update (dev)

View Live: https://app.pulumi.com/guinevere/aws-bucket/dev/previews/2b97f4d4-1113-49de-b460-7377da102024

     Type                 Name            Plan       Info
 +   pulumi:pulumi:Stack  aws-bucket-dev  create     
     └─ aws:s3:Bucket     b                          1 error
 
Diagnostics:
  aws:s3:Bucket (b):
    error: unable to find a AWS Region config. Set the Region by using:
    
     	 • `pulumi config set aws:region us-west-2` or
     	 • An AWS Profile or
     	 • Environment variables:`AWS_REGION` or `AWS_DEFAULT`
@guin:aws-bucket:owl: echo $AWS_PROFILE
default
```

### Steps to reproduce

Run the above code wit only an AWS_PROFILE set via `aws configure`

### Expected Behavior

Pulumi should respect my AWS config

### Actual Behavior

Pulumi [checks for the AWS region via Pulumi config and environment variable](https://github.com/pulumi/pulumi-aws/blob/master/provider/resources.go#L287) but does not respect an AWS profile, despite docs and error messages implying that is all that's needed.

### Output of `pulumi about`

```
CLI          
Version      3.38.0
Go Version   go1.19
Go Compiler  gc

Plugins
NAME    VERSION
aws     5.11.0
nodejs  unknown

Host     
OS       darwin
Version  11.6
Arch     x86_64

This project is written in nodejs: executable='/Users/guin/.nvm/versions/node/v14.18.1/bin/node' version='v14.18.1'

Current Stack: dev

Found no resources associated with dev

Found no pending operations associated with dev

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/guinevere
User           guinevere
[...]

PR #2120, opened 2022-08-30
provider: Remove explicit config and EnvVar Check for Region

stack72:
Fixes: #2108
Fixes: #2104
Fixes: #2105

The need for an explict check on the configuration parameter OR the
environment variable is not quite correct, the TF provider does
fall back to using the sharedCredentials as that also includes the
region.

This also ensures we check `sharedCredentialsFiles` and not just
`sharedCredentialsFile` configuration

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -274,6 +274,25 @@ func stringValue(vars resource.PropertyMap, prop resource.PropertyKey, envs []st
 	return ""
 }
 
+func arrayValue(vars resource.PropertyMap, prop resource.PropertyKey, envs []string) []string {
+	val, ok := vars[prop]
+	var vals []string
+	if ok && val.IsArray() {
+		for _, v := range val.ArrayValue() {
+			vals = append(vals, v.StringValue())
+		}
+		return vals
+	}
+
+	for _, env := range envs {
+		val, ok := os.LookupEnv(env)
+		if ok {
+			return strings.Split(val, ";")
+		}
+	}
+	return vals
+}
+
 func stringRef(s string) *string {
 	return &s
 }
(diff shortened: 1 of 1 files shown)


provider/resources.go
```diff
@@ -337,7 +327,9 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
 
 	// lastly let's set the sharedCreds and sharedConfig file. If these are not found then let's default to the
 	// locations that AWS cli will store these values.
-	sharedCredentialsFile := stringValue(vars, "sharedCredentialsFile", []string{"AWS_SHARED_CREDENTIALS_FILE"})
+	var sharedCredentialsFile string
+	sharedCredentialsFile = stringValue(vars, "sharedCredentialsFile", []string{})
```
guineveresaenger, line 331: why do we need an empty env var array?

provider/resources.go
```diff
@@ -337,7 +327,9 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
...
+	sharedCredentialsFile = stringValue(vars, "sharedCredentialsFiles", []string{"AWS_SHARED_CREDENTIALS_FILE", "AWS_SHARED_CREDENTIALS_FILES"})
```
guineveresaenger, line 332: Are we overwriting the same variable here? if so, can we add a comment as to why?

Also - I'm not sure what it would look like to specify multiple different credentials files paths via env var([AWS does not have a multiple paths options that I could find](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html)), but even if we're just looking at the config value, it is an array type in the schema, so I'm pretty sure the `stringValue` function would hiccup here.

provider/resources.go
```diff
@@ -310,7 +300,7 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
 		SecretKey: stringValue(vars, "secretKey", []string{"AWS_SECRET_ACCESS_KEY"}),
 		Profile:   stringValue(vars, "profile", []string{"AWS_PROFILE"}),
 		Token:     stringValue(vars, "token", []string{"AWS_SESSION_TOKEN"}),
-		Region:    region,
+		Region:    stringValue(vars, "region", []string{"AWS_REGION", "AWS_DEFAULT_REGION"}),
```
guineveresaenger, line 303: thank you! 💟

guineveresaenger (changes requested):
Thank you for this PR!

provider/resources.go
stack72: the order of operations are:

* check sharedCredentialsFile
* check sharedCredentialsFiles
* check AWS_SHARED_CREDENTIALS_FILE
* AWS_SHARED_CREDENTIALS_FILES

That can't be achieved right now - so we added an extra check of the first one. They all look for the same details

stack72: I don't believe we need to make any changes here

guineveresaenger: To the first point - what I'm saying is, declaring the var and then overwriting it twice with lightly different values is confusing enough to read, at least to me, that your explanation above would be helpful to have as a comment.
if I understand correctly, the `stringValue` function will return `""` if the underlying value isn't a string. The underlying value of `sharedCredentialsFiles` is an array type. We would overwrite `sharedCredentialsFile` with an empty string every time it is set as a config var. I don't think that's the desired behavior?

stack72: The schema is mutually exclusive here - you can’t set both values

if you want me to spell that out in code then I can - I don’t see the need to do it at this point but I definitely do it today

stack72: @guineveresaenger this is ready for your review again

guineveresaenger (approved):
Sorry I missed this until now!!

Thank you for the fix 🙇‍♀️ 

:shipit:

Issue #2064, opened 2022-07-25
AWS Certificate Validation: On-page example is not valid.


MitchellGerdisch:

duro:
I am having the same problem with the TypeScript example. The block of code in that example is complex, and non-functional when used. Trying to debug it is driving me crazy.

ericpardee:
@duro The docs are definitely hit or miss.

Try looking at this example: https://github.com/pulumi/examples/blob/master/aws-apigateway-ts-routes/dns.ts

Issue #2032, opened 2022-05-12
Python Usage is incorrect for CertificateValidation


peoplesmeat:
[themes/default/content/registry/packages/aws/api-docs/acm/certificatevalidation/_index.md](https://www.pulumi.com/registry/packages/aws/api-docs/acm/certificatevalidation/)

It looks like someone copied the js version and called it python. The key in enumerate dictionary are not strings, domainValidationOptions should be domain_validation_options

Issue #2029, opened 2022-06-30
Broken example


TapTap21:
[themes/default/content/registry/packages/aws/api-docs/acm/certificate/_index.md](https://www.pulumi.com/registry/packages/aws/api-docs/acm/certificate/)

The example om ACM is still broken, see [this](https://bytemeta.vip/repo/pulumi/tf2pulumi/issues/117) code snippet on how it should look

PR #2122, opened 2022-08-31
aws.acm.CertificateValidation: Refactor the API docs to work

stack72:
Fixes: #2064
Fixes: #2032
Fixes: #2029
Fixes: #https://github.com/pulumi/registry/issues/1224


guineveresaenger (approved):
I assume we know these all work! and that we will add the Java examples later...

stack72:
> I assume we know these all work! and that we will add the Java examples later...

All these have been tested locally :)

PR #2123, opened 2022-09-01
Replace transformWithRandomString with Charset option

Frassle:


stack72 (approved):
Thank you for this :)

PR #2132, opened 2022-09-09
Corrected db_name deprecated doc string

mothran:
Fixed deprecation warning to use correct field name.


thomas11:
Thank you for your contribution, @mothran, and apologies for the late response. Looks like this PR was made obsolete by other changes in the meantime, removing the reference to `dbName` altogether.

PR #2148, opened 2022-09-20
preconfigureCallback: Allow Env Vars for Credentials Validation

stack72:
In order to ensure that the user doesn't need to add anything to
state to skip a credentials validation or to skip the metadata api
check, we now allow a user to set the value via environment variables.

The order of operations is that we check for provider configuration
first and if we can't find it, we check for the environment variable

Therefore, a person who wants to skip the credentials validation can
now set:

AWS_SKIP_CREDENTIALS_VALIDATION=true when running pulumi up


viveklak (approved):

stack72:
@viveklak this pr opens a fix for https://github.com/pulumi/pulumi-kubernetes-operator/issues/291

PR #2149, opened 2022-09-21
Read in role ARN in preConfigure validation

### LINKED ISSUE #2144: Errors with `aws:route53/getZone:getZone` after upgrading to 5.14.0
### What happened?

When i upgraded my project to 5.14.0 i started getting the following error a lot (20+ times) when i run a preview. Downgrading to 5.13.0 makes the problem go away.

```
pulumi:pulumi:Stack (stack-name):
  error: Running program '/path/to/project' failed with an unhandled exception:
  Error: invocation of aws:route53/getZone:getZone returned an error: unable to validate AWS credentials. Make sure you have:
  
      • Set your AWS region, e.g. `pulumi config set aws:region us-west-2`
      • Configured your AWS credentials as per https://pulumi.io/install/aws.html
      You can also set these via cli using `aws configure`.
  
  
      at Object.callback (/path/to/project/node_modules/@pulumi/runtime/invoke.ts:159:33)
      at Object.onReceiveStatus (/path/to/project/node_modules/@grpc/grpc-js/src/client.ts:338:26)
      at Object.onReceiveStatus (/path/to/project/node_modules/@grpc/grpc-js/src/client-interceptors.ts:426:34)
      at Object.onReceiveStatus (/path/to/project/node_modules/@grpc/grpc-js/src/client-interceptors.ts:389:48)
      at /path/to/project/node_modules/@grpc/grpc-js/src/call-stream.ts:276:24
      at processTicksAndRejections (node:internal/process/task_queues:77:11)
```

### Steps to reproduce

1. upgrade from 5.13.0 to 5.14.0
2. run pulumi up

### Expected Behavior

I should get a preview of the my changes

### Actual Behavior

The process crashes.

### Output of `pulumi about`

_No response_

### Additional context

_No response_

### Contributing

guineveresaenger:
Fixes #2144.

I followed the pattern we use in pulumi-azure to surface the underlying error.
I am open to other suggestions but I feel at this point it is more helpful to the user to forward the error from TF or AWS.

A screenshot of the new error surfacing when `aws-region is not set:

Screen Shot 2022-09-20 at 5 20 12 PM

Sample terminal output for an invalid role ARN, with identity details removed:
```
Diagnostics:
  aws:ec2:VpcIpam (guin):
    error: unable to validate AWS credentials. Details: IAM Role (xxxxxxxx/xxxx) cannot be assumed.
    
    There are a number of possible causes of this - the most common are:
      * The credentials used in order to assume the role are invalid
      * The credentials do not have appropriate permission to assume the role
      * The role ARN is not valid
    
    Error: operation error STS: AssumeRole, https response error StatusCode: 403, RequestID: yyyyyyyyyyyyyyy, api error AccessDenied: User: xxxxxxxxxxx is not authorized to perform: xxxxxxxxxx
    
    
    Make sure you have:
    
     	 • Set your AWS region, e.g. `pulumi config set aws:region us-west-2`
     	 • Configured your AWS credentials as per https://pulumi.io/install/aws.html
     	 You can also set these via cli using `aws configure`.
 
```

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -323,14 +323,15 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
 	}
 
 	if details, ok := vars["assumeRole"]; ok {
-		assumeRoleDetails := resource.NewPropertyMap(details)
+
 		assumeRole := awsbase.AssumeRole{
-			RoleARN:     stringValue(assumeRoleDetails, "roleArn", []string{}),
-			ExternalID:  stringValue(assumeRoleDetails, "externalId", []string{}),
-			Policy:      stringValue(assumeRoleDetails, "policy", []string{}),
-			SessionName: stringValue(assumeRoleDetails, "sessionName", []string{}),
+			RoleARN:     stringValue(details.ObjectValue(), "roleArn", []string{}),
+			ExternalID:  stringValue(details.ObjectValue(), "externalId", []string{}),
+			Policy:      stringValue(details.ObjectValue(), "policy", []string{}),
+			SessionName: stringValue(details.ObjectValue(), "sessionName", []string{}),
 		}
 		config.AssumeRole = &assumeRole
+
 	}
 
 	// By default `skipMetadataApiCheck` is true for Pulumi to speed operations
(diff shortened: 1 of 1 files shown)


jkisk (approved):
Thank you, I agree we should pass along the error to assist user with debugging here

guineveresaenger:
@stack72 that was my concern as well - however, the increasing amounts of different errors a user can run into here is not something I'm comfortable swallowing anymore. The alternative is parsing the error string for mentions of `assumeRole` or `"Assume Role: role ARN not set"` and...I think that would be an antipattern.

richmeij:
Hi all, is there any update on this PR?

wannessels:
@stack72 can you elaborate on this?

> I think we should tread carefully here... this may expose TF implementation details

What details might leak here? Is that really problematic?

Right now we (and reading through issues a lot of others are well) are stuck on 5.13.0, and even there we are hitting issues with roleArn not setting/propagating properly. It sounds like this patch could fix a lot if not all of those issues.

viveklak (approved):

PR #2175, opened 2022-10-12
Upgrade terraform-provider-aws to v4.34.0

danielrbradley:
Here's what I ended up doing in the TF fork to get it building: https://github.com/pulumi/terraform-provider-aws/commit/10b16ea5e764e290ea437df39eea6b2b15527728

Points to check:
- Naming of inspector V2 - should this have the `V` in it even though the resource token doesn't?
- There's two new `appconfig` datasources which are plurals ... which I assume mean they list resources. Should we be including these?


guineveresaenger:
To answer your second question: yes, we include plurals; see also [availability zone/s](https://github.com/pulumi/pulumi-aws/blob/master/provider/resources.go#L4620)

The diff won't load for me but I assume the TF resource token doesn't contain a `_v2_` as it does [here for instance](https://github.com/pulumi/pulumi-aws/blob/master/provider/resources.go#L4666) - this seems like an oversight upstream :/ but our mappings hold true regardless of misspellings 😅 so there's a case to be made here to include the V2.

guineveresaenger:
Looking at your branch (without the diff), just one request for changes:

- Instead of `cloudwatchevidently`, should we go with the upstream naming scheme and just use Evidently? It _seems to be_ a sub-category of CloudWatch, but upstream seems to not care. I'd rather be internally consistent than try to second-guess any naming scheme having to do with AWS 😅 

- The breaking changes on `filters` for User are expected via upstream release notes. ✔️

danielrbradley:
Thanks for the feedback @guineveresaenger 

Here's the TF docs for the two new resources:
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/evidently_project
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/inspector2_organization_configuration

It's super confusing because the docs include "Cloudwatch" and the "V" but the resource/module names don't. Are you saying we would normally stick to matching their resource names? I.e. drop the "cloudwatch" and stick with just `inspector2`?

Will aim to get this merged and released first thing in the morning with the changes 👍

PR #2215, opened 2022-11-15
Upgrade to upstream v4.39.0

danielrbradley:
This switches to using the new patching process, so there's a large number of docs changes.


danielrbradley:
Looks like the breaking changes indicate there was a few instances where we've manually implemented `tags_all` in our fork. Will need to investigate and re-add these.

thomas11 (approved):
I only spot-checked a number of sdk/go files, but the changes look great. Approved pending the tags_all issue.

danielrbradley:
For the errors: `missing input "tagsAll"`, it appears that `tagsAll` should never have been an input field - so this is correcting that bug. My understanding is that `tagsAll` is a computed, output-only property which is the combination of the `tags` input, plus any tags configured at the provider level.

For the errors: `Resource "aws:s3/bucketV2:BucketV2" missing input "arn" / missing input "hostedZoneId"`, this is an upstream change which removes the `Optional: true` from the type-spec. Both of these properties are also set to `Computed: true`, which therefore makes sense that they shouldn't be inputs - they're only computed outputs. What's not entirely clear is why they were made into inputs if they were optional AND computed. Perhaps we interpert "optional AND computed" as "this can be computed, but could optionally be provided by the user".

For the errors: `Resource "aws:secretsmanager/secretRotation:SecretRotation" missing input "tags" /  missing output "tags"`, this is an upstream change where the tags property has been removed. Following the PR - this property does not apply to this resource and was therefore useless.

PR #2247, opened 2022-12-05
Upgrade to v4.50.0 with submodule patching

danielrbradley:
Adjust process to apply replacements in the local submodule checkout. This should make the process a little slicker so we're not committing tooling into the fork.

- [x] Add script to auto-reset changes between each apply.
- [x] Add submodule option to ignore dirty worktree - avoid the CI failing.
- [x] Integrate into makefile
- [x] Add submodule checkout during the build
- [ ] Add auto-upgrade workflow to attempt rebase & open PR


viveklak (approved):

thomas11 (approved):

Makefile
```diff
@@ -113,4 +113,18 @@ tfgen: install_plugins
...
 
+init_upstream:
+	@if [ ! -f "upstream/.git" ]; then \
+			echo "Initializing upstream submodule" ; \
+			(cd upstream && git submodule update --init && git remote add source git@github.com:hashicorp/terraform-provider-aws.git) ; \
+		fi; \
+
+update_upstream: init_upstream
+	# Find latest tag, create new branch, rebase on new tag, push
+	export TAG=$$(cd upstream && git for-each-ref refs/tags --sort=-taggerdate --format='%(refname:short)' --count=1) && \
```
thomas11: Does this not `cd upstream` twice?

Even if not, would be a bit simpler as
```
cd upstream
EXPORT ...
git checkout ...
```

thomas11: Ah, fixed in a later commit.

Makefile
```diff
@@ -119,13 +119,15 @@ init_upstream:
...
 		fi; \
 
-patch_upstream:
-	cd upstream-tools && \
-		yarn install --frozen-lockfile --silent
-	cd upstream && git checkout . && cd - || exit
-	cd upstream-tools && \
-		yarn --silent run tf-patch apply --cwd ../upstream && \
-		yarn --silent run tf-patch check --cwd ../upstream
+patch_upstream: init_upstream
```
thomas11, line 122: Don't you need to `cd ..` in between the steps? I don't know make's intricacies like you do though :)

thomas11 (approved):
The approach LGTM!

Makefile
danielrbradley: Make runs every command within a target in an isolated shell by default. This can be turned off in newer versions (4.x) using `.ONESHELL` I think, but we're still supporting 3.x as that's what macos ships with so we're stuck with either chaining everything into a single line with `&&` or just cd'ing before each command.

lblackstone (approved):

PR #2287, opened 2022-12-24
Fix eventBus support in EventRule#onEvent

### LINKED ISSUE #1323: CloudWatch EventRule onEvent does not use eventBusName
When specifying the `eventBusName` on an `EventRule`, the `onEvent` seems to just use the `default` event bus.

```typescript

const eventBusHealth = new aws.cloudwatch.EventBus("health");

const ec2IssueEventRule = new aws.cloudwatch.EventRule("ec2-issue", {
  description: "Trigger Lambda to respond to EC2 issue events",
  eventPattern: JSON.stringify({
    source: ["aws.health"],
    "detail-type": ["AWS Health Event"],
    detail: {
      service: ["EC2"],
      eventTypeCategory: ["issue"],
    },
  }),
  eventBusName: eventBusHealth.name,
});

ec2IssueEventRule.onEvent("ec2-issue", (event) => {
  console.log(event);
});
```

This results an error since the rule doesn't exist on the default bus.

```
  aws:cloudwatch:EventTarget (ec2-issue):
    error: 1 error occurred:
        * Creating CloudWatch Events Target failed: ResourceNotFoundException: Rule ec2-issue-3bc3cfa does not exist on EventBus default.
```

lukehoban:
In cases where the EventRule specified an eventBus other than `default`, this was not being used correctly in the EventTarget.  We can just pass along the eventBusName from the rule to the target.

Fixes #1323.

### DIFF
--- a/sdk/nodejs/cloudwatch/eventRuleMixins.ts
+++ b/sdk/nodejs/cloudwatch/eventRuleMixins.ts
@@ -107,6 +107,7 @@ export class EventRuleEventSubscription extends lambda.EventSubscription {
             rule: this.eventRule.name,
             arn: this.func.arn,
             targetId: name,
+            eventBusName: this.eventRule.eventBusName.apply(x => x!),
         }, parentOpts);
 
         this.registerOutputs();
--- a/examples/eventbus/Pulumi.yaml
+++ b/examples/eventbus/Pulumi.yaml
@@ -1,0 +1,3 @@
+name: eventbus
+runtime: nodejs
+description: A simple example of using the `EventBus` APIs.
--- a/examples/eventbus/index.ts
+++ b/examples/eventbus/index.ts
@@ -1,0 +1,38 @@
+// Copyright 2016-2018, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as pulumi from "@pulumi/pulumi";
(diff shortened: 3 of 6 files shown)


sdk/nodejs/cloudwatch/eventRuleMixins.ts
```diff
@@ -107,6 +107,7 @@ export class EventRuleEventSubscription extends lambda.EventSubscription {
             rule: this.eventRule.name,
             arn: this.func.arn,
             targetId: name,
+            eventBusName: this.eventRule.eventBusName.apply(x => x!),
```
Zaid-Ajaj, line 110: is it possible that `this.eventRule.eventBusName` can be undefined? When constructing an `EventRule`, `eventBusName` is an [optional input](https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/cloudwatch/eventRule.ts#L234), the docs does say that it takes `default` event bus when omitted but I am not sure if that makes sure that `eventBusName` is always populated. 

In case `this.eventRule.eventBusName` can be undefined:
```js
eventBusName: this.eventRule.eventBusName == undefined 
  ? undefined
  : this.eventRule.eventBusName.apply(x => x!),
```
LGTM otherwise

Zaid-Ajaj (approved):

sdk/nodejs/cloudwatch/eventRuleMixins.ts
lukehoban: It is possible it is undefined, and if it is, this code will result in passing that undefined value through into the outer `eventBusName` property, which is correct.

Unfortunately, our TypeScript typing of optional inputs is "wrong" per https://github.com/pulumi/pulumi/issues/6175 (with a proposed fix in https://github.com/pulumi/pulumi/pull/6323 that has likely bitrotted).  This code I added here works around that "wrong"-ness, while still behaving correctly.

The code you suggested unfortunately won't work - because `this.eventRule.eventBusName` itself is not undefined, it is an `Output<undefined>`.  It is the value inside the Output which might be undefined.

Zaid-Ajaj: Aah I see, thanks for explaining the context behind this! 

> The code you suggested unfortunately won't work

Yeah I was assuming the outer output itself could be undefined but it seems that only the data underneath could be that.

PR #2339, opened 2023-01-31
Committing a test with known non-determinism

iwahbe:
The test is built against the latest version of tfbridge.


jazzyfresh:
Mark this as a draft please

iwahbe:
https://github.com/pulumi/pulumi-terraform-bridge/pull/776 closes the original issue. This is no longer needed to coordinate debugging.

PR #2347, opened 2023-02-02
Fix: Upgrading terraform bridge to 3.39.1

rshade:


rshade:
@kpitzen @iwahbe is this a docstring bug? https://github.com/pulumi/pulumi-aws/pull/2347/files#diff-143888737f1085bef2bd2b9c1dc0904a5bb8f20988cac636f73839ecf9de19f5L84

sdk/python/pulumi_aws/directconnect/get_location.py
```diff
@@ -107,7 +107,7 @@ def get_location(location_code: Optional[str] = None,
     Retrieve information about a specific AWS Direct Connect location in the current AWS Region.
     These are the locations that can be specified when configuring `directconnect.Connection` or `directconnect.LinkAggregationGroup` resources.
 
-    > **Note:** This data source is different from the `directconnect.get_locations` data source which retrieves information about all the AWS Direct Connect locations in the current AWS Region.
+    > **Note:** This data source is different from the _directconnect_get_locations_ data source which retrieves information about all the AWS Direct Connect locations in the current AWS Region.
```
thomas11, line 110: This change from code to emphasis is all over the Python SDK. I don't think this will render correctly if the inner `_` are not escaped. It also doesn't seem particularly better? Where does this come from?

iwahbe: Given that this is a change only in specific languages, my guess is it is a codegen change. I agree that it doesn't look like a positive change.

kpitzen: Agreed. This is likely incorrect.

rshade: I opened a ticket on it: https://github.com/pulumi/pulumi/issues/12056

rshade: linking to https://github.com/pulumi/pulumi/issues/12075

thomas11:
superceded by #2366

PR #2390, opened 2023-02-28
feat(node-js-sdk): support for filterCriteria in dynamodb TableEventSubscription mixin

sladkoff:
I noticed that it's not currently possible to pass dynamodb streams filter criteria when using the following mixin form:

```ts
new dynamodb.Table(...).onEvent(...)
```

I drafted up this PR to add support for the parameter. I have no idea if this is enough for a contribution but maybe this can serve as a base for someone more knowledgeable than me.


thomas11:
Hi @sladkoff, thank you for your contribution! It looks good except that it fails a test due to an outdated go.mod. Not quite sure what's up with that... could you try rebasing on the latest master and running `go mod tidy` in `provider/shim` and `provider/`?

thomas11 (approved):

thomas11:
Merged - thank you for the contribution, @sladkoff !

PR #2401, opened 2023-03-06
Use larger GitHub Actions runner for publish step

### LINKED ISSUE #2400: Publish CI step fails with "no space left on device"
This has failed for the past two merges (and again when the workflow was re-run):
https://github.com/pulumi/pulumi-aws/actions/runs/4308251155/jobs/7528823156

squaremo:
The `publish` step has started failing with the error "no space left on device":

    https://github.com/pulumi/pulumi-aws/actions/runs/4308251155/jobs/7528823156

The default runner has 7GB of disk space, and I hypothesise that space is taken up by the Go build cache -- including the AWS SDK, six times (arch x OS), ends up being a lot of object files. As a mitigation, we can use a larger runner instance, which will come with signficantly more disk space.

Fixes #2400

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -257,7 +257,7 @@ jobs:
   publish:
     name: publish
     needs: test
-    runs-on: ubuntu-latest
+    runs-on: pulumi-ubuntu-8core
     steps:
     - name: Checkout Repo
       uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


squaremo:
Attempting the publish step in https://github.com/pulumi/pulumi-aws/actions/runs/4343331208/jobs/7586417422

danielrbradley (approved):

PR #2405, opened 2023-03-08
Add a smoke test of the patched upstream provider

squaremo:
In upstream/ we have the underlying provider code, patched with our own improvements. There are runtime checks -- for example, that resources are not declared more than once -- so, it is worth paying attention to any error return value when calling NewProvider.

 - add a test that instantiating NewProvider doesn't return an error
 - panic if NewProvider returns an error (though we might hope the above test catches that)

Hat tip to @t0yv0 for the suggestion and exemplar code 🙏


kpitzen (approved):

squaremo:
Lastly: the test fails if I do this:

```bash
cd upstream
git checkout origin/patched-4.57.0 # the known-bad branch which exposed this lack of test
cd ../provider/shim
go mod tidy && go test -v
```

So we can have some confidence the test will pick up problems with upstream/.

PR #2408, opened 2023-03-09
Upgrade upstream to v4.57.0

danielrbradley:


.github/workflows/run-acceptance-tests.yml
```diff
@@ -252,6 +252,8 @@ jobs:
       with:
         ref: ${{ env.PR_COMMIT_SHA }}
         submodules: true
+    - name: Init upstream
```
squaremo, line 255: It seems reasonable that leaving this out could upset things; but could you record in the commit message what fails and how this fixes it, please.

I think you'll need to replicate this change in the other places this test runs, too.

upstream-tools/replacements.json
```diff
@@ -698,6 +706,10 @@
     }
   ],
   "website/docs/r/connect_contact_flow.html.markdown": [
+    {
```
squaremo, line 705: This seems to duplicate the entry immediately after it -- am I misinterpreting what's going on?

upstream-tools/replacements.json
```diff
@@ -1849,11 +1881,29 @@
       "new": "Resource for managing an AWS NetworkManager ConnectAttachment."
     }
   ],
+  "website/docs/r/networkmanager_connect_peer.html.markdown": [
+    {
+      "old": "  Terraform resource for managing an AWS NetworkManager Connect Peer.",
```
squaremo, line 1882: Another (near-)duplicate entry. Are both of these needed?

upstream-tools/pre-replacements.json
```diff
@@ -4893,35 +4865,8 @@
     {
       "old": "-> **NOTE:** The parameter `server_side_encryption_configuration` is deprecated.\nUse the resource [`aws_s3_bucket_server_side_encryption_configuration`](s3_bucket_server_side_encryption_configuration.html) instead.\n"
     },
-    {
```
squaremo, line 4896: This looks really similar (and in the same place, I think, under the key [`"website/docs/r/s3_bucket.html.markdown"`](https://github.com/pulumi/pulumi-aws/pull/2408/commits/0265ad6c5e1d91afbbd22a2b42cfdfd9a469250a#diff-a1ca436dd43a4302dee354a5c8a023e3828382901e55b4fc89230d7c49feb412R2263)) to a replacement that was added in the previous commit. I'm not clear what this means -- does the text get changed now, and not removed?

squaremo (commented):
Looks sound; I will have a look at the rebased upstream/, too. The breaking changes reported will need chasing down, to make sure they are expected.
My comments here are mainly queries about what's going on -- it's hard to understand the changes to replacements.json (though I prefer it to the prior state of the art!).

upstream-tools/replacements.json
danielrbradley: It varies by the comma it seems - I've just pulled these in from Thomas. No harm having near-duplicates, but the latter rule could probably be removed with this new one.

upstream-tools/replacements.json
danielrbradley: Yeah, only the second should be required here, but won't cause any issue having both right now.

upstream-tools/replacements.json
squaremo: Well spotted! (or: poorly spotted by me!) OK so the one that comes first does not have commas in the match or the replacement text -- I guess they were removed in the upstream. The second one presumably won't match because the match text doesn't match.

danielrbradley:
Issue investigation:
- EC2 Fleet `launchTemplateConfig` has changed from being exactly 1 element to being between 0 and 50 elements. This causes our property to be renamed to a plural.
- ECS `ServiceServiceConnectConfigurationService` `clientAlias` has now added maxItems 1, but optional - so no more pluralisation.
- Autoscaling `PolicyPredictiveScalingConfigurationMetricSpecification` `targetValue` has had it's type changed upstream from int to float.
- AppFlow `FlowSourceFlowConfigSourceConnectorPropertiesSapoData` `object` property has been renamed to `AttrObjectPath`
Investigating if there's a fix for the first two

upstream-tools/replacements.json
squaremo: Should we expect that second one to be removed as unused in the commit after this one?

squaremo: Ah! There are two very similar snippets, one has a comma and one doesn't:
https://github.com/pulumi/terraform-provider-aws/blob/fc7794404f569e5ee9f1c80963daee017ccb8f0f/website/docs/r/connect_contact_flow.html.markdown?plain=1#L59
https://github.com/pulumi/terraform-provider-aws/blob/fc7794404f569e5ee9f1c80963daee017ccb8f0f/website/docs/r/connect_contact_flow.html.markdown?plain=1#L85
As you were.

.github/workflows/run-acceptance-tests.yml
squaremo: Turns out this wasn't necessary -- updating the shim/go.mod was enough.

squaremo:
Could you explain the justification for the MaxItemsOne reversion in the commit message, please. (I'd include a link to the breaking changes policy, too.) As recently as this particular update, a (rebased) commit that didn't come with an explanation held us up until we could chase down why it was done. This change is less likely to cause difficulties during a rebase specifically, but there are other situations in which having the explanation with the change avoids hold-ups.

PR #2414, opened 2023-03-15
Export patches to local repo

danielrbradley:
Aim: We want to be able to review how patches have evolved between upgrades.

1. Find commits since the lastest tag in the upstream repo
2. Commit patch files into a subfolder for review

Example diff for last upgrade (noisy because of SDK changes): https://github.com/pulumi/pulumi-aws/compare/5.30.1-patches..export-patches-to-repo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,6 +132,10 @@ patch_upstream: init_upstream
 	cd upstream-tools && yarn --silent run apply
 	@# Check for any pending replacements
 	cd upstream-tools && yarn --silent run check
+	rm upstream-patches/*
+	cd upstream && \
+		LAST_TAG=$$(git describe --abbrev=0 --tags) && \
+		git format-patch  -o ../upstream-patches --minimal --no-signature HEAD...$${LAST_TAG}
 
 update_upstream: init_upstream
 	@echo "\033[1;33mupdate_upstream is still under construction and will likely fail.\033[0m"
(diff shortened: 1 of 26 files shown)


Makefile
```diff
@@ -140,7 +140,15 @@ init_upstream:
 			(cd upstream && git submodule update --init && git remote add source git@github.com:hashicorp/terraform-provider-aws.git) ; \
 		fi; \
 
-patch_upstream: init_upstream
+export_upstream_patches: init_upstream
+ifeq ($(shell cd upstream && git rev-parse --is-shallow-repository), false)
+	find upstream-patches -type f -delete
+	cd upstream && \
+		LAST_TAG=$$(git describe --abbrev=0 --tags) && \
+		git format-patch  -o ../upstream-patches --minimal --no-signature HEAD...$${LAST_TAG}
```
thomas11, line 148: Generally, `HEAD` could be anything. You're relying on the assumption that it's the latest `patched-...` branch as per the upgrade playbook? Can we make that safer?

Makefile
```diff
@@ -140,7 +140,15 @@ init_upstream:
 			(cd upstream && git submodule update --init && git remote add source git@github.com:hashicorp/terraform-provider-aws.git) ; \
 		fi; \
 
-patch_upstream: init_upstream
+export_upstream_patches: init_upstream
+ifeq ($(shell cd upstream && git rev-parse --is-shallow-repository), false)
+	find upstream-patches -type f -delete
+	cd upstream && \
+		LAST_TAG=$$(git describe --abbrev=0 --tags) && \
```
thomas11, line 147: Depends on one's depth of Git expertise, these two lines might warrant a comment. I had to test or look up `describe --abbrev=0` and `...` (as opposed to `..`).

Makefile
danielrbradley: The challenge here is that if we've just checked out the submodule directly we won't be on a branch, but a detached head - pointing directly to the commit hash in the parent repo. I think the risk is low as you'd just see a completely different set of patches exported which should raise questions in the review.

thomas11 (approved):

PR #2418, opened 2023-03-20
Mapped TopicSubscription args

andreujuanc:
This should be enough to correctly expose the rest of the arguments of the topic subscription.


andreujuanc:
Oh, so this is Autogenrated? I haven't touch the go files before. And haven't play with that.

andreujuanc:
@thomas11 Hi, glad you found this PR worth merging. Cheers.

PR #2439, opened 2023-03-27
Adds a phone make target to check local schema against master

kpitzen:
The idea behind this addition is to avoid the CI round-trip for ensuring our currently generated schema produces the results that we expect. Currently missing from schema-tools is an implementation of the version command which means we'll download a new artifact every time we run.

We also don't have a very safe OS check, etc, but it will at least work on apple machines for the time being.


kpitzen:
The most recent run:
image

danielrbradley:
Really like having the schema check in here. In azure-native we download binaries, with a couple of minor differences that you might want to adopt here too:

1. We download locally into the repo's `bin` directory - so this is isolated from any global installations.
2. We use a `[PROGRAM].version` file to specify the exact version of each binary to download e.g. `schema-tools.version`
3. We take a dependency on `schema-tools.version` in the makefile so it will be re-downloaded and overwritten if the version file changes (either from a local edit or when checking out).

Example:

https://github.com/pulumi/pulumi-azure-native/blob/db073eee29704f3aa2b36524ffb3be59b6d258ca/Makefile#L157-L167

provider/cmd/pulumi-resource-aws/schema.json
```diff
@@ -440295,7 +440295,7 @@
...
...\n          - effect: Allow\n            principals:\n              - type: Federated\n                identifiers:\n                  - cognito-identity.amazonaws.com\n            actions:\n              - sts:AssumeRoleWithWebIdentity\n            conditions:\n              - test: StringEquals\n                variable: cognito-identity.amazonaws.com:aud\n                values:\n                  - ${mainIdentityPool.id}\n              - test: ForAnyValue:StringLike\n                variable: cognito-identity.amazonaws.com:amr\n                values:\n                  - authenticated\n  authenticatedRolePolicyPolicyDocument:\n    fn::invoke:\n      Function: aws:iam:getPolicyDocument\n      Arguments:\n        effect: Allow\n        actions:\n          - mobileanalytics:PutEvents\n          - cognito-sync:*\n          - cognito-identity:*\n        resources:\n          - '*'\n```\n{{% /example %}}\n{{% /examples %}}\n\n## Import\n\nCognito Identity Pool Roles Attachment can be imported using the Identity Pool ID, e.g.,\n\n```sh\n $ pulumi import aws:cognito/identityPoolRoleAttachment:IdentityPoolRoleAttachment example us-west-2:b64805ad-cb56-40ba-9ffc-f5d8207e6d42\n```\n\n ",
```
danielrbradley, line 440298: Can't figure out what's changed here ... or why it's changed - might just need to revert this.

danielrbradley (approved):
All good to go if the build passes! It's possible that the schema change might cause a dirty working tree though.

provider/cmd/pulumi-resource-aws/schema.json
danielrbradley: Oh, something related to the java version maybe?

PR #2451, opened 2023-04-04
Update vpcPeeringConnectionAccepter.go

georgettica:
after touching peering connections (which will be changed in another PR) I found this caused me some issues in the documentation

LMK if I should change this deeper in the stack or is this enough


mikhailshilkov:
@georgettica Thank you for opening this PR.

All the examples are auto-generated, so it doesn't make much sense to edit them manually, since they will be overwritten on the next run. It probably makes sense to open an issue to track the problem and eventually fix it in the generation.

PR #2453, opened 2023-04-04
WIP: Serve a mix of traditional and Plugin Framework based resources

t0yv0:


provider/cmd/pulumi-resource-aws/main.go
```diff
@@ -22,15 +22,19 @@ import (
...
 	pf "github.com/pulumi/pulumi-terraform-bridge/pf/tfbridge"
-	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+	//"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 )
 
 //go:embed schema-embed.json
 var pulumiSchema []byte
 
+//go:embed bridge-metadata.json
+var bridgeMetadata []byte
```
t0yv0, line 32: It think this bit was really needed to get past an error, since the PF provider needs to read renames data generated by tfgen. Error msgs need work unfortunately.

t0yv0:
Still failing with these changes.

There is something in PF that's unhappy/confused about the provider-level configuration schema. We'll have to work around that .. Hm..

```
./bin/pulumi-resource-aws                                         ~/code/pulumi-aws
error: fatal: failed to Init GRPC to register RPC handlers: failed to create resource provider: NewConfigEncoder failed: cannot derive an encoder for provider config: missing property "ec2_metadata_service_endpoint"
```

t0yv0:
Progress, this can now deploy this program:

```
name: check
runtime: yaml
description: A minimal Pulumi YAML program
outputs: {}
resources:
  bucket1:
    type: aws:s3:Bucket
  cidrcol1:
    type: aws:route53/cidrCollection:CidrCollection
    properties:
     name: collection-1
variables: {}
```

provider/shim/shim.go
```diff
@@ -2,14 +2,26 @@ package shim
...
+}
+
+func NewUpstreamProvider(ctx context.Context) (UpstreamProvider, error) {
+	primary, err := provider.New(ctx)
 	if err != nil {
-		panic(err)
+		return UpstreamProvider{}, err
 	}
-	return prov
+	pf := fwprovider.New(primary)
```
t0yv0, line 22: @iwahbe this is the state sharing that removes the config-related panic. This seems to work. We end up calling Configure twice through muxer code but it seems "idempotent enough".

t0yv0:
I've accidentally moved `upstream` pointer here from master branch and some of the schema changes might be coming out of that, need to revert.

t0yv0:
Closing, still working out the right approach here.

PR #2461, opened 2023-04-12
Upgrade to v4.62.0

iwahbe:
Fixes #2457

The breaking change corresponds to this item in the [upstream changelog](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v4.62.0):
> - resource/aws_medialive_channel: Fix attribute certificate_mode spelling in rtmp_output_settings (https://github.com/hashicorp/terraform-provider-aws/issues/30224)

Closes https://github.com/pulumi/pulumi-aws/issues/2458

PR #2464, opened 2023-04-13
WIP: expose PF-based resources

t0yv0:


provider/shim/shim.go
```diff
@@ -2,14 +2,26 @@ package shim
...
 	if err != nil {
-		panic(err)
+		return UpstreamProvider{}, err
 	}
-	return prov
+	pf := fwprovider.New(primary)
+	return UpstreamProvider{
+		SDKV2Provider:           primary,
+		PluginFrameworkProvider: pf,
+	}, nil
```
iwahbe, line 26: @t0yv0 I think the reason we are seeing behavior from the SDK side effect the PF side is because we are passing back the same provider (with its backing state).

t0yv0: We have to do it. If we don't do it PF side is unable to deploy resources.

t0yv0: Specifically it panics with malconfigured AWS client, which only gets configured in sdkv2-style callback. I can dig this up if it doesn't ring a bell.

t0yv0:
Noticing weird SetAutoNaming interaction with @iwahbe - new resources based on PF do not yet support auto naming a runtime until pulumi/pulumi-terraform-bridge#917 is done but as written the schema-time SetAutonaming call works and makes their required "name" properties optional. The warning is elided due to incorrect check. 

Could move resource definitions around so that SetAutoNaming does not apply to them. 

Wondering if this is a blocker for releasing an update, can we release with a required "name" property and then relax to optional for the new resources?

t0yv0:
Getting closer to passing checks, now need to regenerate the SDKs.

iwahbe:
@t0yv0 With regards to https://github.com/pulumi/pulumi-aws/actions/runs/4745234484/jobs/8428050038?pr=2464. We have seen some weirdness with this resource before: https://github.com/pulumi/pulumi-aws/pull/2461#discussion_r1169296239

CC @kpitzen

t0yv0:
Oof. SO this is non-deterministic?

t0yv0:
I also have a Diff on schema.json for sdk/java/src/main/java/com/pulumi/aws/cognito/IdentityPoolRoleAttachment.java - that must be why, checking up on this.

t0yv0:
Interesting. We finally get to a place where it's running acceptance tests. Seeing a lot of this:

```
Diagnostics:
[411](https://github.com/pulumi/pulumi-aws/actions/runs/4747040603/jobs/8450477430#step:23:417)
    aws:elasticloadbalancingv2:LoadBalancer (test):
[412](https://github.com/pulumi/pulumi-aws/actions/runs/4747040603/jobs/8450477430#step:23:418)
      error: Resource type 'aws:elasticloadbalancingv2/loadBalancer:LoadBalancer' not found.
[413](https://github.com/pulumi/pulumi-aws/actions/runs/4747040603/jobs/8450477430#step:23:419)
  
[414](https://github.com/pulumi/pulumi-aws/actions/runs/4747040603/jobs/8450477430#step:23:420)
    pulumi:pulumi:Stack (alb-legacy-py-p-it-fv-az254-2-alb-new-py-d1204a7c):
[415](https://github.com/pulumi/pulumi-aws/actions/runs/4747040603/jobs/8450477430#step:23:421)
      warning: LoadBalancer is deprecated: aws.elasticloadbalancingv2.LoadBalancer has been deprecated in favor of aws.lb.LoadBalancer
[416](https://github.com/pulumi/pulumi-aws/actions/runs/4747040603/jobs/8450477430#step:23:422)
```

mikhailshilkov:
Is this PR still relevant?

iwahbe:
> Is this PR still relevant?

I believe it was rendered unnecessary by https://github.com/pulumi/pulumi-aws/pull/2557.

PR #2471, opened 2023-04-19
Fetches upstream source remote after adding

kpitzen:


thomas11 (approved):

Makefile
```diff
@@ -137,7 +137,7 @@ bin/pulumi-java-gen:
 init_upstream:
 	@if [ ! -f "upstream/.git" ]; then \
 			echo "Initializing upstream submodule" ; \
-			(cd upstream && git submodule update --init && git remote add source git@github.com:hashicorp/terraform-provider-aws.git) ; \
+			(cd upstream && git submodule update --init && git remote add source git@github.com:hashicorp/terraform-provider-aws.git && git fetch source) ; \
```
t0yv0, line 140: Consider removing the warning "this file is auto-generated" if we're editing it as a source file now?

t0yv0 (approved):
Thanks a ton!

Do you think this is the right thing to do elsewhere also? I'm a bit confused about git submodules. If this is the right thing I can adopt it into ci-mgmt and roll out to ecosystem repos.

LGTM.

kpitzen:
> Do you think this is the right thing to do elsewhere also? 

I think so, yes - at least a `git fetch --all`to be sure any remotes are up to date.

PR #2472, opened 2023-04-19
Allow custom build parallelism

t0yv0:
Same change as just got rolled out to ecosystem providers https://github.com/pulumi/ci-mgmt/pull/386 - applied manually to this provider.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ JAVA_GEN_VERSION := v0.7.1
 TESTPARALLELISM := 10
 WORKING_DIR := $(shell pwd)
 SCHEMA_TOOLS_VERSION := $(shell cat .schema-tools.version)
+PULUMI_PROVIDER_BUILD_PARALLELISM ?= -p 2
 
 # Override repo path discovery because we're using a local checkout instead of the go mod
 export PULUMI_REPO_PATHS=github.com/hashicorp/terraform-provider-aws=$(WORKING_DIR)/upstream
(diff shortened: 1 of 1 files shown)


iwahbe (approved):
Looks good from my end. We should probably make sure that someone from @pulumi/providers gives it the 👌.

Makefile
```diff
@@ -13,6 +13,7 @@ JAVA_GEN_VERSION := v0.7.1
 TESTPARALLELISM := 10
 WORKING_DIR := $(shell pwd)
 SCHEMA_TOOLS_VERSION := $(shell cat .schema-tools.version)
+PULUMI_PROVIDER_BUILD_PARALLELISM ?= -p 2
```
t0yv0, line 16: This is a behavior change from -p 1 but @thomas11 recommended this. We can make it -p 1. Still lets me run with -p 6 locally.

t0yv0:
b/sdk/go/aws/cognito/identityPoolRoleAttachment.go strikes again, interesting.

Makefile
rquitales, line 16: Should we maintain consistency with `TESTPARALLELISM`, where we define the value only, eg `2` instead of `-p 2`?

rquitales (changes requested):

Makefile
rquitales: NVM. Just refferred to https://github.com/pulumi/ci-mgmt/pull/386 and including the flag within the variable makes it easier to unset the flag if required.

rquitales (approved):
LGTM

Makefile
t0yv0: This is something I'd like in my .bashrc hence the longer name PULUMI_PROVIDER_BUILD_PARALLELISM

t0yv0:
b/sdk/go/aws/cognito/identityPoolRoleAttachment.go  is flaky. rerunning fixed it.

PR #2476, opened 2023-04-20
Upgrade terraform-provider-aws to v4.63.0

mikhailshilkov:


kpitzen:
Is the breaking change here expected?

mikhailshilkov:
> Is the breaking change here expected?

They silently did it in https://github.com/hashicorp/terraform-provider-aws/pull/30533 but it makes sense to me that `arn` is an output-only property (computed in TF terms)

kpitzen (approved):

PR #2478, opened 2023-04-20
Add go build caching for the shim testing job

t0yv0:
Introduce Go caching on the "Run test of provider shim" job to reduce flaky timeouts and reduce runtime from 30min to <2 min total.


thomas11 (approved):
The repetition is painful but obviously unrelated to this PR.

PR #2485, opened 2023-04-24
Skip flaky example

t0yv0:
AWS Cognito Identity Pool Roles Attachment example has flaky rendering to Pulumi languages causing unstable builds. Removing the example should make CI builds deterministic until such time as we can fix the root cause of this.


danielrbradley (approved):
LGTM

PR #2501, opened 2023-04-28
Update pulumi-terraform-bridge to v3.45.3

rquitales:
Fixes:

- https://github.com/pulumi/pulumi-terraform-bridge/issues/1044


t0yv0 (approved):
Thanks a ton!

Issue #995, opened 2020-06-05
CallbackFunction should accept Input and not just prompt strings


leezen:
https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/lambda/lambdaMixins.ts#L180

https://pulumi-community.slack.com/archives/CRH5ENVDX/p1591352061275400

The way it is, there's no easy way to use an a policy's ARN `Output` within the same program

lukehoban:
The key reason we haven't done this is the requirements of the code below which (a) needs to be able to create a resource per policy - meaning the array part must be prompt and (b) needs to create a unique name for each attachment derived from the policy so needs prompt access to the policy contents.  Not clear how we would best loosen these requirements while maintaining correctness here.

```ts
            for (const policy of args.policies) {
                // RolePolicyAttachment objects don't have a physical identity, and create/deletes are processed
                // structurally based on the `role` and `policyArn`.  So we need to make sure our Pulumi name matches the
                // structural identity by using a name that includes the role name and policyArn.
                const attachment = new iam.RolePolicyAttachment(`${name}-${utils.sha1hash(policy)}`, {
                    role: role,
                    policyArn: policy,
                }, opts);
            }
```

amichel:
I use following in few places:
```ts
 policyArns.map(policy => {
            let p = pulumi.interpolate`${policy}`;
            p.apply(arn =>
                getResourceFromPolicyArn(arn).apply(resource =>
                    new aws.iam.RolePolicyAttachment(`RoleAttachment-${roleName}-${resource}}`, { role: role, policyArn: policy })
                ))
        });

function getResourceFromPolicyArn(arn: string) {
    return pulumi.output(getArn({ arn: arn })).resource.apply(r => { return r.split("/").pop() });
}
```

lukehoban:
> I use following in few places

This approach results in creating `RolePolicyAttachment` resources inside an `apply`, which means that the preview will not in general be accurate - when the number of ARNs or any of the individual ARNs are not yet known (because they are dependent on the output of other resource), Pulumi will not be able to report whether or how many `RolePolicyAttachment` resources will be created.  

For this built-in component, we've aimed to ensure that the preview can be completely reliable, which today means that the resource. creations must not be inside `apply`s.

lukehoban:
Going to close this out as a dupe of https://github.com/pulumi/pulumi/issues/4834#issuecomment-644827733.

[...]

PR #2505, opened 2023-05-01
Nodejs lambda callbackfunction lifted policies

gsuess:
On the NodeJS SDK, `lambda.CallbackFunction` has a handy property for directly attaching policies.  However practically only managed policies used to be supported, because policies provisioned in the same pulumi stack would always have a lifted ARN.

This constraint was in place, because the ARN is used to generate a unique URN for the policy attachment and URNs cannot be lifted.

In this PR I have addressed the problem, allowing user-supplied unique labels for such policies, while still maintaining backward compatibility to avoid any breaking changes.

### DIFF
--- a/examples/callbackfunction/index.ts
+++ b/examples/callbackfunction/index.ts
@@ -22,6 +22,17 @@ const role = new aws.iam.Role("role", {
     assumeRolePolicy: aws.iam.assumeRolePolicyForPrincipal({ Service: "lambda.amazonaws.com" }),
 }, providerOpts);
 
+const policy = new aws.iam.Policy("policy", {
+    policy: {
+        Version: "2012-10-17",
+        Statement: [{
+            Action: ["s3:ListBucket"],
+            Effect: "Allow",
+            Resource: "arn:aws:s3:::*",
+        }],
+    }
+});
+
 const functions = [
     new aws.lambda.CallbackFunction("a", {
         callback: async () => ({ success: true }),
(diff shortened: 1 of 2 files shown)


lukehoban (approved):
One note related to the documentation for the `policies` property - but otherwise this looks great!

lukehoban:
BTW - I believe this actually provides a very good way of resolving https://github.com/pulumi/pulumi-aws/issues/995 (and the couple of issues that were duped against it).  We can likely re-close that as fixed when this lands.

gsuess:
I looked into the failed CI, am I correct wit that this is an unrelated credential setup problem, rather than some breaking change?

thomas11:
> I looked into the failed CI, am I correct wit that this is an unrelated credential setup problem, rather than some breaking change?

@gsuess I'm not sure it's unrelated. `TestAccCallbackFunction` is the only test using `getAwsSession` from `examples_nodejs_test.go`. Maybe your change uncovered an issue with the setup code there. The test passed in the last CI run for the master branch.

I'll try to take a deeper look later.

gsuess:
@thomas11 any progress on this? Is there anything I can do to help getting this merged?

thomas11:
Hi @gsuess , sorry for the delay. #2545 fixes one of the two test failures which is unrelated to your PR.

I'm a bit baffled by the other test failure. It doesn't seem related to your PR but the test passes in `master`. I'm going to try some experiments and maybe ask a colleague.

iwahbe:
@gsuess This merged as part of https://github.com/pulumi/pulumi-aws/pull/2553. Thanks for your contribution!

PR #2541, opened 2023-06-04
Change references to `filename` to `code`

duongdominhchau:
On Terraform docs there is an option named `filename`, but there is none on Pulumi docs, and the `code` option is mentioned as conflict with `s3_*` options, so I think `code` is the correct one here, not `filename`.


duongdominhchau:
Ouch, looks like I modified generated files, closing this PR then, sorry.

Issue #2547, opened 2023-06-06
Upgrade bridge to latest version


mnlumi:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Upgrade bridge to latest version

t0yv0:
https://github.com/pulumi/pulumi-terraform-bridge/releases/tag/v3.49.0 would be most helpful.

PR #2556, opened 2023-06-09
Update pulumi-terraform-bridge to v3.49.1

t0yv0:
Fixes https://github.com/pulumi/pulumi-aws/issues/2547

Figuring out the bridge and pulumi/pkg update and verifying the diffs *before* doing further surgery can be very beneficial. The V6 PR may need to be built on top of something like this. PTAL.


rquitales:
This looks good to me. Is there a reason we have a schema change with the missing property from just the bridge upgrade though?

t0yv0:
That doesn't make sense to me, need to dig in.

t0yv0:
Here is the definition:

https://github.com/pulumi/terraform-provider-aws/blob/a82c36c5556e2b8de91ba2a82a24bcd6074345be/internal/service/ecs/service.go#L383

Here is the override:

https://github.com/pulumi/pulumi-aws/blob/f854197a6ec1a3479edbbaba61b304ee84a26270/provider/resources.go#L1997

A quick derivation:

```bash
cd ~/code/pulumi-aws
jq '.resources["aws:ecs/service:Service"].properties.serviceConnectConfiguration' $(find . -name 'schema.json')
jq '.types["aws:ecs/ServiceServiceConnectConfiguration:ServiceServiceConnectConfiguration"].properties.services' $(find . -name 'schema.json')
jq '.types["aws:ecs/ServiceServiceConnectConfigurationService:ServiceServiceConnectConfigurationService"].properties.clientAliases' $(find . -name 'schema.json')


{
  "$ref": "#/types/aws:ecs/ServiceServiceConnectConfiguration:ServiceServiceConnectConfiguration",
  "description": "The ECS Service Connect configuration for this service to discover and connect to services, and be discovered by, and connected from, other services within a namespace. See below.\n"
}
{
  "type": "array",
  "items": {
    "$ref": "#/types/aws:ecs/ServiceServiceConnectConfigurationService:ServiceServiceConnectConfigurationService"
  },
  "description": "The list of Service Connect service objects. See below.\n"
}
{
  "type": "array",
  "items": {
    "$ref": "#/types/aws:ecs/ServiceServiceConnectConfigurationServiceClientAlias:ServiceServiceConnectConfigurationServiceClientAlias"
  },
  "description": "The list of client aliases for this Service Connect service. You use these to assign names that can be used by client applications. The maximum number of client aliases that you can have in this list is 1. See below.\n"
}
```

This field is now cientAliases instead of clientAlias. This seems to be correct - it's defined as MaxItems=1 in the schema but it's overridden in resources.go to ignore MaxItems=1. Ignoring it makes it a list, which also means the name is pluralized in Pulumi.

t0yv0:
before the change this field was an array but called clientAlias

```
    "clientAlias": {
      "type": "array",
      "items": {
        "$ref": "#/types/aws:ecs/ServiceServiceConnectConfigurationServiceClientAlias:ServiceServiceConnectConfigurationServiceClientAlias"
      },
      "description": "The list of client aliases for this Service Connect service. You use these to assign names that can be used by client applications. The maximum number of client aliases that you can have in this list is 1. See below.\n"
    },
```

t0yv0:
I think the change is https://github.com/pulumi/pulumi-terraform-bridge/releases/tag/v3.47.1 specifically https://github.com/pulumi/pulumi-terraform-bridge/pull/1117

@rquitales how does it sound? Should we keep this change as is, or should we restore the previous name by manually specifying the name to be "clientAlias"?

rquitales:
@t0yv0 Thanks for digging! Can we create an alias for that field name change as well to maintain forwards compatibility for existing users? Creating a pluralized name for that field makes sense to me.

t0yv0:
How do you create an alias for a field? Something to set on SchemaInfo? Are there some examples? Thanks.

t0yv0:
Update pulumi-terraform-bridge to v3.49.1

Issue #2572, opened 2023-06-23
After tagging v6-alpha main branch fails to build


t0yv0:
### What happened?

After creating v6-alpha tags, main branch builds are affected as they are confused about the major version. `pulumictl` now thinks that master builds are on 6.0.0 whereas other parts of the code are assuming a v5 major version. This causes failures like https://pulumi.slack.com/archives/C02FXTZEZ6W/p1687524638095169?thread_ts=1687492809.750579&channel=C02FXTZEZ6W&message_ts=1687524638.095169

I can reproduce by calling `make tfgen` on master and observing a v6 version being linked into the binary.

### Expected Behavior

main branch builds correctly on the v5 version

### Steps to reproduce

run make tfgen provider on master and observe non-empty diff

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

rquitales:
Closing this issue as this is resolved by removing the `v6.0.0-alpha.0` tag which was mistakenly created.

t0yv0:
@rquitales resolved it by removing the -alpha tag that accidentally marked the master branch tip commit as 6.x.x.

PR #2573, opened 2023-06-23
Pin VERSION_PREFIX=5.0.0

t0yv0:
This is a workaround to make master builds succeed until pulumi-aws v6.0.0 is merged to master and released.

Fixes https://github.com/pulumi/pulumi-aws/issues/2572

```
git grep -l 'pulumictl get version' | 
    xargs sed -i '' -e 's/pulumictl get version/VERSION_PREFIX=5.0.0 pulumictl get version/g'                                            
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,7 +88,7 @@ jobs:
     - name: Update path
       run: echo "${{ github.workspace }}/bin" >> $GITHUB_PATH
     - name: Set PACKAGE_VERSION to Env
-      run: echo "PACKAGE_VERSION=$(pulumictl get version --language generic)" >>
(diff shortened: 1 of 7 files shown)


t0yv0:
Actually the problem is that https://github.com/pulumi/pulumi-aws/tree/v6.0.0-alpha.0 tags a commit on master when the major upgrade was not performed yet. This is not good. @rquitales instead of taking this change here we could consider dropping the tag from the repository... do you think that's acceptable? It seems like it was tagged in error.

t0yv0:
Good, let's try it and see if that makes the master builds succeed. If so we can close this PR.

t0yv0:
Not needed anymore, TYVM!

PR #2574, opened 2023-06-23
Remove QuickSpace related TODOs

t0yv0:
The TODOs were around generated code breaking C# compilation rules due to a conflict between class and property name. This is worked around by renaming the property names for C# only using a CSharpName tweak in resources.go.


rquitales:
@t0yv0 Is it expected for us to have this many new schema changes?

t0yv0:
@rquitales I don't thin the report is accurate, it's not computing the schema changes from the base of this PR but instead is computing them from master. 

on v6 branch this yields a hit:

```
jq '.functions["aws:auditmanager/getControl:getControl"]' $(find . -name '*schema.json') 
```

The only expected changes are 2 new resources that are not skipped anymore.

rquitales (approved):

Issue #2549, opened 2023-06-06
Update implementation method of backported resources for v6


mnlumi:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Update implementation method of backported resources for v6

The majority of this work is just renaming the `_reverted` and removing the corresponding backported files. We will also need to track down backported resources which do not follow this pattern, since it was only adopted relatively recently. A decent starting point is just pulling a diff between our fork and the upstream provider, and looking for diffs where ours is using the SDK, and theirs is not.

t0yv0:
We believe  https://github.com/pulumi/pulumi-aws/issues/2533 might be fixes as part of but will need to verify.

mikhailshilkov:
Fixed by #2578

PR #2578, opened 2023-06-27
Use PF for backported resources for v6

rquitales:
This updates `pulumi-aws` to point to upstream `patched-v5.1.0-alpha.2` which removes reverted resources. The resources were compared against the latest upstream to ensure all reverted resources have been switched to use the plugin framework.

Most changes of interest were done to the `patched-v5.1.0-alpha.2` branch.

Now pinned to `patched-v5.1.0-alpha.3`

Fixes: #2549


t0yv0:
Difficult to review but looks good at a glance. bridge-metadata.json shows resources and functions moving from 0 (sdkv2) to pf (1).

t0yv0:
Perhaps walk us through this today we can figure out together?

rquitales:
@kpitzen Yes, that is the right branch to look at. I dropped all the revert framework patches/commits we have, so that should be reflected in the `upstream-patches` folder. That got rid of most reversions done. Went through all the other patches after to clean up/revert the other resources.

mikhailshilkov (approved):

danielrbradley (approved):

PR #2583, opened 2023-06-28
 Update to patched-v5.5.0-alpha.1

rquitales:


iwahbe:
I believe we can close this, given we just merged https://github.com/pulumi/pulumi-aws/pull/2605.

PR #2584, opened 2023-06-29
Remove aws.sdk property

### LINKED ISSUE #2568: `aws.sdk` usage raises "NOTE: We are formalizing our plans to enter AWS SDK for JavaScript (v2) into maintenance mode in 2023."
Any program which uses `aws.sdk` sees this warning:

```
    (node:61320) NOTE: We are formalizing our plans to enter AWS SDK for JavaScript (v2) into maintenance mode in 2023.
    Please migrate your code to use AWS SDK for JavaScript (v3).
    For more information, check the migration guide at https://a.co/7PzMCcy
    (Use `node --trace-warnings ...` to show where the warning was created)
```

This is because `@pulumi/aws` uses `"aws-sdk": "^2.0.0",`.  

We likely need to move to exposues `^3.0.0`, though that would be a breaking change.  We might want to include this change into v6?

lukehoban:
Note: This needs to be rebased against the `v6` branch.

The `aws.sdk` property was introduced several years ago to offer a convenient way for callbacks using function serialization to get ahold of the AWS SDK which is available within the Lambda runtime.

However, the AWS SDK v2 that was exposed here is being deprecated, and has been emitted a deprecation warning for some time for anyone using this via Pulumi.

The AWS SDK v3 doesn't expose a monolithic module we can expose here - providing instead many separate SDKs for each service.

So we expect the best path is to drop this from the AWS provider in the upcoming major version release.

Users who were using it in <6.0.0 versions will need to move to using `@aws-sdk/client-s3` or similar libraries directly.

Fixes #2568


mikhailshilkov (commented):
This LGTM. Great to have all the extra tests and examples updated. Do you want to rebase it on v6 or should I? Anything else remaining here?

iwahbe (approved):

mikhailshilkov (approved):

iwahbe:
Hi @catmeme. Sorry for the confusion. We called it out in the [migration guide](https://www.pulumi.com/registry/packages/aws/how-to-guides/6-0-migration/). I see that the migration guide isn't linked super clearly in the blog post. I'm going to see if we can update the blog post to make it clearer where the migration guide is.

PR #2592, opened 2023-06-30
Fix non-deterministic build

AaronFriel:


kpitzen (approved):

AaronFriel:
All checks are green, but GitHub branch protection is reporting a number of jobs "pending", each of which appears green here: https://github.com/pulumi/pulumi-aws/pull/2592/checks

PR #2611, opened 2023-07-12
Remove aliases and max items one patches

### LINKED ISSUE #2610: Remove Aliases and MaxItemsOne fixes
We should remove all resource and datasource aliases as well as `MaxItemsOne` pinning in preparation for the next major version. 

In future updates, this will be handled by `ProviderInfo.ApplyAutoAliasing`.

iwahbe:
We are taking the opportunity of the major upgrade to remove legacy artifacts like aliases.

Fixes #2610


iwahbe:
Closing in favor of https://github.com/pulumi/pulumi-aws/pull/2615.

PR #2618, opened 2023-07-17
Ensure that bin/pulumi-java-gen is up to date

### LINKED ISSUE #2297: `make build_java` does not download newer versions of pulumi-java-gen
Building the Java SDK correctly depends on `bin/pulumi-java-gen`, but since the version `JAVA_GEN_VERSION` is not used in the Makefile target, it will only download a current binary when none exists locally. When one exists but is outdated, make will do nothing.

We should explicitly store the version number as a file so Make can detect changes.

iwahbe:
This change checks that the downloaded version of `pulumi-java-gen` matches the desired version.

Fixes #2297


thomas11 (approved):
thanks!

iwahbe:
Hanging comments look like a GitHub bug. Tests have passed, so I'm going to manually merge.

Issue #2639, opened 2023-07-25
v6 is incompatible with `@pulumi/aws` and `@pulumi/aws-apigateway`


lukehoban:
It appears that some of the breaking changes in v6 lead to very unfortunate incompatibility with some of our most popular libraries.

For example - updating https://github.com/pulumi/examples/blob/master/aws-apigateway-ts-routes/index.ts to use `@pulumi/aws` v6 leads to errors like the following:

```
Type 'CallbackFunction<unknown, { statusCode: number; body: string; }>' is not assignable to type 'Input<Function> | undefined'.
  Type 'CallbackFunction<unknown, { statusCode: number; body: string; }>' is not assignable to type 'Function'.
    Types of property 'roleInstance' are incompatible.
      Type 'import("/Users/lukehoban/go/src/github.com/pulumi/examples/aws-apigateway-ts-routes/node_modules/@pulumi/aws/iam/role").Role | undefined' is not assignable to type 'import("/Users/lukehoban/go/src/github.com/pulumi/examples/aws-apigateway-ts-routes/node_modules/@pulumi/aws-apigateway/node_modules/@pulumi/aws/iam/role").Role | undefined'.
        Property 'roleLastUseds' is missing in type 'import("/Users/lukehoban/go/src/github.com/pulumi/examples/aws-apigateway-ts-routes/node_modules/@pulumi/aws/iam/role").Role' but required in type 'import("/Users/lukehoban/go/src/github.com/pulumi/examples/aws-apigateway-ts-routes/node_modules/@pulumi/aws-apigateway/node_modules/@pulumi/aws/iam/role").Role'.
```

That is - the v6 instance of `aws.lambda.Function` can not be passed to the MLCs expected v5 version of `aws.lambda.Function`.

This appears to be because of a confluence of two things:
1. The `roleLastUseds` property of `aws.iam.Role` was removed by upstream as a breaking change: https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md#510-june--1-2023
2. The `aws.lambda.Function` type has, by a weird mechanism, a `roleInstance: aws.iam.Role` as a property.

This prevents passing a v6 `Function` to something expecting a v5 `Function`.

I think there are two potential fixes:
[...]

PR #2645, opened 2023-07-26
Move field `roleInstance` from `Function` to `CallbackFunction`

iwahbe:
Fixes https://github.com/pulumi/pulumi-aws/issues/2639

This actually makes the mixin more self-contained, since we previously added a field to a normal pulumi resource, making TS's `lambda.Function` different then Go's `lambda.Function`. With this change, `lambda.Function` is the same in all languages.

--- 
https://github.com/pulumi/pulumi-aws/pull/2645/commits/2c09c53a812078047021ad9eb01d54218ed52b01: `serverless.Function` has been deprecated in favor of `lambda.CallbackFunction` since 2018 (https://github.com/pulumi/pulumi-aws/pull/327). Since it depends on the removed functionality, I have bundled its removal.


thomas11 (commented):
Looks like some tests need to be updated after the change. How disruptive will this be for existing `Function` users? I know we can make breaking changes for v6 but is it an easy source code update?

iwahbe:
> Looks like some tests need to be updated after the change. How disruptive will this be for existing `Function` users? I know we can make breaking changes for v6 but is it an easy source code update?

It should be an easy update. `roleInstance` was always `undefined` on all `lambda.Function`s except `lambda.CallbackFunction`, so the user is not loosing access to any data. Because this is TS with progressive typing, even if you had a collection of `lambda.Function`, you could always check ` if ('roleInstance' in f) { // then...` to recover old information.

thomas11 (approved):

PR #2649, opened 2023-07-27
Refactor artifact cleanup

### LINKED ISSUE #2603: Workflow failure: cleanup
## Workflow Failure

[cleanup](https://github.com/pulumi/pulumi-aws/blob/master/.github/workflows/artifact-cleanup.yml) has failed. See the list of failures below:

- [2023-07-07T01:33:11.000Z](https://github.com/pulumi/pulumi-aws/actions/runs/5481782417)

- [2023-07-08T01:33:05.000Z](https://github.com/pulumi/pulumi-aws/actions/runs/5491952285)

- [2023-07-09T01:34:50.000Z](https://github.com/pulumi/pulumi-aws/actions/runs/5497356249)

- [2023-07-10T01:34:14.000Z](https://github.com/pulumi/pulumi-aws/actions/runs/5503074854)

- [2023-07-11T01:34:14.000Z](https://github.com/pulumi/pulumi-aws/actions/runs/5514784896)

- [2023-07-12T01:34:12.000Z](https://github.com/pulumi/pulumi-aws/actions/runs/5526582201)

- [2023-07-13T01:34:19.000Z](https://github.com/pulumi/pulumi-aws/actions/runs/5538102357)

- [2023-07-14T01:35:30.000Z](https://github.com/pulumi/pulumi-aws/actions/runs/5549528392)

- [2023-07-21T01:31:46.000Z](https://github.com/pulumi/pulumi-aws/actions/runs/5617601098)

- [2023-07-24T01:31:46.000Z](https://github.com/pulumi/pulumi-aws/actions/runs/5639795426)

- [2023-07-30T01:31:07.000Z](https://github.com/pulumi/pulumi-aws/actions/runs/5703559554)

guineveresaenger:
Fixes #2603

Co-authored by: Alex Qiu <aqiu@pulumi.com>


.github/workflows/release.yml
```diff
@@ -103,6 +103,7 @@ jobs:
       with:
         name: ${{ matrix.language  }}-sdk.tar.gz
         path: ${{ github.workspace}}/sdk/${{ matrix.language }}.tar.gz
+      retention-days: 30
```
thomas11, line 106: There's no equivalent of the previous `skip-tags: true` here. Doesn't that mean we'll lose release artifacts after 30 days?

guineveresaenger: Yes, it would! Sounds like I was too eager here, and we should not add this to the release or prerelease workflows, using the default instead.

mikhailshilkov: I think I saw you mentioned somewhere that 30 is an override of the default value which is 90. Don't we have the same problem after 90 days currently?

guineveresaenger: @mikhailshilkov: This is the current status quo as far as I'm aware. [Here is a release run of 2 months ago with existing artifacts](https://github.com/pulumi/pulumi-aws/actions/runs/4898369557). Compare it to [this release of 3 months ago](https://github.com/pulumi/pulumi-aws/actions/runs/4823795386) - the artifacts are `Expired`.
If we want to keep them around for longer, we would have to come up with a new idea, but I'm unaware of any need for that.

thomas11 (approved):

Issue #1655, opened 2021-10-08
unexpected behavior of defaultTags


filip-zyzniewski:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

The AWS provider ignores changes to defaultTags after a resource has been created unless other attributes are being changed

### Steps to reproduce

1. Specify some defaultTags and create a bucket. The bucket has the tags as expected:

```
~/tmp/wtf/defaultTags$ cat index.ts
import * as aws from "@pulumi/aws";

const bucket = new aws.s3.Bucket("bucket-with-default-tags", {bucket: "bucket-with-default-tags"});
~/tmp/wtf/defaultTags$ cat Pulumi.dev.yaml
encryptionsalt: v1:ggyom6IhdG8=:v1:sJqljlbglc8JJgpe:jYzdmym/+StNPwPbn+7uIpAIL+fKLA==
config:
  aws:region: eu-west-1
  aws:defaultTags:
    tags:
      Tag1: value1
      Tag2: value2
~/tmp/wtf/defaultTags$ pulumi up --diff
Previewing update (dev):
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev::defaultTags::pulumi:pulumi:Stack::defaultTags-dev]
    + aws:s3/bucket:Bucket: (create)
        [urn=urn:pulumi:dev::defaultTags::aws:s3/bucket:Bucket::bucket-with-default-tags]
        [provider=urn:pulumi:dev::defaultTags::pulumi:providers:aws::default_4_23_0::04da6b54-80e4-46f7-96ec-b56ff0331ba9]
        acl         : "private"
        bucket      : "bucket-with-default-tags"
        forceDestroy: false

import * as aws from "@pulumi/aws";
Resources:
    + 1 to create
    1 unchanged
Do you want to perform this update? yes
Updating (dev):
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev::defaultTags::pulumi:pulumi:Stack::defaultTags-dev]
    + aws:s3/bucket:Bucket: (create)
        [urn=urn:pulumi:dev::defaultTags::aws:s3/bucket:Bucket::bucket-with-default-tags]
        [provider=urn:pulumi:dev::defaultTags::pulumi:providers:aws::default_4_23_0::8510e26a-009d-460f-a5e1-4f5e95a6af22]
[...]

PR #2655, opened 2023-07-31
Compute `"tags"` in Check

### LINKED ISSUE #2550: Test and validate changes related to default tagging behavior
## Hello!

## Issue details
Test and validate changes related to default tagging behavior
We should verify whether #2549 _actually_ closes out #2533 - ideally by adding a programtest to reproduce the issue, and validating that it passes once the reverted SDK resource is removed.

### LINKED ISSUE #2633: v6 default tags test failing in CI
The final step (5) in the test `TestAccDefaultTagsGo` is currently failing in CI but passes locally.

We're disabling the test temporarily to unblock the release but should diagnose the failure as a matter of urgency post-release.

Step 5 is asserting that when all default tags are removed from the provider, then they're also removed from all resources. This fails and the previous default tags remain on the resources.

iwahbe:
Fixes #2633
Fixes #2550

Relies on https://github.com/pulumi/pulumi-terraform-bridge/pull/1310.

Description copied from `applyTags`'s doc comment:

Historically, Pulumi has struggles to handle the "tags" and "tags_all" fields correctly:
- https://github.com/pulumi/pulumi-aws/issues/2633
- https://github.com/pulumi/pulumi-aws/issues/1655

terraform-provider-aws has also struggled with implementing their desired behavior:
- https://github.com/hashicorp/terraform-provider-aws/issues/29747
- https://github.com/hashicorp/terraform-provider-aws/issues/29842
- https://github.com/hashicorp/terraform-provider-aws/issues/24449

The Terraform lifecycle simply does not have a good way to map provider configuration onto resource values, so terraform-provider-aws is forced to work around limitations in unreliable ways. For example, terraform-provider-aws does not apply tags correctly with -refresh=false.

This gives pulumi the same limitations by default. However, unlike Terraform, Pulumi does have a clear way to insert provider configuration into resource properties: Check. By writing a custom check function that applies "default_tags" to "tags" before the Terraform provider sees any resource configuration, we can give a consistent, reliable and good experience for Pulumi users.

---

This strategy gives concrete benefits for Pulumi users. Diffs are applied correctly with high resolution. Consider this change:

```patch
name: our-example
runtime: yaml
resources:
  p:
    type: pulumi:providers:aws
    properties:
      defaultTags:
        tags:
          foo: buzz
          abc: def
      region: us-west-2
  b:
    type: aws:s3:BucketV2
-   properties:
-     tags:
-       foo: fizz
    options:
      provider: ${p}
```

We now can give users an accurate diff:

```
𝛌 pulumi preview --diff
Previewing update (test)

View Live: https://app.pulumi.com/pulumi/our-example/test/previews/*

  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:test::our-example::pulumi:pulumi:Stack::our-example-test]
warning: provider config warning: skip_metadata_api_check: the use of values other than "true" and "false" is deprecated and will be removed in a future version of the provider
    ~ aws:s3/bucketV2:BucketV2: (update)
        [id=b-f4d2684]
        [urn=urn:pulumi:test::our-example::aws:s3/bucketV2:BucketV2::b]
        [provider=urn:pulumi:test::our-example::pulumi:providers:aws::p::aaefc0bd-e75f-47b8-9d61-dc8c5cba05b0]
      ~ tags: {
          ~ foo: "fizz" => "buzz"
        }
Resources:
    ~ 1 to update
    2 unchanged
```

Changing `defaultTags` in a way that doesn't effect a resource no longer shows as an update for that resource, just for the provider:

```patch
name: our-example
runtime: yaml
resources:
  p:
    type: pulumi:providers:aws
    properties:
      defaultTags:
        tags:
-         foo: buzz
          abc: def
      region: us-west-2
  b:
    type: aws:s3:BucketV2
    properties:
      tags:
        foo: fizz
    options:
      provider: ${p}
```

```
𝛌 pulumi preview --diff
Previewing update (test)

View Live: https://app.pulumi.com/pulumi/our-example/test/previews/0*

  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:test::our-example::pulumi:pulumi:Stack::our-example-test]
warning: provider config warning: skip_metadata_api_check: the use of values other than "true" and "false" is deprecated and will be removed in a future version of the provider
    ~ pulumi:providers:aws: (update)
        [id=aaefc0bd-e75f-47b8-9d61-dc8c5cba05b0]
        [urn=urn:pulumi:our-example::pulumi:providers:aws::p]
      ~ defaultTags              : "\"{\\\"tags\\\":{\\\"abc\\\":\\\"def\\\",\\\"foo\\\":\\\"buzz\\\"}}\"" => "\"{\\\"tags\\\":{\\\"abc\\\":\\\"def\\\"}}\""
        region                   : "us-west-2"
        skipCredentialsValidation: "false"
        skipMetadataApiCheck     : "true"
        skipRegionValidation     : "true"
Resources:              
    ~ 1 to update
    2 unchanged
```

Of course, this also fixes the original issue; removing all `defaultTags` correctly removes them from the resource.

```patch
name: our-example
runtime: yaml
resources:
  p:
    type: pulumi:providers:aws
    properties:
-     defaultTags:
-       tags:
-         foo: buzz
-         abc: def
      region: us-west-2
  b:
    type: aws:s3:BucketV2
    options:
      provider: ${p}
```

```
𝛌 pulumi preview --diff
Previewing update (test)

View Live: https://app.pulumi.com/pulumi/our-example/test/previews/*

  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:test::our-example::pulumi:pulumi:Stack::our-example-test]
warning: provider config warning: skip_metadata_api_check: the use of values other than "true" and "false" is deprecated and will be removed in a future version of the provider
    ~ pulumi:providers:aws: (update)
        [id=aaefc0bd-e75f-47b8-9d61-dc8c5cba05b0]
        [urn=urn:pulumi:test::our-example::pulumi:providers:aws::p]
      - defaultTags              : (json) {
          - tags: {
              - abc: "def"
              - foo: "buzz"
            }
        }

        region                   : "us-west-2"
        skipCredentialsValidation: "false"
        skipMetadataApiCheck     : "true"
        skipRegionValidation     : "true"
    ~ aws:s3/bucketV2:BucketV2: (update)
        [id=b-f4d2684]
        [urn=urn:pulumi:test::our-example::aws:s3/bucketV2:BucketV2::b]
        [provider=urn:pulumi:test::our-example::pulumi:providers:aws::p::aaefc0bd-e75f-47b8-9d61-dc8c5cba05b0]
      - tags: {
          - abc: "def"
          - foo: "buzz"
        }
    --outputs:--        
  ~ tags: {
      - abc: "def"
      - foo: "buzz"
    }
Resources:
    ~ 2 to update
    1 unchanged
```

--- 

A bonus for maintainers is that because this change takes effect at the pulumi level, we no longer rely on changing how our involved diff method works, allowing us to remove `XComputedInput` from this provider (and then from the bridge).


iwahbe:
This will show diffs on `tags` when upgrading from v5 to v6, so it must be introduced as part of the major version upgrade and called out in the changelog.

iwahbe:
Currently, this works for:
- [X] Create
- [X] Check (by construction)
- [X] Diff
- [X] Update
- [X] Destroy (noop)
- [ ] Read

Read is the only CRUD+ method that does not go through Check, so it will also need special behavior so that `pulumi up --yes && pulumi refresh --expect-no-changes` doesn't show a diff.

t0yv0:
> For example, terraform-provider-aws does not apply tags correctly with -refresh=false.

I was not able to isolate this, do you have a repro? Can we file a bug upstream with a repro?

t0yv0:
What happens if a provider with this set of changes runs against stacks that used the old method and perhaps populated tags_all?

t0yv0:
Overall I appreciate this approach, if we cannot find a clean way to interface with the functionality in upstream provider, maintaining custom code around default tags seems worth it as it is a feature impacting every resource; if we can do this cleanly. There is a cost of doing so in maintaining this new code and ensuring that it does not interact in weird ways with the upstream machinery that attempts to solve the problem as well.

iwahbe:
> How can we demonstrate that import and refresh scenarios exercising Read work as expected?

By adding a test that exercises the behavior. We can't do that yet, since any such test will fail until https://github.com/pulumi/pulumi-terraform-bridge/pull/1312 merges.

iwahbe:
> What happens if a provider with this set of changes runs against stacks that used the old method and perhaps populated tags_all?

They will see a diff on `"tags"` (see https://github.com/pulumi/pulumi-aws/pull/2655#issuecomment-1658204832), but otherwise everything will work as expected. This doesn't change anything about `"tags_all"`, we won't see a problem here.

t0yv0:
>  We can't do that yet, since any such test will fail until

Ah super, you have a way to demonstrate in a test, that's great, let's try it. I think you can make the CI exercise pre-release code via go mod-replace. Alternatively, I had some nits on https://github.com/pulumi/pulumi-terraform-bridge/pull/1312 but it's also mergeable once we add a quick test in there.

examples/default-tags-yaml/Pulumi.yaml
```diff
@@ -10,4 +10,4 @@ resources:
   my-bucket:
     type: aws:s3:Bucket
 outputs:
-  bucketTags: ${my-bucket.tagsAll}
+  bucketTags: ${my-bucket.tags}
```
danielrbradley, line 13: What's the reasoning to switch `tagsAll` to `tags` in the test? Does `tagsAll` still work too?

danielrbradley (approved):
Overall happy - would be good to get more resources covered as discussed, but the implementation looks good to me.

examples/default-tags-yaml/Pulumi.yaml
iwahbe: Because we are the ones setting `tags`. `tagsAll` is managed by TF (derived from `tags`). They are equivalent. We can tests for both.

provider/resources.go
```diff
@@ -7122,3 +7113,89 @@ func Provider() *tfbridge.ProviderInfo {
...
+	ctx context.Context, config resource.PropertyMap, meta resource.PropertyMap,
+) (resource.PropertyMap, error) {
+	var defaultTags awsShim.TagConfig
+
+	unknown := func() (resource.PropertyMap, error) {
+		current := config["tags"]
+		if current.IsOutput() {
+			output := current.OutputValue()
+			output.Known = false
+			config["tags"] = resource.NewOutputProperty(output)
```
t0yv0, line 7148: This mutates config besides returning it..

provider/resources.go
```diff
@@ -7122,3 +7113,89 @@ func Provider() *tfbridge.ProviderInfo {
...
+
+	if defaults, ok := meta["defaultTags"]; ok {
+		if defaults.ContainsUnknowns() {
+			return unknown()
+		}
+		if defaults.IsObject() {
+			defaults := defaults.ObjectValue()
+			tags, ok := defaults["tags"]
+			if ok {
+				defaultTags = awsShim.NewTagConfig(ctx, tags.Mappable())
```
t0yv0, line 7163: Eh. This awsShim.NewTagConfig works on TF "any" domain but tags.Mappable() is a Pulumi "any" domain, perhaps a comment here that we expect representations to match?

provider/tests/provider_test.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+// Repro https://github.com/pulumi/pulumi-aws/issues/2633
+func TestRegression2633(t *testing.T) {
+	t.Parallel()
+	t.Run("changed tags", func(t *testing.T) { replaySequence(t, regress2633(true)) })
+	t.Run("removed tags", func(t *testing.T) { replaySequence(t, regress2633(false)) })
+}
+
+func regress2633(hasTags bool) string {
```
t0yv0, line 164: Can you walk me through this? I don't understand. Suppose the scenario is hasTags=true.

The provider is configured with defaultTags having "foo" set.

Why is the diff deciding that "detailedDiff": { "tags.foo": { "kind": "DELETE" } }?

Shouldn't tags.foo be propagated from provider defaults and be stable?

upstream-patches/0029-Match-the-tags-behavior-of-other-resources.patch
```diff
@@ -0,0 +1,48 @@
...
+Subject: [PATCH 29/29] Match the "tags" behavior of other resources
+
+---
+ internal/service/s3legacy/bucket_legacy.go | 14 +++-----------
+ 1 file changed, 3 insertions(+), 11 deletions(-)
+
+diff --git a/internal/service/s3legacy/bucket_legacy.go b/internal/service/s3legacy/bucket_legacy.go
+index 0563bd3a25..e6bd8e7013 100644
+--- a/internal/service/s3legacy/bucket_legacy.go
++++ b/internal/service/s3legacy/bucket_legacy.go
```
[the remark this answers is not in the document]
t0yv0: This is interesting. The patch is still needed it just got shorter?

provider/resources.go
iwahbe: It does. `applyTags` is only called like this:

```go
	if check := res.Schema.PreCheckCallback; check != nil {
		news, err = check(ctx, news, p.configValues.Copy())
		if err != nil {
			return nil, err
		}
	}
```

[definition](https://github.com/pulumi/pulumi-terraform-bridge/blob/86acdc26fa8cd7bb6a081162de0771ee715554e8/pkg/tfbridge/provider.go#L570-L576)

It's ok to mutate. Its an "in/out" parameter in practice, but it's not idiomatic go to write `func(ctx context.Context, config *resource.PropertyMap, meta resource.PropertyMap) error`.

examples/examples_yaml_test.go
```diff
@@ -53,3 +55,66 @@ func getYamlBaseOptions(t *testing.T) integration.ProgramTestOptions {
...
+		EditDirs: []integration.EditDir{
+			getEditDir(1),
+			getEditDir(2),
+			getEditDir(3),
+			getEditDir(4),
+			getEditDir(5),
+		},
+
+		Config:           map[string]string{"aws:region": getEnvRegion(t)},
+		Quick:            true,
```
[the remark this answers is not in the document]
iwahbe: It skips preview but not refresh:

```go
	if opts.Quick {
		opts.SkipPreview = true
		opts.SkipExportImport = true
		opts.SkipEmptyPreviewUpdate = true
	}
```
[definition](https://github.com/pulumi/pulumi/blob/244d4265d16ffedea6db1f65090d9d87652ffca9/pkg/testing/integration/program.go#L760-L765)

I'm writing the more comprehensive test suite for this now. I'll make sure all of this is covered.

provider/tests/provider_test.go
iwahbe: That test is outdated. It is not run by CI. I have removed the test in https://github.com/pulumi/pulumi-aws/pull/2655/commits/cc12424c508051839607eee283419ef11860c7fe.

Issue #2656, opened 2023-08-01
[nodejs sdk] Update aws-sdk to 3


dirien:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

A customer is asking, if and when we can update the `aws-sdk 2.0.0` dependency in the nodejs sdk to version `aws-sdk 3.x.x`

They use the 3.x libary in their code base and it causes trouble for them in some Pulumi operations.

## Possible workaround

I will suggest this [article](https://dev.to/3imed_jaberi/how-to-install-multiple-versions-of-a-npm-package-at-the-same-time-project-2a29#:~:text=I%20found%20a%20solution%20by,package%20in%20the%20same%20project.&text=When%20you%20want%20to%20install,%40%20.) as workaround for the time.

### Affected area/feature

nodejs sdk

dirien:
could me solved with #2584

danielrbradley:
Yup - we should no longer need this dependency at all in v6.

PR #2657, opened 2023-08-01
Remove aws-sdk npm dependency

danielrbradley:
We no longer use this since https://github.com/pulumi/pulumi-aws/pull/2584

Closes https://github.com/pulumi/pulumi-aws/issues/2656


iwahbe (approved):
Thanks!

PR #2675, opened 2023-08-07
Bridge dependency updates

t0yv0:
bridge: v3.56.2
bridge/pf: v0.15.2

Propagates bugfixes, most importantly GetRawConfig() no longer panics.

### DIFF
--- a/provider/go.mod
+++ b/provider/go.mod
@@ -7,8 +7,8 @@ require (
 	github.com/hashicorp/aws-sdk-go-base/v2 v2.0.0-beta.32
 	github.com/hashicorp/terraform-provider-aws/shim v0.0.0
 	github.com/mitchellh/go-homedir v1.1.0
-	github.com/pulumi/pulumi-terraform-bridge/pf v0.15.0
-	github.com/pulumi/pulumi-terraform-bridge/v3 v3.56.0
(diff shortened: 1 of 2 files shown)


provider/go.mod
```diff
@@ -7,8 +7,8 @@ require (
 	github.com/hashicorp/aws-sdk-go-base/v2 v2.0.0-beta.32
 	github.com/hashicorp/terraform-provider-aws/shim v0.0.0
 	github.com/mitchellh/go-homedir v1.1.0
-	github.com/pulumi/pulumi-terraform-bridge/pf v0.15.0
-	github.com/pulumi/pulumi-terraform-bridge/v3 v3.56.0
+	github.com/pulumi/pulumi-terraform-bridge/pf v0.15.1
```
guineveresaenger, line 10: let's use pf/v0.15.2 - it pulls in the bridge changes from 3.56, since I missed that we have to update both.

guineveresaenger (changes requested):

provider/go.mod
t0yv0: Nice, sorry I missed that. Meant to use the latest.

aq17 (approved):

Issue #2678, opened 2023-08-08
Pre-release workflow not publishing releases correctly


t0yv0:
### What happened?

We'd like to be able to cut an -alpha tag and have it usable in the same way as a regular release.  This is not currently working in the GitHub Actions.

First issue we noticed is that the GitHub Release is not being opened; a Release with pre-built binary artifacts needs to be opened so that Pulumi CLI will download the release from GitHub Releases. Also, the pre-built binary artifacts need to be pushed to S3 as a fallback for GitHub Release. None of this was happening.

The first attempt to fix that https://github.com/pulumi/pulumi-aws/pull/2673 edited the GoReleaser configuration.

However this created incorrectly versioned releases on GitHub like this one and needs to be reverted:

https://github.com/pulumi/pulumi-aws/releases/tag/v6.0.0-alpha.1691487285%2Bab1f2fba

Leaving the ticket here to figure out a comprehensive fix for the problem.

### Expected Behavior

Adding a tag like v6.0.0-alpha.8 should create a fully functional release.

### Steps to reproduce

See above.

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #2679, opened 2023-08-08
Revert "Do make releases from -prerelease tags"

t0yv0:
This reverts commit 2692e6edca57a633db3f6a9498d9731621c2405e.

Per https://github.com/pulumi/pulumi-aws/issues/2678 this did not work as expected but created incorrectly versioned releases.


iwahbe (approved):
LGTM

Ok merging with failing tests

PR #2681, opened 2023-08-08
Enable auto-aliasing

iwahbe:
Now that we have a stable release and no longer intend to make backwards incompatible changes, we turn on auto-aliasing. This prevents several classes of breaking changes from effecting our users.

Resolves https://github.com/pulumi/pulumi-aws/issues/2691


iwahbe:
Merged as part of https://github.com/pulumi/pulumi-aws/pull/2766

Issue #2682, opened 2023-08-08
RDS Instance replacement when upgrading to 6.0


mikhailshilkov:
### What happened?

Upgrading from 5.x to 6.0 causes a replacement in `aws.rds.Instance`.

```
     Type                 Name        Plan        Info
     pulumi:pulumi:Stack  aws-ts-dev              
 +-  └─ aws:rds:Instance  postgresdb  replace     [diff: ~name]

Resources:
    +-1 to replace
    2 unchanged

Do you want to perform this update? details
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev::aws-ts::pulumi:pulumi:Stack::aws-ts-dev]
    ++aws:rds/instance:Instance: (create-replacement)
        [id=postgresdbd65b294]
        [urn=urn:pulumi:dev::aws-ts::aws:rds/instance:Instance::postgresdb]
        [provider: urn:pulumi:dev::aws-ts::pulumi:providers:aws::default_5_41_0::5199a248-3c92-4b11-ae4e-dbf860fd2c76 => urn:pulumi:dev::aws-ts::pulumi:providers:aws::default_6_0_0::output<string>]
      ~ name: "airflow" => "airflow"
    +-aws:rds/instance:Instance: (replace)
        [id=postgresdbd65b294]
        [urn=urn:pulumi:dev::aws-ts::aws:rds/instance:Instance::postgresdb]
        [provider: urn:pulumi:dev::aws-ts::pulumi:providers:aws::default_5_41_0::5199a248-3c92-4b11-ae4e-dbf860fd2c76 => urn:pulumi:dev::aws-ts::pulumi:providers:aws::default_6_0_0::output<string>]
      ~ name: "airflow" => "airflow"
    --aws:rds/instance:Instance: (delete-replaced)
        [id=postgresdbd65b294]
        [urn=urn:pulumi:dev::aws-ts::aws:rds/instance:Instance::postgresdb]
        [provider=urn:pulumi:dev::aws-ts::pulumi:providers:aws::default_5_41_0::5199a248-3c92-4b11-ae4e-dbf860fd2c76]
```

### Expected Behavior

No change in preview.

### Steps to reproduce

1. Deploy this program with the latest `^5.0.0` of AWS provider:

```
import * as pulumi from "@pulumi/pulumi";
import * as aws from "@pulumi/aws";
import * as random from "@pulumi/random";

const dbName = "airflow"
const dbUser = "airflow"

[...]

PR #2685, opened 2023-08-09
Fix RDS instance name problem

t0yv0:
The deprecated "name" property is removed from rds.Instance. This is breaking change. 

This is one possible fix for https://github.com/pulumi/pulumi-aws/issues/2682

This is what happens to the repro from 2682 with this fix for the provider and TS SDK:

```
Previewing update (dev)

View in Browser (Ctrl+O): https://app.pulumi.com/t0yv0/aws-2682/dev/previews/e8baa033-ae0a-4e1f-b1f6-21541f567adb

     Type                 Name          Plan     Info
     pulumi:pulumi:Stack  aws-2682-dev           1 error


Diagnostics:
  pulumi:pulumi:Stack (aws-2682-dev):
    error: Running program '/Users/t0yv0/tmp/2023/08/aws-2682/index.ts' failed with an unhandled exception:
    TSError: ⨯ Unable to compile TypeScript:
    index.ts(21,5): error TS2345: Argument of type '{ engine: string; instanceClass: string; allocatedStorage: number; name: string; username: string; password: pulumi.Output<string>; skipFinalSnapshot: true; }' is not assignable to parameter of type 'InstanceArgs'.
      Object literal may only specify known properties, and 'name' does not exist in type 'InstanceArgs'.
```

This is what happens to the repro from 2682 with this fix for JavaScript users (no types):

```
t0yv0@Antons-MacBook-Pro> pulumi preview                                                                                                                                                              ~/tmp/2023/08/aws-2682
Previewing update (dev)

View in Browser (Ctrl+O): https://app.pulumi.com/t0yv0/aws-2682/dev/previews/ef2cc00a-196a-4c70-9395-a989b127a498

     Type                 Name          Plan     Info
     pulumi:pulumi:Stack  aws-2682-dev
     └─ aws:rds:Instance  postgresdb             1 error


Diagnostics:
  aws:rds:Instance (postgresdb):
    error: aws:rds/instance:Instance resource 'postgresdb' has a problem: Invalid or unknown key. Examine values at 'postgresdb.name'.
```

After modifying the example like this:

```typescript
const db = new aws.rds.Instance("postgresdb", {
    engine: "postgres",

    instanceClass: "db.t4g.micro",
    allocatedStorage: 30,

    // name: dbName,
    username: dbUser,
    password: airflowPassword.result,

    skipFinalSnapshot: true,
});
```

The update succeeds without modifying the resource.


t0yv0:
The downside of this fix is that we don't seem to have a convenient way to communicate in-band with the users here and explain why this property is gone. AFAIK it was not marked deprecated in 5.x. If we mark it as deprecated now, we need due diligence to make sure it's really removed before it has a chance to interact with diff internals.

t0yv0:
Closing in favor of #2686 that provides a more gradual transition path for users via a deprecation warning.

PR #2686, opened 2023-08-09
Deprecate `name` on `aws.rds.Instance`

### LINKED ISSUE #2682: RDS Instance replacement when upgrading to 6.0

iwahbe:
Fixes #2682

We should have deprecated `name` in `v5.x`. We could then remove it here. Because we didn't remove it here, we need to deprecate it now, so we can remove at `v7.0.0`. 

We want the following properties when we deprecate `name`:
- Doesn't require a patch upstream.
- `name` is deprecated in the schema.
- `name` warns when used, directing the user to use `dbName` instead.
- `name` is still valid as an input.
- `name` is still valid as an output.

This PR works by back-porting `name` into the upstream schema from the `Provider() tfbridge.ProviderInfo` function. Since the upstream provider is no longer aware of `name`, we set `dbName` to the value of `name` and delete `name`. If both were provided, we error, just like `v5.42.0` did.

As implemented, we get:
- [X] Doesn't require a patch upstream.
- [X] `name` is deprecated in the schema.
- [x] `name` warns when used, directing the user to use `dbName` instead.
- [X] `name` is still valid as an input.
- [X] `name` is still valid as an output.

To warn when we need to, we need to patch the bridge to allow passing a logger into `PreCheckCallback`. This change will be non-breaking.


provider/resources.go
```diff
@@ -7095,7 +7127,19 @@ func Provider() *tfbridge.ProviderInfo {
...
-		prov.Resources[key].PreCheckCallback = applyTags
+		if callback := prov.Resources[key].PreCheckCallback; callback != nil {
+			prov.Resources[key].PreCheckCallback = func(
+				ctx context.Context, config resource.PropertyMap, meta resource.PropertyMap,
+			) (resource.PropertyMap, error) {
+				config, err := callback(ctx, config, meta)
+				if err != nil {
+					return nil, err
+				}
+				return applyTags(ctx, config, meta)
```
guineveresaenger, line 7138: I'm not sure I understand the logic here.

- if a resource that hits this code path has a non-nil callback, we want to apply tags _after_ applying the existing callback function to the config?
- if a resource  that hits this code path _doesn't_ have a callback, or it is nil, we behave as before

Since `rds.Instance` is currently the only resource with an explicit callback, could there ever be a new PreCheckCallback that _doesn't_ want this to happen?

iwahbe:
We can get the warning we want easily after we merge https://github.com/pulumi/pulumi-terraform-bridge/pull/1333.

provider/resources.go
```diff
@@ -2840,6 +2840,38 @@ func Provider() *tfbridge.ProviderInfo {
...
+				PreCheckCallback: func(
+					ctx context.Context, config resource.PropertyMap, meta resource.PropertyMap,
+				) (resource.PropertyMap, error) {
+					if name, ok := config["name"]; ok {
+						// Both `name` and `dbName` are set, so error
+						if _, ok := config["dbName"]; ok {
+							return nil, fmt.Errorf("Cannot specify both name and dbName, " +
+								"please set only dbName")
+						}
+						// Name doesn't actually exist on the underlying provider anymore,
```
t0yv0, line 2869: Question, don't we need to emit a deprecation warning here where we do `config["dbName"] = name`? Or something else will emit a deprecation warning anyway?

provider/resources.go
iwahbe: If there is a callback, we give re-assign `prov.Resources[key].PreCheckCallback` so both the old callback and `applyTags` are called. Otherwise we just set `applyTags`.

If we introduce a new callback where we don't want this to happen, we can change this code. I don't expect that we will though.

provider/cmd/pulumi-resource-aws/schema.json
```diff
@@ -261784,7 +261784,7 @@
                     },
                     "name": {
                         "type": "string",
-                        "willReplaceOnChanges": true
+                        "deprecationMessage": "This property has been deprecated. Please use 'dbName' instead."
```
t0yv0, line 261787: I think as written it will still replace on changes, no? Because you copy the value into dbName which does replace on changes. Can we get "willREplaceOnChanges" marker back? I know it's only for docs.

sdk/dotnet/Rds/Instance.cs
```diff
@@ -489,7 +489,7 @@ public partial class Instance : global::Pulumi.CustomResource
         public Output<bool> MultiAz { get; private set; } = null!;
 
         [Output("name")]
-        public Output<string> Name { get; private set; } = null!;
+        public Output<string?> Name { get; private set; } = null!;
```
t0yv0, line 492: Interesting. Benign? This might be coming out of tfgen changes actually

provider/resources.go
iwahbe: We need to emit the warning, but we don't have a way to do so here. https://github.com/pulumi/pulumi-terraform-bridge/pull/1333 provides the logging context to `PreCheckCallback`. When that merges, we add this patch:

```patch
 provider/resources.go | 2 ++
 1 file changed, 2 insertions(+)

diff --git a/provider/resources.go b/provider/resources.go
index 9706f52d0d..e44a569b3f 100644
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -2866,6 +2866,8 @@ func Provider() *tfbridge.ProviderInfo {
 							return nil, fmt.Errorf("Cannot specify both name and dbName, " +
 								"please set only dbName")
 						}
+						tfbridge.GetLogger(ctx).Warn("The `name` property is deprecated, " +
+							"please use `dbName` instead")
 						// Name doesn't actually exist on the underlying provider anymore,
 						// so we make sure it only sees `dbName`, not `name`.
 						config["dbName"] = name
```

upstream-patches/0015-Un-deprecate-RDS-instance-name.patch
```diff
@@ -1,33 +0,0 @@
...
-
-diff --git a/internal/service/rds/instance.go b/internal/service/rds/instance.go
-index 45afa19e7f..cdecabe089 100644
---- a/internal/service/rds/instance.go
-+++ b/internal/service/rds/instance.go
-@@ -408,6 +408,16 @@ func ResourceInstance() *schema.Resource {
- 				Optional: true,
- 				Computed: true,
- 			},
-+			"name": {
```
t0yv0, line 21: Yes! Here we go, delete this one.

provider/cmd/pulumi-resource-aws/schema.json
iwahbe: It will. I'll add that message back in.

sdk/dotnet/Rds/Instance.cs
iwahbe: This is real and correct. `name` is not set if the user correctly supplies `dbName`.

provider/resources.go
t0yv0: Well you can send a warning to glog  or even printf "WARNING" that should work already AFAIK, even if not ideal (not tied to the resource). 

How does the experience look like right now? We need this for go/no-go.

https://github.com/pulumi/pulumi-terraform-bridge/pull/1333 needs tests and some iteration it seems like, trying to weight in if we have enough time to land that first.

provider/resources.go
guineveresaenger: > If there is a callback, we give re-assign prov.Resources[key].PreCheckCallback so both the old callback and applyTags are called. Otherwise we just set applyTags.

Makes sense!

> If we introduce a new callback where we don't want this to happen, we can change this code. I don't expect that we will though.
How will a future maintainer know to change this code? 
If we don't expect new callbacks to be introduced - should we use the resource name as an explicit trigger here?

iwahbe: > How will a future maintainer know to change this code?

If a future maintainer wants to add a new `PreCheckCallback` to a specific resource that overrides the behavior of tags, they will notice their change is not working and investigate. It's all in this file. They can grep for `PreCheckCallback` in `resources.go` and find this.

> If we don't expect new callbacks to be introduced - should we use the resource name as an explicit trigger here?

I'm not sure what you mean? Can you elaborate?

guineveresaenger: > they will notice their change is not working and investigate.
I don't think that is ideal. I think the design of PreCheckCallback invites one to think "oh neat, I can add a callback in this place, on a resource". Getting a surprise can be really frustrating, regardless of how easy it would be to grep for PreCheckCallback.

> Can you elaborate?
If this particular callback is unique to `rds.Instance`, I think I would prefer logic that explicitly triggers on "if this resource is `rds.Instance`, use this callback". That will avoid a future maintainer having to look why something seemingly unrelated is getting involved with their callback.

I guess my TL;DR is: I don't think that "PreCheckCallback is not nil" is a discerning enough evaluation to decide whether we should call this code. Let's just use `rds.Instance`. After all, we would want to _remove_ this on v7, correct?

fruttu:
Hello, I noticed this PR as we are facing issues when upgrading from version 5.42.0 to 6.2.0 with Python.

Firstly, this warning is probably incorrect:
```warning: name is deprecated: This property has been deprecated. Please use 'dbName' instead.```

As there is no 'dbName' property here: https://github.com/pulumi/pulumi-aws/blob/master/sdk/python/pulumi_aws/rds/instance.py#L19-L81

Pulumi is also attempting to execute a replacement of the ```aws.rds.Instance``` in our case. This is probably not as intended?

When using the deprecated **name** property,Pulumi preview shows:
```aws:rds:Instance   **-*****-***   replace   [diff: +dbName]; 2 warnings``` 

If we change the **name** property to **db_name**, Pulumi preview displays:
```aws:rds:Instance   **-*****-***   replace   [diff: +dbName]```

If we don't set properties **name** or **db_name** Pulumi preview runs without any changes (update/replace) to the RDS instance.

PR #2708, opened 2023-08-22
Begin normalizing workflow files, based off of pulumi/ci-mgmt/#518

iwahbe:
This PR is based off of https://github.com/pulumi/ci-mgmt/pull/518. The Makefile has greater changes, which will be handled in a separate PR.

There are outstanding changes in these files:
- [ ]  Makefile
- [x] .github/workflows/main.yml
- [x] .github/workflows/master.yml
- [x] .github/workflows/nightly-test.yml
- [x] .github/workflows/prerelease.yml
- [x] .github/workflows/release.yml
- [x] .github/workflows/run-acceptance-tests.yml
- [x] .github/workflows/update-bridge.yml
- [X] .github/workflows/command-dispatch.yml
- [X] .github/workflows/community-moderation.yml
- [X] .github/workflows/pull-request.yml
- [X] .github/workflows/resync-build.yml
- [X] .github/workflows/upgrade-bridge.yml
- [X] .github/workflows/upgrade-provider.yml
- [X] .goreleaser.prerelease.yml
- [X] .goreleaser.yml
- [X] .upgrade-config.yml
- [X] scripts/upstream.sh

We will merge as soon as we release v6. I expect the Makefile normalization to occur later.


provider/shim/shim_test.go
```diff
@@ -1,29 +0,0 @@
...
-// limitations under the License.
-
-package shim
-
-import (
-	"context"
-	"testing"
-)
-
-// This checks that any runtime checks in the underlying provider (with patches) are passed.
```
t0yv0, line 22: Can we bring this back in some form? This was a quick test that was introduced because it was useful.

.github/workflows/run-acceptance-tests.yml
```diff
@@ -262,28 +246,6 @@ jobs:
     steps:
     - name: Is workflow a success
       run: echo yes
-
-  go_test_shim:
```
t0yv0, line 266: This is the same useful test being removed here.

.github/workflows/resync-build.yml
```diff
@@ -0,0 +1,123 @@
+# WARNING: This file is autogenerated - changes will be overwritten if not made via https://github.com/pulumi/ci-mgmt
```
t0yv0, line 1: Curious what does this do, is it something working?

t0yv0:
Looks fairly benign except for losing the shim test. Is there a way to link here a PR to aws repo that uses these updated workflows to get through PR verification? Or is there another strategy to test it out? Thanks a ton.

iwahbe:
> Looks fairly benign except for losing the shim test. Is there a way to link here a PR to aws repo that uses these updated workflows to get through PR verification? Or is there another strategy to test it out? Thanks a ton.

I'm not sure what you mean. This is a PR to pulumi-aws that uses the new workflow files to get through PR verification.

t0yv0:
Oh right, perfect. Then this is it. It needs to pass tests and I think I'd like the shim test back. Or at least backlog it. Running go test on the shim module could be a part of standard ci-mgmt things anyway.

provider/shim/shim_test.go
iwahbe: I brought this back as an `extraTest`. It now correctly ties into `sentinel` and `publish`, preventing merging and publishing, respectively, when the test fails.

.github/workflows/resync-build.yml
iwahbe: It looks like it is a manually triggered repo action to sync with ci-mgmt. I didn't know it existed before, but it's apparently standard in our repos.

iwahbe:
> Oh right, perfect. Then this is it. It needs to pass tests and I think I'd like the shim test back. Or at least backlog it. Running go test on the shim module could be a part of standard ci-mgmt things anyway.

I just added back the shim test.

t0yv0 (approved):
🚢

PR #2711, opened 2023-08-23
WIP on upgrade testing

t0yv0:
IDEA: https://docs.google.com/document/d/1AEjG78Kvgu_sLG4OXSJxy6SXPcaM2IX_DGyn_GBC9hQ/edit?usp=sharing

- [x] recording works (ProgramTest runs against the cloud)
- [x] full upgrade test works (ProgramTest runs against the cloud, but does stack import from pre-recorded V-prev state)
- [x] unit testing works without the cloud 
   - [ ] however replaying Diff methods is not quite accurate and can generate false positives
- [ ] hook into CI

My next steps is to add an rds.Instance configuration to confirm this can detect the regression on rds.Instance, re-record and try to port to master branch representing v6.


tests/provider_upgrade_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+	"github.com/pulumi/pulumi/pkg/v3/testing/integration"
+)
+
+const (
+	providerBinary = "pulumi-resource-aws"
+)
+
+var (
+	accept bool = os.Getenv("PULUMI_ACCEPT") != ""
```
iwahbe, line 27: I know this is common, but sometimes I run `PULUMI_ACCEPT=false` and then it gets really confusing. 

```go
import 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil"
``` 
```suggestion
	accept bool = cmdutils.IsTruthy(os.Getenv("PULUMI_ACCEPT"))
```

tests/provider_upgrade_test.go
```diff
@@ -0,0 +1,182 @@
...
+)
+
+// Recording works, but how do we test state import?
```
iwahbe, line 30: What properties are we trying to assert?

t0yv0:
Closing in favor of https://github.com/pulumi/pulumi-aws/pull/2714 that supersedes this work.

PR #2763, opened 2023-08-30
Combined: Bridge Update & #2761 & #2681

iwahbe:
This PR combines #2761, #2681 and a bridge update into 1 PR. 

We are combining PRs to reduce the CI time needed to make a release.

PR #2761, opened 2023-08-30
Upgrade upstream v5.14.0

iwahbe:
Fixes https://github.com/pulumi/pulumi-aws/issues/2712
Fixes https://github.com/pulumi/pulumi-aws/issues/2703
Fixes https://github.com/pulumi/pulumi-aws/issues/2702
Fixes https://github.com/pulumi/pulumi-aws/issues/2693

PR #2766, opened 2023-08-30
Prepare for v6.0.3 Patch

iwahbe:
https://github.com/pulumi/pulumi-aws/pull/2763 is blocked since https://github.com/pulumi/pulumi-aws/pull/2761 fails CI tests due to OOM. This PR has excluded #2761.


mjeffryes (approved):
I'm not totally in context here, but approving based on the preveious approvals for https://github.com/pulumi/pulumi-aws/pull/2763

PR #2768, opened 2023-08-31
Always run extra test

iwahbe:
Under the new workflow files, this test is included whenever we run our other tests. Importantly, the post-test phase of the relevant workflows depend on this test running successfully. We can't skip the test (for example [during releases](https://github.com/pulumi/pulumi-aws/actions/runs/6032096288), since that blocks the post-test actions).

This PR reflects the changes made in https://github.com/pulumi/ci-mgmt/pull/538.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -450,7 +450,6 @@ jobs:
         - go
         - java
   go_test_shim:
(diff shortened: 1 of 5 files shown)


t0yv0:
It's a nice fail-fast that checks issues with incorrectly patching upstream. We introduced it as follow up on some failure @thomas11 - I can dig it up if needed.

PR #2770, opened 2023-08-31
Switch from ssh to https for submodule

iwahbe:
Fixes https://github.com/pulumi/ci-mgmt/issues/539

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,7 +157,7 @@ init_upstream:
 		cd upstream && \
 		git submodule update --init && \
 		(git remote rm source || true) && \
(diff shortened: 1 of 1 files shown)


mjeffryes (approved):

danielrbradley:
Why does SSH break ci-mgmt?

iwahbe:
Github rejects the ssh connection ([failure](https://github.com/pulumi/ci-mgmt/actions/runs/6033255088/job/16369683045)). The rest of our providers use https, which works fine. I don't believe that GH actions come with pre-configured SSH keys.

Issue #2774, opened 2023-09-01
EKS tests fail with a panic


t0yv0:
### What happened?

Looks like it's possible for apply tags to panic on V6:

https://github.com/pulumi/pulumi-aws/pull/2655/files

From https://github.com/pulumi/pulumi-eks/actions/runs/6049094022/job/16415977219?pr=901

```
  panic: interface conversion: interface {} is resource.Computed, not map[string]interface {}
  
  goroutine 235 [running]:
  
  github.com/pulumi/pulumi-aws/provider/v6.applyTags({0xe943ba8, 0xc00e8a7ad0}, 0xc00e8a7b30, 0x6?)
  
  	/home/runner/work/pulumi-aws/pulumi-aws/provider/resources.go:7231 +0x954
  
  github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge.(*Provider).Check(0xc0004a5340, {0xe943ba8?, 0xc00e8a7a40?}, 0xc00e8b22a0)
  
  	/home/runner/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.58.0/pkg/tfbridge/provider.go:573 +0xbb7
  
  github.com/pulumi/pulumi-terraform-bridge/x/muxer.(*muxer).Check.func1({0xe96b790?, 0xc0004a5340?})
  
  	/home/runner/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/x/muxer@v0.0.7-0.20230801203955-5d215c892096/muxer.go:332 +0x36
  
  github.com/pulumi/pulumi-terraform-bridge/x/muxer.resourceMethod[...](0xc00b9d94a0?, 0x40, 0xc000b05780?)
  
  	/home/runner/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/x/muxer@v0.0.7-0.20230801203955-5d215c892096/muxer.go:303 +0xbd
  
  github.com/pulumi/pulumi-terraform-bridge/x/muxer.(*muxer).Check(0x0?, {0xe943ba8?, 0xc00e8a7a40?}, 0x40?)
  
  	/home/runner/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/x/muxer@v0.0.7-0.20230801203955-5d215c892096/muxer.go:331 +0x68
  
[...]

PR #2776, opened 2023-09-01
Fix panics in applyTags when tags are unknown

t0yv0:
Fixes https://github.com/pulumi/pulumi-aws/issues/2774

We detected an issue in pulumi-eks test suite against the latest v6.0.3 release. This theoretically could affects users that set resource and or provider tags to unknown Outputs, although I tried and failed to reproduce outside of the pulumi-eks test.

The fix for https://github.com/pulumi/pulumi-aws/issues/1655 included reimplementing tag merging (computing effective tags from the set of provider-level tags and resource-level tags) on Pulumi side as a callback to the Check Pulumi gRPC method. This works against resource.PropertyValue model and did not fully account for receiving a Computed when tags are unknown, resulting in a panic in the eks test.

The fix puts some combinators in place to shield resource.PropertyValue transformations from observing Unknowns, Computeds, and first-class Outputs. Note that this is a bit forward looking since this is a bridged provider, and as such on current version of tf-bridge it opts into a simplified version of the wire protocol where for example first-class Output values should not be passed to the bridge code. So it's not strictly the simplest fix, but hopefully this can be useful down the line.

I'd like to consider upstreaming propertyvalue.TransformErr (should it be called Walk in Go?) and similar combinators to pulumi/pulumi at some point if they prove useful.


provider/tests/provider_test.go
```diff
@@ -51,3 +51,37 @@ func replaySequence(t *testing.T, sequence string) {
 func TestReplay(t *testing.T) {
 	replaySequence(t, `[]`)
 }
+
+func TestRegressUnknownTags(t *testing.T) {
+	repro := `
+	[
+	  {
+	    "method": "/pulumirpc.ResourceProvider/Check",
```
t0yv0, line 59: This got extracted as a repro from the failing EKS test.

```
#!/usr/bin/env bash

set -euo pipefail

export AWS_PROFILE=devsandbox
export PATH="$PWD/bin:$PATH"
export AWS_REGION=us-west-2
export PULUMI_DEBUG_GRPC="$PWD/here.json"

echo "RECORDING TO $PULUMI_DEBUG_GRPC"

export PATH="/Users/t0yv0/code/pulumi-aws/bin:$PATH"
(cd examples && go test -tags=nodejs -v  --run=TestAccTags)

```

provider/tags_test.go
```diff
@@ -0,0 +1,261 @@
...
+// limitations under the License.
+
+package provider
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"pgregory.net/rapid"
```
iwahbe, line 22: 🎉

provider/pvutil.go
```diff
@@ -0,0 +1,121 @@
...
+//
+//	composePropertyValue[T](func (c *pvComposer) (resource.PropertyValue, T, error) {
+//	      x, _ := c.Simplify(pv1)
+//	      y, _ := c.SimplifyPropertyMap(pv2) // etc
+//	      return resource.NewArrayProperty([]resource.PropertyValue{x, y}), result, nil
+//	})
+//
+// User code accessing values that passed through Simplify is guaranteed to never observe Secret,
+// Computed, or Output values. All the metadata bits about these is floated to top-level and
+// re-applied to the value the user code receives out of composePropertyValue.
```
iwahbe, line 35: I think this is fine. It means that `tags: { "foo": "bar", "secret": [secret] }` will come out as `[secret]`. I doubt that this will be a problem in practice. I don't want to block merging on it.

provider/tests/provider_test.go
t0yv0: Ouch, good catch 😢

provider/tags.go
```diff
@@ -0,0 +1,135 @@
...
+	if t, ok := config["tags"]; ok {
+		configTags = t
+	}
+	allTags, hasTags, err := mergeTags(ctx, configTags, meta)
+	if err != nil {
+		return nil, err
+	}
+	// If there are 0 tags, delete the tags entry rather than sending an empty map. The unknown
+	// case is quirky though, prefer to send the unknown marker out rather than deleting it.
+	if !hasTags && !allTags.ContainsUnknowns() {
```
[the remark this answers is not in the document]
t0yv0: Ah but it also can be an Output{Element: nil, Dependencies}, among other things.

provider/pvutil.go
t0yv0: Hmm you're right. We're losing precision. Unlike what happens in the restricted CheckConfigure, there are true nested secrets here.  That's bad especially when merging is not even involved, like empty provider tags + secret resource tags. Hm.. I may do a quick fix to this by keeping a track of secret string hashes and re-secreting matching in the output.

provider/tags.go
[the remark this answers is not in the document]
t0yv0: Well you're right. IsNil() is the only case we truly want to elide. The condition should just be a nil-check

t0yv0:
Verifying secret handling for tags at integration level, it's not what you'd expect.

First off all if you pass secrets, the provider unwraps the secrets and stores them in tagsAll property. So the protections are defeated.

Second if you start from a program that does not have secrets, and you mark a resource input such as tags secret, you don't get an update plan..... I think we need to trace this down as this is more generic than tags. It might be an issue with providers per se.

Third. Secrets never reach the code edited here.


```typescript
import * as awsx from "@pulumi/awsx";

const b2 = new aws.s3.Bucket("my-bucket-40", {}, {});

// Create an AWS provider for the us-east-1 region.
let useast1 = new aws.Provider("useast1", {
    region: "us-east-1",
    defaultTags: {tags: {"this": b2.arn }},
});


// Create an AWS resource (S3 Bucket)
const bucket = new aws.s3.Bucket(
    "my-bucket",
    {tags: {"that": pulumi.secret(b2.arn)}},
    {provider: useast1});

// Export the name of the bucket
export const bucketName = bucket.id;

// interesting doe adding a <secret> wrapper not update ?
```

Gives

```
{
  "request": {
    "urn": "urn:pulumi:dev::repro-aws-tagpanic::aws:s3/bucket:Bucket::my-bucket",
    "olds": {},
    "news": {
      "tags": {
        "that": "arn:aws:s3:::my-bucket-40-f07f095"
      }
    },
    "randomSeed": "25PggOkN2OdDhrlXlBtn2O4buWU6YkMyRY0F9JfeExw="
  },
  "response": {
    "inputs": {
      "__defaults": [
        "acl",
        "bucket",
        "forceDestroy"
      ],
      "acl": "private",
      "bucket": "my-bucket-d35e552",
      "forceDestroy": false,
      "tags": {
        "__defaults": [],
        "that": "arn:aws:s3:::my-bucket-40-f07f095",
        "this": "arn:aws:s3:::my-bucket-40-f07f095"
      }
    }
  },
  "metadata": {
    "kind": "resource",
    "mode": "client",
    "name": "aws"
  }
}
```

But in the state we get:

```
                    "tags": {
                        "that": {
                            "4dabf18193072939515e22adb298388d": "1b47061264138c4ac30d75fd1eb44270",
                            "ciphertext": "AAABAM8fGIsN3ZQv0HSv4nfSh0oCujcIdKGLU6lEgZHtfFIvmXRZt4tCv8XlUfAtJZSaYOTaMZHKm2TgBuBSiraF7w=="
                        },
                        "this": "arn:aws:s3:::my-bucket-40-f07f095"
                    },
                    "tagsAll": {
                        "that": "arn:aws:s3:::my-bucket-40-f07f095",
                        "this": "arn:aws:s3:::my-bucket-40-f07f095"
                    },
```

Squinting at this, the engine must have applied some secrets handling so the provider did not even get exposed to it.

t0yv0:
OK. I poked around some more. First-class outputs or nested computed values or secrets never reach this code, they never reach Check.. This is all the idiosyncratic behavior of the bridge.

The only place is CheckConfig.

The repro is:

```typescript
const b2 = new aws.s3.Bucket("my-bucket-40", {}, {});

// Create an AWS provider for the us-east-1 region.
let useast1 = new aws.Provider("useast1", {
    region: "us-east-1",
    defaultTags: {tags: {"this": b2.id }},
});
```

IN this case we get the unknown sentinel.

```
{
  "method": "/pulumirpc.ResourceProvider/CheckConfig",
  "request": {
    "urn": "urn:pulumi:dev::repro-aws-tagpanic::pulumi:providers:aws::useast1",
    "olds": {},
    "news": {
      "defaultTags": "04da6b54-80e4-46f7-96ec-b56ff0331ba9",
      "region": "us-east-1",
```

Note that the "this" structure is destroyed already, it's one big unknown. Unfortunately, PreCheckCallback gets this down because it recalls the map from CheckConfig and gets sent it.. So this is where we get the unknown from. This is the only realistic case that can happen given how current bridge + pulumi CLI work.

t0yv0:
Simplified based on the findings, PTAL.

iwahbe (approved):
LGTM

Issue #2757, opened 2023-08-29
Fixup docs with `tfbridge.ProviderInfo.DocRuleInfo` instead of `upstream-tools`


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Move find-and-replace from upstream-tools to tfbrige.ProviderInfo.DocRuleInfo. By increasing the number of automated replaces, and only requiring replaces on docs we intend to embed, we should see more automatable upgrades.

We should also see a speed increase on building locally, since we only need to read each docs file once, instead of two reads and write per file (write is only on changes). This will be on the order of 1-2 seconds, and is not the primary motivator.

Issue #2756, opened 2023-08-29
Normalize `Makefile` to ci-mgmt


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

`pulumi-aws` currently uses a bespoke Makefile. To simplify use and allow automated upgrades, we should convert to a Makefile generated by ci-mgmt.

This involves switching upstream to a patch managed system, as opposed to a fork managed  one. The important part of this PR is that the upgrade process looks just like a normal bridged provider, and that we can automate with `upgrade-provider`.

To accommodate for the special behavior still needed by pulumi-aws (specifically running `upstream-tools`), we can add a hook system to the generated Makefile and import a local Makefile (if it exists). The goal here is to keep any pulumi-aws specific Make targets in the pulumi-aws repo without breaking the build.

Related: https://github.com/pulumi/ci-mgmt/pull/518 disabled `Makefile` generation for pulumi-aws. As part of closing this issue, we should undo [this commit](https://github.com/pulumi/ci-mgmt/pull/518/commits/0e7a98ddb312f3d4b1c4eceb1d0a9bb556dcf9fc).

Like https://github.com/pulumi/pulumi-aws/pull/2708, the PR that closes this issue will need to be duplicated in pulumi-aws and ci-mgmt.

### Affected area/feature

The Makefile.

PR #2784, opened 2023-09-05
Remove upstream fork and normalize Makefile

guineveresaenger:
This pull request represents part of the work for #2756.

Three overall changes occur in this PR:

1. Removal of dependency on pulumi/terraform-provider-aws in favor of hashicorp/terraform-provider-aws as a submodule.
Existing patches were successfully applied to the new upstream submodule, which is checked out at v5.11.0. `make tfgen` showed no diff.

2. Transformation of shim folder to patch applied to upstream

2. Normalization of Makefile to follow paradigms in other bridged providers; specifically:

- Removal of extra `check_schema` target
- Removal of `patch_upstream` target as unused
- Removal of check for existing `bin/pulumi-java-gen` file in favor of always downloading the newest version. This is worth following up on and perhaps reintroducing in all providers as an optimization.
- Removal of PULUMI_REPO_PATHS environment variable in favor of the `UpstreamRepoPath` provider option
- Use of `printf` over `echo`
- Removal of `--overlays` flag in language build targets: to the best of my knowledge, these were unused
- Update of `build_go` to check for modules

One variation in the Makefile will be preserved: 

1. the`upstream` target includes additional steps for running the patching tool. A corresponding change to pulumi/ci-mgmt will correctly generate this variation. This variation is planned to be removed with completion of #2757.

EDITED TO ADD:

Provider Upgrade tests were added in #2714. They were added to the Makefile without a corresponding change in pulumi/ci-mgmt. This pull request removes these for the purposes of speedy alignment with ci-mgmt. As per discussion in code review, we can add a way to include extra manually maintained Makefiles to our providers via `-include` directive, and we could add these targets there. (TODO: file an issue to this effect).

These Upgrade tests rely on resources present in our upstream patches. The integration tests link to `pulumi-aws` as a dependency, and thus we needed to add an extra `make upstream` step to the testing workflow. This change is reflected in the [corresponding ci-mgmt PR](https://github.com/pulumi/ci-mgmt/pull/551).


Makefile
```diff
@@ -141,66 +124,28 @@ tfgen: install_plugins upstream
...
-	cd upstream && LAST_TAG=$$(git describe --abbrev=0 --tags) && \
-		git format-patch  -o ../upstream-patches --minimal --no-signature HEAD...$${LAST_TAG}
+upstream::
+# Use upstream.sh script to apply patches
+ifneq ("$(wildcard upstream)","")
+	@$(SHELL) ./scripts/upstream.sh "$@" apply
 endif
 
-upstream: init_upstream export_upstream_patches
+upstream::
```
iwahbe, line 133: Instead of including this twice in the same Makefile, can we instead put a guarded [include](https://www.gnu.org/software/make/manual/html_node/Include.html) at the top and bottom of all our generated Makefiles, and supply this upstream definition there?

We could put this before we start defining targets:
```make
-include pre.mk
```

After we have defined all targets, we could put this:
```make
-include post.mk
```

In this case, `post.mk` would be:
```make
upstream::
	# Ensure tool is installed
	cd upstream-tools && yarn install --frozen-lockfile
	# Apply all automated changes
	cd upstream-tools && yarn --silent run apply
	# Check for any pending replacements
	cd upstream-tools && yarn --silent run check
```

`pre.mk` and `post.mk` would be checked in the provider repo, but never seen by ci-mgmt.

iwahbe (approved):
> Provider Upgrade tests were added in https://github.com/pulumi/pulumi-aws/pull/2714. They were added to the Makefile without a corresponding change in pulumi/ci-mgmt. It is not clear to me if these extra `test.upgrade` etc. targets should be preserved in the main Makefile and I would like to reach consensus here.

These tests should continue to be invoked by CI, and ideally they would be invoked when running `make test`.

---

I would like to merge this PR asap to prevent continued drift from ci-mgmt. I'm comfortable merging this PR as is, and then cleaning up how we include additions and tests later.

Makefile
guineveresaenger: Why do we want a pre.mk?

iwahbe (approved):

PR #2802, opened 2023-09-11
Remove deprecation warning on `aws_s3_bucket_object`

### LINKED ISSUE #2796: warning: use the aws_s3_object resource instead
In any program that uses aws.s3.BucketObject, I see this warning.  

I don't see anything about deprecation of this resource in our migration guide, and the error message suggests a TF name, and the name it suggests doesn't have any obvious equivalent in Pulumi. 

```
aws:s3:BucketObject (index.html):
    warning: urn:pulumi:dev::aws-ts-s3-folder::aws:s3/bucketObject:BucketObject::index.html verification warning: use the aws_s3_object resource instead
```

iwahbe:
Fixes #2796

Upstream has added a depreciation warning to `aws_s3_bucket_object`, deprecating it in favor of `aws_s3_object` (`aws:s3/bucketObjectV2:BucketObjectV2`).

This PR is the quick fix, simply removing the message.

Long term, we should come up with a plan to right upstream against s3:
1. Deprecate `aws:s3/bucketObject:BucketObject` and `aws:s3/bucketObjectV2:BucketObjectV2` in favor of `aws:s3/object:Object`.

  If we can confirm it is safe to use `aws_s3_object` with our `aws:s3/bucket:Bucket`, we can hard alias `aws:s3/bucketObjectV2:BucketObjectV2` to `aws:s3/object:Object` and restore the deprecation warnings. If not, we will need to wait until we can resolve (2) to resolve (1).

2. Find a migration path back to where `aws:s3/bucket:Bucket` is the preferred bucket to use (mirroring upstream). Then deprecate `aws:s3/bucketV2:BucketV2` in favor of `aws:s3/bucket:Bucket`.

### DIFF
--- a/examples/examples_test.go
+++ b/examples/examples_test.go
@@ -3,7 +3,9 @@
 package examples
 
 import (
+	"bytes"
 	"context"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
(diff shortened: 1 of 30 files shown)


iwahbe:
Cursory tests show that we can attach a `aws:s3/bucketObjectV2:BucketObjectV2` to an `aws:s3/bucket:Bucket`:
```yaml
resources:
  bucket:
    type: aws:s3:Bucket
  o1:
    type: aws:s3:BucketObjectv2
    properties:
      bucket: ${bucket}
      content: "This seems to work in a legacy bucket"
```

This indicates we can accomplish step [(1)](https://github.com/pulumi/pulumi-aws/pull/2802#issue-1891362700) now.

t0yv0:
I think adding a test is in order here that would have caught this. If this was added in the first place we wouldn't have shipped this. Unfortunately.. I know there's probably no clean way to test.

iwahbe:
> I think adding a test is in order here that would have caught this. If this was added in the first place we wouldn't have shipped this. Unfortunately.. I know there's probably no clean way to test.

I'll look for a way to add a test. In general, we wouldn't add a test, since we accept most of upstreams deprecations.

guineveresaenger (approved):
This looks fine to me.

I'm not sure tests are needed here - the message is unfortunate as it causes user confusion, but removing it will not change any fundamental resource behavior.

I'm in favor of a followup issue/deprecation plan for `V2` that aims to accomplish the steps outlaid by @iwahbe. Testing can assert these resources function as expected with each other.

PR #2825, opened 2023-09-22
Speed up provider startup time through improved metadata parsing

t0yv0:
Upgrading to a pre-release version to propagate the mitigation to provider startup time.

The changes to the bridge look benign, can be reviewed here:

https://github.com/pulumi/pulumi-terraform-bridge/compare/v3.60.1...b0e78c54cff8

Next week we will do a proper bridge release and this provider will catch up.


iwahbe (approved):
Why not release a patch of the bridge now?

PR #2834, opened 2023-09-26
Seamlessly Migrate from `name` to `dbName`

### LINKED ISSUE #2823: aws.rds.Instance replacement when upgrading from 5.42.0 to 6.2.0
### What happened?

Pulumi is attempting to execute a replacement of the ```aws.rds.Instance``` in our case when updating to newest pulumi-aws. This is probably not as intended?

### Example

When using the deprecated **name** property.
Pulumi preview shows:
```aws:rds:Instance   **-*****-***   replace   [diff: +dbName]; 2 warnings```

Pulumi preview --diff shows:
```
warning: "name" is deprecated, use "dbName" instead.
warning: "name" is deprecated, use "dbName" instead.
        +-aws:rds/instance:Instance: (replace)
            [id=xxx-xxxx-rdsd5e5988]
            [urn=urn:pulumi:xxx::xxxx::Vpc$aws:rds/instance:Instance::xx-xxx-rds]
```

If we change the **name** property to **db_name**
Pulumi preview displays:
```aws:rds:Instance   **-*****-***   replace   [diff: +dbName]```

Pulumi preview --diff displays:
```
warning: name is deprecated: This property has been deprecated. Please use 'dbName' instead.
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:xxx::xxx::pulumi:pulumi:Stack::xxx]
        +-aws:rds/instance:Instance: (replace)
            [id=xxx-xxxx-rdsd5e5988]
            [urn=urn:pulumi:xxx::xxxx::Vpc$aws:rds/instance:Instance::xx-xxx-rds]
```

If we don't set properties **name** or **db_name** Pulumi preview runs without any changes (update/replace) to the RDS instance.[](url)

### Output of `pulumi about`

```
CLI          
Version      3.84.0
Go Version   go1.21.1
Go Compiler  gc

Plugins
NAME        VERSION
aws         6.2.0
docker      3.6.1
kubernetes  4.2.0
python      unknown

Host     
OS       ubuntu
Version  20.04
Arch     x86_64

This project is written in python: executable='/home/xxxxx' version='3.8.10
```

### Additional context

This warning is probably incorrect:
```warning: name is deprecated: This property has been deprecated. Please use 'dbName' instead.```

As there is no 'dbName' property here: https://github.com/pulumi/pulumi-aws/blob/master/sdk/python/pulumi_aws/rds/instance.py#L19-L81

My original post can be found here https://github.com/pulumi/pulumi-aws/pull/2686#issuecomment-1729223015

### Contributing

### PR DESCRIPTION
**Pull request by iwahbe**
Fixes #2823 
Depends on https://github.com/pulumi/pulumi-terraform-bridge/pull/1400 in the bridge.

This commit relies on an unreleased commit in the bridge: https://github.com/pulumi/pulumi-terraform-bridge/commit/e3fd9503f2d36e0f77fbed32a373ec0dd607f222.

### CODE REVIEW

File: examples/go.mod [hunk 1]
```diff
@@ -263,7 +262,7 @@ require (
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pkg/term v1.1.0 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	github.com/pulumi/pulumi-terraform-bridge/v3 v3.60.2-0.20230922172747-b0e78c54cff8 // indirect
+	github.com/pulumi/pulumi-terraform-bridge/v3 v3.60.2-0.20230926212355-e1ac6a1023a9 // indirect
```
t0yv0, line 265: This commit is on master correct?

provider/resources.go
```diff
@@ -2868,6 +2868,20 @@ func Provider() *tfbridge.ProviderInfo {
...
+					if _, ok := inputs["db_name"]; ok {
+						return inputs, nil
+					}
+					name, ok := inputs["name"]
+					if !ok {
+						// It's seems like inputs are not in a valid state
+						return inputs, nil
+					}
+					tfbridge.GetLogger(ctx).Debug(`migrating from "name" to "dbName"`)
+					inputs["dbName"] = name
```
t0yv0, line 2881: Can you explain what's the difference between inputs["db_name"] and inputs["dbName"]?

Should not this always be "dbName" in Pulumi states?

provider/resources.go
```diff
@@ -2868,6 +2868,20 @@ func Provider() *tfbridge.ProviderInfo {
 					}
 					return config, nil
 				},
+				TransformFromState: func(ctx context.Context, inputs resource.PropertyMap) (resource.PropertyMap, error) {
```
t0yv0, line 2871: Maybe call it stateInputs to emphasize?

provider/tests/go.mod
```diff
@@ -5,15 +5,14 @@ go 1.21
...
-	github.com/hashicorp/terraform-provider-aws/shim => ../shim
 	github.com/hashicorp/vault => github.com/hashicorp/vault v1.2.0
 	github.com/pulumi/pulumi-aws/provider/v6 => ../
 )
 
 require (
 	github.com/pulumi/pulumi-aws/provider/v6 v6.0.0-00010101000000-000000000000
-	github.com/pulumi/pulumi-terraform-bridge/pf v0.16.0
-	github.com/pulumi/pulumi-terraform-bridge/testing v0.0.1
+	github.com/pulumi/pulumi-terraform-bridge/pf e1ac6a1023a9840cd59e1e68ba7ab87174c5939b
```
t0yv0, line 14: This doesn't look like it received a `go mod tidy`? Or does Go allow hash refs now?

provider/resources.go
```diff
@@ -2868,6 +2868,20 @@ func Provider() *tfbridge.ProviderInfo {
...
+						return inputs, nil
+					}
+					name, ok := inputs["name"]
+					if !ok {
+						// It's seems like inputs are not in a valid state
+						return inputs, nil
+					}
+					tfbridge.GetLogger(ctx).Debug(`migrating from "name" to "dbName"`)
+					inputs["dbName"] = name
+					delete(inputs, "name")
```
t0yv0, line 2882: I'm curious that this helps out the diff but it doesn't actually migrate right. It doesn't rewrite the inputs in the state. Perhaps Check does that already?

Basically my question is whether after all is said and done and users upgrade successfully, they still have "name" in their state, or the normal pathway completely cleans it out now.

provider/resources.go
iwahbe: Your right. That should be `dbName`.

provider/resources.go
iwahbe: It depends. Pulumi doesn't have a mechanism to persist to the state beyond the normal CRUD lifecycle. If the user runs an update (preview=false) or refresh (preview=false), then their new state will be corrected. If they just run `pulumi preview` then this hook will run each time but never correct their state.

provider/tests/go.mod
iwahbe: I don't believe that these tests are run, so my assumption is it missed the go mod tidy.

provider/resources.go
t0yv0: But we're asking the users to correct the program inputs in Check with a warning right. So I think once they correct program inputs and do a pulumi up, I wonder if we update to new state or it's a no-op diff. Eh. This is tricky to reason about - I worry that the old states are going to hang around indefinitely, possibly; there's no easy way for Pulumi providers to issue state corrections that migrate Pulumi states with Updates or some such but otherwise leave things as-is.

t0yv0 (approved):
Worth including the P1 fix. Thank you!

provider/resources.go
iwahbe: I believe that they will stay on the old state until they do something else to the resource. To resolve the underlying issue, either the pulumi CLI or the bridge would need to support a "migration" concept, where a resource declares that it needs to update state because of a new provider version instead of a user change.

PR #2858, opened 2023-10-06
Format diagnostics from GetAwsConfig as readable strings instead of %v format dump

mikhailshilkov:
A tactical improvement for https://github.com/pulumi/pulumi-aws/issues/2285 to avoid Go pointers leaking into error messages.

Before:

> unable to validate AWS credentials.
    Details: [{0x14009acaee0}]

After:

> unable to validate AWS credentials.
    Details: validating provider credentials: retrieving caller identity from STS: operation error STS: GetCallerIdentity, failed to resolve service endpoint, endpoint rule error, Invalid Configuration: Missing Region

### DIFF
--- a/provider/diag.go
+++ b/provider/diag.go
@@ -1,0 +1,22 @@
+// Copyright 2016-2023, Pulumi Corporation.
+
+package provider
+
+import (
+	"github.com/hashicorp/aws-sdk-go-base/v2/diag"
+)
+
+// formatDiags converts AWS SDK Diagnostics into a human-readable string.
+func formatDiags(diag diag.Diagnostics) string {
+	details := ""
+	for _, d := range diag {
+		if details != "" {
+			details += "\n"
+		}
+		details += d.Summary()
+		if d.Detail() != "" {
+			details += ". " + d.Detail()
+		}
+	}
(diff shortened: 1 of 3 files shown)


provider/diag.go
```diff
@@ -0,0 +1,22 @@
...
+
+package provider
+
+import (
+	"github.com/hashicorp/aws-sdk-go-base/v2/diag"
+)
+
+// formatDiags converts AWS SDK Diagnostics into a human-readable string.
+func formatDiags(diag diag.Diagnostics) string {
+	details := ""
```
thomas11, line 11: [nit] `strings.Builder`? Not worth it here?

thomas11 (approved):

provider/diag.go
t0yv0: I'd say it would not be noticeable for end-users unless something pushes MBs into it.

t0yv0 (approved):

provider/diag.go
mikhailshilkov: Changed to strings.Builder. I think it's reasonable readable and overall good hygiene, thank you for the suggestion.

PR #2886, opened 2023-10-13
Move replacements.json into bridge

### LINKED ISSUE #2884: Move `replacements.json` application to a `Provider().RuleInfo`
## Hello!

## Issue details

This will allow us to apply our TODO filters only to text that winds up in our docs, and move the replacement logic into this repo for fast iteration.

iwahbe:
This PR replaces the docs changes described in `replacements.json` with equivalent logic in `doc_edits.go`. Instead of applying changes to the upstream docs repo, changes are applied in memory as part of docs reading.

Fixes #2884


provider/cmd/pulumi-tfgen-aws/main.go
```diff
@@ -28,5 +28,11 @@ func main() {
 		removeUnusedQuicksightTypes(spec)
 	}
 
+	defer func() {
+		for _, f := range aws.PostTfgenHook {
+			f()
+		}
+	}()
```
danielrbradley, line 35: This looks a little odd - why is it here?

provider/doc_edits.go
```diff
@@ -15,30 +15,166 @@
...
+	for k, v := range replacements {
+		i := strings.LastIndexByte(k, '/')
+		if i == -1 {
+			continue
+		}
+		delete(replacements, k)
+
+		key := k[i+1:]
+		replacements[key] = append(replacements[key], v...)
+	}
```
danielrbradley, line 79: What's the purpose of this block? We're deleting entries which are actually folders (based on the trailing `/`)?

provider/doc_edits.go
```diff
@@ -15,30 +15,166 @@
...
+
+	fmt.Printf("Gathered %d replacements\n", len(replacements))
+	var applied int
+
+	checkForTODOs := make(chan struct {
+		path    string
+		content []byte
+	})
+	doneWithTODOs := make(chan struct{})
```
danielrbradley, line 88: Why are we creating channels which are then awaited on in a `PostTfgenHook`? This seems quite a convoluted setup for applying replacements - what options do we have to make this simpler to follow and debug and reason about?

danielrbradley (commented):
Why were so many replacements able to be removed (going from 20k lines to 3k)? Was this done by just generalising the code block import replacement?

The end result looks good - a quick click around the affected resource docs looks okay. There's a few notes which seemed to have been dropped - are these deliberate?

The implementation seems a little convoluted with channels - where has this complexity come from?

Makefile
```diff
@@ -132,8 +132,6 @@ endif
 	cd upstream-tools && yarn install --frozen-lockfile
 	# Apply all automated changes
 	cd upstream-tools && yarn --silent run apply
-	# Check for any pending replacements
```
t0yv0, line 135: If not enabled the checks would be failing right? Skipping this doesn't forgo any validation we had before? Just double-checking.

iwahbe: We do not forgo any "validation" that we used to have. `yarn run check` looked for instances of `[tT]erraform` still remaining in the upstream docs, and added them as TODOs to `replacements.json`. This is now done in `PostTfgenHook`.

This step never failed, and still doesn't, so I wouldn't really call it validation.

provider/cmd/pulumi-tfgen-aws/main.go
iwahbe: We need a mechanism to do things after generation has happened: We need to write out the new `replacements.json` with added TODOs.

provider/doc_edits.go
iwahbe: To move the work of adding TODOs to a different thread then the one doing replacements. I have removed them.

Makefile
t0yv0: Sounds good. Just making sure since I didn't stop to read what it did but you probably read it. Thank you! All good then.

sdk/dotnet/CleanRooms/Collaboration.cs
```diff
@@ -61,11 +61,11 @@ namespace Pulumi.Aws.CleanRooms
...
     /// ## Import
     /// 
-    /// terraform import {
+    /// Using `pulumi import`, import `aws_cleanrooms_collaboration` using the `id`. For example:
     /// 
-    ///  to = aws_cleanrooms_collaboration.collaboration
-    /// 
-    ///  id = "1234abcd-12ab-34cd-56ef-1234567890ab" } Using `pulumi import`, import `aws:cleanrooms/collaboration:Collaboration` using the `id`. For exampleconsole % pulumi import aws:cleanrooms/collaboration:Collaboration 1234abcd-12ab-34cd-56ef-1234567890ab
+    /// ```sh
+    ///  $ pulumi import aws:cleanrooms/collaboration:Collaboration collaboration 1234abcd-12ab-34cd-56ef-1234567890ab
```
t0yv0, line 67: Nice

sdk/dotnet/S3/BucketObjectv2.cs
```diff
@@ -155,8 +155,6 @@ namespace Pulumi.Aws.S3
     /// 
     /// ## Import
     /// 
-    /// Import using the `id`, which is the bucket name and the key together:
```
t0yv0, line 158: This looks good, it was dangling.

t0yv0 (approved):
Spot-checking some resources, this looks like a massive improvement, thank you!

PR #2890, opened 2023-10-16
Build bridge on SDK runners

guineveresaenger:
This PR is a manual run of `make ci-mgmt` to apply SDK-build capable runners to the Upgrade Bridge workflow.

~Will follow this up with either a ticket or a PR in ci-mgmt, because we should be applying the CI config every time we update workflows, and either there's a bug, or a missing implementation here.~

Looks like we're just behind in updating the Workflows: https://github.com/pulumi/pulumi-aws/pull/2889/files

Prereq for fixing #2848.

### DIFF
--- a/.github/workflows/upgrade-bridge.yml
+++ b/.github/workflows/upgrade-bridge.yml
@@ -6,7 +6,7 @@ env:
 jobs:
   upgrade_provider:
     name: upgrade-provider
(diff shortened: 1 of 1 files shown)


guineveresaenger:
Closing in favor of https://github.com/pulumi/pulumi-aws/pull/2889/files.

PR #2892, opened 2023-10-16
Enforce no TODOs

### LINKED ISSUE #2885: Do not apply TODO docs replaces when the TODO is not fixed
## Hello!

## Issue details

We apply replaces for lines that contain "terraform", such as:

```json
  "website/docs/d/default_tags.html.markdown": [
    {
      "old": "With this data source, you can apply default tags to resources not _directly_ managed by a Terraform resource, such as the instances underneath an Auto Scaling group or the volumes created for an EC2 instance.",
      "new": "With this data source, you can apply default tags to resources not _directly_ managed by a TODO resource, such as the instances underneath an Auto Scaling group or the volumes created for an EC2 instance."
    }
  ]
```

We shouldn't apply any changes unless the `"new"` has the `"TODO"` replaced with a new value. That way replaces that have not been addressed will be omitted from our docs, instead of filled with confusing values.

iwahbe:
Fixes #2885

I removed all unused replacements from `replacements.json`, added a new replace rule to reduce the number of TODOs that needed to be manually fixed up. I then fixed up all remaining TODOs. https://github.com/pulumi/pulumi-aws/pull/2892/commits/324d355935664f036b95a7eac939a22ee8d7b5af prevents unaddressed TODOs from showing up in our docs.


sdk/dotnet/CleanRooms/ConfiguredTable.cs
```diff
@@ -32,15 +32,15 @@ namespace Pulumi.Aws.CleanRooms
...
+    ///         Description = "I made this table with Pulumi!",
     ///         TableReference = new Aws.CleanRooms.Inputs.ConfiguredTableTableReferenceArgs
     ///         {
     ///             DatabaseName = "example_database",
     ///             TableName = "example_table",
     ///         },
     ///         Tags = 
     ///         {
-    ///             { "Project", "TODO" },
+    ///             { "Project", "Pulumi" },
```
t0yv0, line 43: Nice

sdk/java/src/main/java/com/pulumi/aws/lightsail/inputs/BucketState.java
```diff
@@ -78,14 +78,14 @@ public Optional<Output<String>> createdAt() {
     }
 
     /**
-     * Force Delete non-empty buckets using `TODO destroy`. AWS by default will not delete an s3 bucket which is not empty, to prevent losing bucket data and affecting other resources in lightsail. If `force_delete` is set to `true` the bucket will be deleted even when not empty.
+     * Force Delete non-empty buckets using `pulumi down`. AWS by default will not delete an s3 bucket which is not empty, to prevent losing bucket data and affecting other resources in lightsail. If `force_delete` is set to `true` the bucket will be deleted even when not empty.
```
t0yv0, line 81: TIL pulumi down.

t0yv0 (approved):
Looks useful. Spot-checking example changes looks good to me.

sdk/java/src/main/java/com/pulumi/aws/lightsail/inputs/BucketState.java
iwahbe: That was a mistake. It should be `pulumi destroy`, but `pulumi down` does work.

PR #2894, opened 2023-10-17
Upgrade terraform provider aws to v5.21.0

danielrbradley:
Automation failed due to:
1. Conflict in patches within the upstream `go.mod`. Manually resolved but also, as suggested by @VenelinMartinov, moved the one added require line to be on its own to reduce future conflicts.
2. Manually mapped `aws_ec2_image_block_public_access`.


iwahbe (approved):
New patch looks fine.

What are we adding when we patch `go.mod`? Would it be possible to not patch `go.mod` at all?

danielrbradley:
@iwahbe ta, it's related to the legacy S3 bucket - we need an extra module referenced

iwahbe:
> @iwahbe ta, it's related to the legacy S3 bucket - we need an extra module referenced

Thanks for checking.

PR #2897, opened 2023-10-17
Upgrade pulumi terraform bridge to v3.62.0

mjeffryes:
generated with 
`upgrade-provider pulumi/pulumi-aws --target-bridge-version v3.62.0 --kind bridge`


mjeffryes:
Hmm...there seems to be some regression in docs with this change.

t0yv0:
TypeError: _configure() missing 1 required positional argument: 'compute_type' ? 

One second, I think this is something we blocked the 3.62.0 https://github.com/pulumi/pulumi-terraform-bridge/issues/1439 release on

t0yv0:
Let's investigate tomorrow. This could be another issue in p/p but would be good to eliminate the possibility that the test/example was somehow quietly incorrect until now.

t0yv0:
https://github.com/pulumi/pulumi/issues/14274 this is a platform bug.

iwahbe:
Closing in favor of https://github.com/pulumi/pulumi-aws/pull/2918.

PR #2899, opened 2023-10-18
Improve instructions when rebasing

danielrbradley:
When going into the `make upstream.rebase` it’s not clear that you have to run `git rebase --continue` yourself having resolved the conflict. If you don’t, `make upstream.finalize` just deletes half of the patches.

### DIFF
--- a/scripts/upstream.sh
+++ b/scripts/upstream.sh
@@ -45,12 +45,12 @@ message_rebase_clean() {
   cat <<EOF
 
 The full patch set has been cleanly applied to the './upstream' repository.
-To "edit" the patch set, commit changes directly to './upstream'. When
-you are done making changes, run:
+To "edit" the patch set, manipulate the commits on the 'pulumi-patch' branch
+in the './upstream' submodule. When you are done making changes, run:
 
     make upstream.finalize
 
-to finish the rebase and commit those changes into the patch set.
+to finish the rebase and convert these commits back into patches.
 
 EOF
 }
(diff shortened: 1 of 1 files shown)


scripts/upstream.sh
```diff
@@ -59,12 +59,12 @@ message_rebase_dirty() {
 cat <<EOF
 
 The patch set did not apply cleanly. You need to manually resolve
-rebase conflicts directly in './upstream'. When you have completed
-the rebase, run
+rebase conflicts directly in './upstream' and run 'git rebase --continue'.
+When you have completed the rebase, run
```
iwahbe, line 63: You might not want to run `git rebase --continue`. There are multiple relevant rebase commands (`git rebase --skip` comes to mind in this situation).

I understand wanting to add emphasis here, but we should be careful to not be ovecrperscriptive.

```suggestion
rebase conflicts directly in './upstream' finish the rebase.
In the simple case, this could mean running 'git rebase --continue'.
When you have completed the rebase, run
```

iwahbe (changes requested):
This file is generated from ci-mgmt (see line 2).

In addition to changing the help text, can we change `upstream.sh`'s semantics to error when `make rebase.finalize` is run in a dirty/rebasing upstream?

danielrbradley:
@iwahbe thanks for the feedback. I realised this is managed by ci-mgmt just after pushing. Ended up going down a little rabbit hole and so just opened up this ticket instead which I think will also lead to improving these messages when we can determine the state correctly:

- https://github.com/pulumi/ci-mgmt/issues/694

I'll close this out for now.

PR #2907, opened 2023-10-20
Add text to inform users about ESC on creds error

zchase:
Part of https://github.com/pulumi/marketing/issues/816

This PR adds some text to current text emitted when a user hits a credentials to let them know about ESC and that they can configure their AWS creds with ESC. When I was testing this locally I noticed the error emitted twice (same with a warning about using the aws plugin from my path) so not sure if that is expected with local dev. 

Example output: 

promote-esc-on-error-aws

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -617,7 +617,10 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
 	if _, _, diag := awsbase.GetAwsConfig(context.Background(), config); diag != nil && diag.HasError() {
 		return fmt.Errorf("unable to validate AWS credentials. \n"+
 			"Details: %s\n"+
-			"Make sure you have set your AWS region, e.g. `pulumi config set aws:region us-west-2`. \n", formatDiags(diag))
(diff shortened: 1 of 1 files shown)


lukehoban:
I would like to have a great page we can point to here specifically for configuring AWS OIDC + ESC. I believe that is in progress but not available yet. Suggest we hold on this till that lands, then link to that directly.

> When I was testing this locally I noticed the error emitted twice

Yeah - a number of existing issues with this error message tracked in https://github.com/pulumi/pulumi-aws/issues/2285.

joeduffy:
> I would like to have a great page we can point to here specifically for configuring AWS OIDC + ESC.

Can someone link to that work item? I'd personally bank Zach's change and then revise it once the new page is available, unless it is imminent (as in <48 hours). It would be awesome if it is!

iwahbe:
@zchase Are you still interested in getting this change in?

iwahbe (approved):

zchase:
@iwahbe Yup! This link currently is the most up to date resource for this. Once we have "guides" in our docs we will want to update this link and apply this change to some of our other most used providers.

Issue #2895, opened 2023-10-17
Incorrect tags applied when tag is empty


t0yv0:
### What happened?

Customer reports a regression where having an empty tag value causes all tags to disappear with 6.5.x version of the provider (I also reproduced on 6.2.x).

This program:

```go

package main

import (
	"fmt"

	"github.com/pulumi/pulumi-aws/sdk/v6/go/aws/s3"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

func main() {

	pulumi.Run(func(ctx *pulumi.Context) error {

		bucket, err := s3.NewBucketV2(ctx, "test-bucket-anton-4", &s3.BucketV2Args{
			Bucket:       pulumi.String("test-bucket-anton-4"),
			ForceDestroy: pulumi.Bool(false),
			Tags: pulumi.StringMap{
				"cust_app":       pulumi.String(""),
				"cust_ver":  pulumi.String("2020-04-29"),
				"cust_url": pulumi.String("https://cust.com/xywvBZgKsUsN"),
			},
		})

		fmt.Printf("%p\n", bucket)

		if err != nil {
			return err
		}

		ctx.Export("tags", bucket.Tags)
		return nil
	})
}

```

pulumi up produces an empty tags export. Commenting out `cust_app` with empty string makes it produce 2 tags.

### Example

See above.

### Output of `pulumi about`

```
CLI          
Version      3.86.0
Go Version   go1.21.1
Go Compiler  gc

Plugins
NAME  VERSION
aws   6.2.1
go    unknown

Host     
OS       darwin
Version  14.0
Arch     x86_64

This project is written in go: executable='/Users/t0yv0/bin/go' version='go version go1.21.0 darwin/amd64'

Current Stack: t0yv0/test-chris-bug/dev

TYPE                      URN
pulumi:pulumi:Stack       urn:pulumi:dev::test-chris-bug::pulumi:pulumi:Stack::test-chris-bug-dev
pulumi:providers:aws      urn:pulumi:dev::test-chris-bug::pulumi:providers:aws::default
aws:s3/bucketV2:BucketV2  urn:pulumi:dev::test-chris-bug::aws:s3/bucketV2:BucketV2::test-bucket-anton-4


Found no pending operations associated with dev

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/t0yv0
User           t0yv0
Organizations  t0yv0, pulumi
Token type     personal

Dependencies:
NAME                                 VERSION
[...]

PR #2944, opened 2023-10-31
Fixes empty tag handling

t0yv0:
Fixes https://github.com/pulumi/pulumi-aws/issues/2895

Requires a bridge change, currently still pre-release: https://github.com/pulumi/pulumi-terraform-bridge/compare/v3.63.2...1e955db434c6

Due to confusing empty values with missing values at several layers of the stack, working with empty tag values was problematic before this change both in the Pulumi provider and in the upstream provider; this is now fixed by:

- removing SetTagsDiff customizer from upstream; it is not working correctly for empty tags
- similarly, removing tags_all computation for Plugin Framework based resources
- completely taking over tagsAll handling from the TF layer to Pulumi layer; tagsAll is no longer marked as computed when working with TF (TF no longer sees it as computed), instead Pulumi computes tagsAll as a copy of tags and passes it down
- fixing a bridge bug that ignored adding an empty tag (incorrect DIFF_NONE)

This is change is now passing an aggressive random-sampled state transition matrix for various combinations of provider-level and resource-level tags.


t0yv0:
The test failures indicate that PF support is still required specially; and that this regressed on handling unknown and secrets in tags.

provider/go.mod
```diff
@@ -9,8 +9,8 @@ require (
 	github.com/hashicorp/terraform-provider-aws v1.60.1-0.20220923175450-ca71523cdc36
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/pulumi/providertest v0.0.2
-	github.com/pulumi/pulumi-terraform-bridge/pf v0.18.3
-	github.com/pulumi/pulumi-terraform-bridge/v3 v3.63.2
+	github.com/pulumi/pulumi-terraform-bridge/pf v0.18.4-0.20231103010134-1e955db434c6
```
iwahbe, line 12: Is this SHA on master or a branch in the bridge?

scripts/patch_computed_only.sh
iwahbe, line 1: When should this script be run?

iwahbe (commented):
This looks good to me. I have 2 quick questions then lets merge.

scripts/patch_computed_only.sh
t0yv0: That's a good question. I think I need to link it to the test. When upstream pushes more PF resources, a test will fail detecting more of the resources have computed tags_all; we'll need to patch these resources, and this script can help generating the patch. I'd love not to use patching for this problem, but I couldn't find a quick way to do that for PF so falling back to patches.

provider/go.mod
t0yv0: Hm, this should be on master AFAIK but I will double-check.

scripts/patch_computed_only.sh
iwahbe: `pulumi-aws` already has a bunch of in-code fixes:
https://github.com/pulumi/pulumi-aws/blob/c77690106ac0415901f302ea3c7bf307242fc5f3/Makefile#L134
Would it make sense to put these changes there instead of in a patch.

t0yv0: Indeed! That'd be better, while still safe via the test. I'll give it a try.

provider/resources.go
```diff
@@ -7179,16 +7179,20 @@ $ pulumi import aws:networkfirewall/resourcePolicy:ResourcePolicy example arn:aw
 		if _, ok := fields["tags_all"]; !ok {
 			fields["tags_all"] = &tfbridge.SchemaInfo{}
 		}
+
 		fields["tags_all"].Secret = &yes
 		fields["tags_all"].DeprecationMessage = "Please use `tags` instead."
 
+		// Upstream provider is edited to unmark tags_all as computed internally so that
+		// Pulumi provider internals can set it, but the user should not be able to set it.
+		fields["tags_all"].MarkAsComputedOnly = &yes
```
guineveresaenger, line 7188: TIL about `tfbridge.True()` that avoids having to construct a bool var just so we can reference it.

iwahbe: Most of my go code ends up with a `ref` function:

```go
func ref[T any](t T) *T { return &t }
```

provider/go.mod
t0yv0: Yes.

provider/tags_test.go
```diff
@@ -21,6 +21,9 @@ import (
 	"github.com/stretchr/testify/require"
 	"pgregory.net/rapid"
 
+	"github.com/hashicorp/terraform-provider-aws/shim"
+	"github.com/pulumi/pulumi-terraform-bridge/pf/tfbridge"
+	tfshim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
```
iwahbe, line 27:
```suggestion
	"pgregory.net/rapid"
	"github.com/pulumi/pulumi-terraform-bridge/pf/tfbridge"
	tfshim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"'
	
	"github.com/hashicorp/terraform-provider-aws/shim"
```

provider/tags_test.go
```diff
@@ -211,7 +214,123 @@ func TestApplyTags(t *testing.T) {
...
+	t.Run("plugin_framework", func(t *testing.T) {
+		p := tfbridge.ShimProviderWithContext(ctx, upstreamProvider.PluginFrameworkProvider)
+		p.ResourcesMap().Range(func(key string, res tfshim.Resource) bool {
+			tagsAll, ok := res.Schema().GetOk("tags_all")
+			if ok {
+				require.Falsef(t, tagsAll.Computed(), "tags_all is Computed for %s", key)
+			}
+			return true
+		})
+	})
```
iwahbe, line 335: Can we have a test for `Bucket`? It's one of the few resources that pulumi itself maintains and it has had tags bugs in the past.

t0yv0: Yeah that's a very good catch. I'll add rn.

t0yv0: Added. Passing locally 🤞

t0yv0:
@iwahbe PTAL. This is passing again.

iwahbe (approved):
LGTM

I think we will need to prioritize https://github.com/pulumi/pulumi-aws/issues/2962, since this patch seems relatively conflict prone. That said, the fix itself looks good.

t0yv0:
Yeah!

PR #3018, opened 2023-11-17
Use trimmed runtime metadata

mjeffryes:
Enable the trimmed metadata from https://github.com/pulumi/pulumi-terraform-bridge/pull/1524 to squeeze some milliseconds out of the provider init.

### Benchmarks 
```
❯ go test -bench "Provider$" -run "^$" -v -benchmem
goos: darwin
goarch: arm64
BenchmarkProvider
BenchmarkProvider-12           	       6	 188479375 ns/op	331332504 B/op	 2225300 allocs/op
BenchmarkRuntimeProvider
BenchmarkRuntimeProvider-12    	       8	 141554162 ns/op	331336897 B/op	 2225309 allocs/op
```
Note that these benchmarks are run with aliasing history generation enabled (we don't expose an api for disabling it to the provider), so they are slower than actual runtime performance.


iwahbe (approved):
LGTM

iwahbe:
@mjeffryes Is there any reason not to turn this flag on for our other large providers (azure, gcp)? If not I'll create issues to roll this benefit out.

mjeffryes:
> @mjeffryes Is there any reason not to turn this flag on for our other large providers (azure, gcp)? If not I'll create issues to roll this benefit out.

No reason not to try it.

PR #3021, opened 2023-11-17
adds support for aws arm64 batch jobs in v5

jaxxstorm:
This updates the AWS Go SDK to allow the container properties section in
the `aws.batch.JobDefinition` to support the ARM64 architecture.

This property is untyped, so it doesn't exist in the schema. Prior to
this version of the AWS Go SDK, the property would silently drop
unsupported fields, so this allows users to use these properties without
having to update to v6

PR #3025, opened 2023-11-18
V5 upgrades

t0yv0:
We need to be able to ship patch releases on top of the v5 branch which currently points to the latest release in use, v5.42.0. 

- [x] adopt latest Makefile and .ci-mgmt
- [x] convert repo from using an upstream fork to using ./patches
- [ ] upgrade to the latest bridge and codegen
- [ ] ensure tests run on v5 branch and branches pointing to it
- [ ] ensure releases can be made 
- [ ] ship https://github.com/pulumi/pulumi-aws/pull/3021 

Unfortunately there are some incompatibilities between the ./upstream and the latest bridge; upgrades up to v3.59.0 and v3.81.0 work OK but are missing some things. Notably this version-set wants to emit pulumix* files for Go, but the branch does not seem to have wafv2 type compression so that for the Go SDK compilation time and RAM requirements are quite excessive.


t0yv0:
Hm, the SAT problem on dependencies is really tricky here. Necessitates either back-porting https://github.com/pulumi/pulumi-aws/pull/2589 or forward-porting codegen to a version that does not emit pulumix. Otherwise we're looking at Go compile times in 30 min range.

PR #3031, opened 2023-11-20
Update AWS Go SDK

t0yv0:
Adapts https://github.com/pulumi/pulumi-aws/pull/3021 to make it releasable. Tests added as Python acceptance tests. CI configs tweaked to run the tests, they are now passing.

Original description below.

This updates the AWS Go SDK to allow the container properties section in
the aws.batch.JobDefinition to support the ARM64 architecture.

This property is untyped, so it doesn't exist in the schema. Prior to
this version of the AWS Go SDK, the property would silently drop
unsupported fields, so this allows users to use these properties without
having to update to v6


t0yv0:
Suggested next steps:

- merge to v5
- cut alpha from v5 to make sure the release pipeline works
- edit the release pipeline as needed, possibly increasing runner sizes
- when alpha publishes successfully, tag a patch version for the v5 series

t0yv0:
Looks like schema-tools output is irrelevant as it does not respect the base branch of the PR, comparing to master.

examples/examples_py_test.go
```diff
@@ -93,7 +97,52 @@ func TestAccFifoSqsQueuePy(t *testing.T) {
 			Dir:           filepath.Join(getCwd(t), "sqs-fifo-queue", "python"),
 			RunUpdateTest: false,
 			ExtraRuntimeValidation: func(t *testing.T, stack integration.RuntimeValidationStackInfo) {
-				assert.Contains(t, stack.Outputs["name"].(string), ".fifo")
+				expect := map[string]string{
```
t0yv0, line 100: This looks like an unintended edit, let me revert.

PR #3043, opened 2023-11-23
Update the interface for ECS Container PortMapping with current options

chan-vince:
Update the ECS Container PortMapping interface to add in some fields that are in the API but currently missing from the interface.  More context provided in the issue #2968.

I have done what felt obvious which was to update the interface. However I'm not really familiar with the full codegen(?) or dev process so I wouldn't be surprised if this is just the first step. Happy to continue further work with a bit of guidance :)


corymhall (approved):
@chan-vince thanks for the PR, sorry it took so long!

PR #3045, opened 2023-11-25
Correct CloudFront documentation

kema-dev:
Fixes CloudFront documentation accross SDKs for enabled, trusted sgner and trusted key group.

I took comments from [TrustedSigners](https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_TrustedSigners.html) and [TrustedKeyGroups](https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_TrustedKeyGroups.html) API reference, feel free to edit it as per your conventions.

PR #3046, opened 2023-11-26
Correct CloudFront functions available runtimes

kema-dev:
Correct CloudFront functions available runtimes as `cloudfront-js-2.0` is now available and works out of the box with Pulumi.
Branch is based on https://github.com/pulumi/pulumi-aws/pull/3045


iwahbe:
Similar to #3045, I believe that all the files you have edited are auto-generated.

PR #3047, opened 2023-11-27
Use pre-release bridge for more instrumentation

t0yv0:
Changes in the bridge dependency: https://github.com/pulumi/pulumi-terraform-bridge/compare/v3.66.0...2f132e750bb5

This change should enable more logs being captured through TF_LOG=trace troubleshooting, and add additional spans in --tracing shedding light on where the time is being spent around Configure and CheckConfig.


provider/resources.go
```diff
@@ -633,12 +634,21 @@ func preConfigureCallback(vars resource.PropertyMap, c shim.ResourceConfig) erro
 	// if we skipCredentialsValidation then we don't need to do anything in
 	// preConfigureCallback as this is an explicit operation
 	if skipCredentialsValidation {
+		log.Printf("[INFO] pulumi-aws: skip credentials validation")
 		return nil
 	}
 
 	var err error
 	credentialsValidationOnce.Do(func() {
+		log.Printf("[INFO] pulumi-aws: starting to validate credentials. " +
```
t0yv0, line 643: @VenelinMartinov adding these logs around credentials validation; log module should add timestamps I think.

PR #3070, opened 2023-11-30
Optimize startup performance

t0yv0:
Zooming in a lot on AWS Provider startup costs, noticing a recurring pattern. In SDKv2 we have schema.Resource that has Schema of type map but also SchemaFunc  The latter is designed specifically to defer / lazy instantiate schemas that are big. It looks like in AWS QuickSight and some other very large schemas make use of this. However over time transformations baked into upstream providers and our own bridge added up to a lot of passes that force-call these SchemaFunc() on startup, even if you don't use QuickSight. When done sequentially it all adds up. Moreover since there's no true laziness we end up paying the QuickSight etc cost multiple times.

This change does the following to avoid `SchemaFunc()` calls:

- turn off upstream tags check - this seems to be a validation-only measure, that somehow manages to call SchemaMap and SchemaFunc and slow things down; it now disabled in the bridged provider

- rewrite markTagsAllNotComputedForResources to avoid calling `SchemaMap()` eagerly; instead call on-demand and try to cache the call so QuickSight and friends are not initialized twice

### DIFF
--- a/patches/0031-Optimize-startup-performance.patch
+++ b/patches/0031-Optimize-startup-performance.patch
@@ -1,0 +1,163 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Anton Tayanovskyy <anton@pulumi.com>
+Date: Thu, 30 Nov 2023 14:28:37 -0500
+Subject: [PATCH 31/31] Optimize startup performance
+
+
+diff --git a/internal/provider/provider.go b/internal/provider/provider.go
+index 60554782a2..1017fd2bb6 100644
+--- a/internal/provider/provider.go
++++ b/internal/provider/provider.go
+@@ -306,7 +306,7 @@ func New(ctx context.Context) (*schema.Provider, error) {
+ 			interceptors := interceptorItems{}
+ 
+ 			if v.Tags != nil {
+-				schema := r.SchemaMap()
++				schema := schemaMapForTagsChecking(ctx, r, true)
+ 
+ 				// The data source has opted in to transparent tagging.
+ 				// Ensure that the schema look OK.
+@@ -383,7 +383,7 @@ func New(ctx context.Context) (*schema.Provider, error) {
+ 			interceptors := interceptorItems{}
+ 
+ 			if v.Tags != nil {
+-				schema := r.SchemaMap()
++				schema := schemaMapForTagsChecking(ctx, r, false)
+ 
+ 				// The resource has opted in to transparent tagging.
+ 				// Ensure that the schema look OK.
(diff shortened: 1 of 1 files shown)


t0yv0:
It's part of what got me to 60ms from 600ms on the plugin init but I don't have this handy in isolation. With the Benchmark would show the delta on the PR with https://github.com/benchmark-action/github-action-benchmark

t0yv0:
I guess one risk here is that we're disabling the tags checking. I can add a test that runs with tags checking enabled - but outside of the runtime provider..

iwahbe (approved):
I'd love a test that runs this without tags checking, but otherwise this looks good to me. We don't see any test failures, so I'm happy to assume that no part of the tags check is required.

---

### I ran a benchmark on this change: 203070767 to 130972106 ns/op. Great work!

### Methodology

I ran tests using from `/provider`.
```sh
(cd .. && make upstream) && go test -run XXX -bench BenchmarkProvider -count=10
```

#### Baseline (SHA 08345fd01d17774084f6a11412f5fc8eb66cdf7d)
```
goos: darwin
goarch: arm64
BenchmarkProvider-12                   5         202594883 ns/op
BenchmarkProvider-12                   5         202670817 ns/op
BenchmarkProvider-12                   5         202280158 ns/op
BenchmarkProvider-12                   5         203070767 ns/op
BenchmarkProvider-12                   5         201846733 ns/op
BenchmarkProvider-12                   5         202844192 ns/op
BenchmarkProvider-12                   5         203908467 ns/op
BenchmarkProvider-12                   5         204008425 ns/op
BenchmarkProvider-12                   5         203547117 ns/op
BenchmarkProvider-12                   5         203355017 ns/op
PASS
ok      github.com/pulumi/pulumi-aws/provider/v6        23.472s
```

#### On gh:pulumi/pulumi-aws#3070 (SHA 8b8166d7dfe9b16bb909ffa5b423f62953aa8d14)

```
goos: darwin
goarch: arm64
BenchmarkProvider-12                   8         133698875 ns/op
BenchmarkProvider-12                   8         133265208 ns/op
BenchmarkProvider-12                   9         125642042 ns/op
BenchmarkProvider-12                   9         125897708 ns/op
BenchmarkProvider-12                   9         126640361 ns/op
BenchmarkProvider-12                   9         125530806 ns/op
BenchmarkProvider-12                   9         125807250 ns/op
BenchmarkProvider-12                   9         126111014 ns/op
BenchmarkProvider-12                   9         125148741 ns/op
BenchmarkProvider-12                   9         130972106 ns/op
PASS
ok      github.com/pulumi/pulumi-aws/provider/v6        16.159s
```

PR #3074, opened 2023-12-01
CI tests for OIDC authentication

thomas11:
Test web identity (OIDC) authentication in CI for regression testing.

The approach is to pick one of the existing tests in `examples` and run it two additional times, authenticating via web identity/OIDC:
1. via the GH action `aws-actions/configure-aws-credentials`
2. manually via retrieving the token from GH and configuring the provider with it.

I don't love the duplication of much of the workflow definition for YAML but maybe that's the trade-off here for using ci-mgmt?

Related: #3084, pulumi/registry#3567


examples/examples_nodejs_test.go
```diff
@@ -120,6 +120,16 @@ func TestAccCloudWatch(t *testing.T) {
 	integration.ProgramTest(t, &test)
 }
 
+func TestAccCloudWatchOidcManual(t *testing.T) {
+	test := getJSBaseOptions(t).
+		With(integration.ProgramTestOptions{
+			Dir:           filepath.Join(getCwd(t), "cloudwatchOidcManual"),
+			RunUpdateTest: true,
```
t0yv0, line 127: Do you need RunUpdateTest? Does it work for you? Last time I spelunked there in this ProgramTest functionality it didn't really do anything useful.

provider/resources.go
```diff
@@ -581,11 +578,13 @@ func validateCredentials(vars resource.PropertyMap, c shim.ResourceConfig) error
 			SourceIdentity:    stringValue(details.ObjectValue(), "sourceIdentity", []string{}),
 			TransitiveTagKeys: arrayValue(details.ObjectValue(), "transitiveTagKeys", []string{}),
 		}
-		duration, err := durationFromConfig(details.ObjectValue(), "durationSeconds", []string{})
```
t0yv0, line 584: Ah so env vars were not used.

provider/resources.go
```diff
@@ -538,25 +538,22 @@ func arrayValue(vars resource.PropertyMap, prop resource.PropertyKey, envs []str
 	return vals
 }
 
-func durationFromConfig(vars resource.PropertyMap, prop resource.PropertyKey, envs []string) (time.Duration, error) {
+// returns a pointer so we can distinguish between a zero value and a missing value
```
t0yv0, line 541: Nit: slight preference of adding more Go files in under ./provider instead of growing "resources.go" it's very big already, but certainly not blocking on this!

.github/workflows/run-acceptance-tests.yml
```diff
@@ -439,6 +441,97 @@ jobs:
...
+          - name: Install dependencies
+            run: make install_${{ matrix.language}}_sdk
+          - name: Install gotestfmt
+            uses: GoTestTools/gotestfmt-action@v2
+            with:
+              token: ${{ secrets.GITHUB_TOKEN }}
+              version: v2.4.0
+          - name: Make upstream
+            run: make upstream
+          - name: Run selected tests with manual web identity/OIDC auth
```
t0yv0, line 511: So this works I guess but do we have any alternatives that don't create a brand new test target for this particular test to keep the workflows a little simper? 

What are the extra requirements for the test? I see you need OIDC_ROLE_ARN but perhaps this could have been exposed to all tests? Then you got `aws-actions/configure-aws-credentials@v4` that's assuming the role with `role-to-assume`. Is that the only way to do it? 

Can Pulumi program under test assume this role, perhaps by using explicit provider configuration?

If it can't, can the test harness take care of it in the Go code?

t0yv0 (approved):
Have a few suggestions to simplify but not blocking, his looks super useful.

examples/examples_nodejs_test.go
thomas11: I just copied the test above and changed the target directory. I'd leave it alone since whether it's useful here or not, it's unrelated to this PR.

.github/workflows/run-acceptance-tests.yml
thomas11: I made a copy of the `test` job because it runs `aws-actions/configure-aws-credentials` and I want to run it, too, but with a different configuration. If I could edit this file directly, I could just append the steps to the existing `test` job but in ci-mgmt, I can't.

This new OIDC test job runs two `go test` runs:
1. with manual OIDC configuration
2. with `aws-actions/configure-aws-credentials`

I believe I could port (1) to the existing test job. Since (1) proves that OIDC works, (2) mainly tests the OIDC part of `aws-actions/configure-aws-credentials` which we don't own. So I guess we could just drop it, although I would have liked to have an e2e test for a common user scenario.

t0yv0: Yup. It makes sense to have an e2e test for a common user scenario.

Issue #2252, opened 2022-12-07
Support for AssumeRoleWithWebIdentity


MMartyn:
### What happened?

The credential validation that is currently in here doesn't handle AssumeRoleWithWebIdentity. So if you use AssumeRoleWithWebIdentity then you currently need to skip the validation checks.

### Steps to reproduce

Attempt to use AssumeRoleWithWebIdentity

### Expected Behavior

It to seemlessly work regardless of if you are using keys, roles, or web identities.

### Actual Behavior

You get an error saying that the credentials couldn't be validated.

### Additional context

Relevant code: 
https://github.com/hashicorp/terraform-provider-aws/blob/46dbaeb5a40328b9a5200de13fa842700c5a1a35/internal/provider/provider.go#L2316
https://github.com/pulumi/pulumi-aws/blob/master/provider/resources.go#L340

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

roothorp:
Hi @MMartyn, thanks for the issue! I've raised this with the team to be prioritized and added to our work stack.

PR #3084, opened 2023-12-04
Add assume role with web identity, expand assume role properties in creds validation

thomas11:
- Add `assumeRoleWithWebIdentity` to creds validation #2252
- Add most missing properties to assumeRole credentials check
- Add duration to supported assumeRole properties

Resolves ##2252

The supported properties of both `assumeRole` and `assumeRoleWithWebIdentity` should now match [the ones documented in the registry](https://www.pulumi.com/registry/packages/aws/installation-configuration/#configuration-options).

The remaining exception is `tags` where I wasn't sure about the correct conversion from `PropertyMap` to Go `map`.

I have an e2e test for `assumeRoleWithWebIdentity` in progress as part of #3074 but wanted to get this change out there.

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -558,7 +558,18 @@ func validateCredentials(vars resource.PropertyMap, c shim.ResourceConfig) error
 			SessionName: stringValue(details.ObjectValue(), "sessionName", []string{}),
 		}
 		config.AssumeRole = &assumeRole
+	}
 
+	if details, ok := vars["assumeRoleWithWebIdentity"]; ok {
+		assumeRole := awsbase.AssumeRoleWithWebIdentity{
+			RoleARN:              stringValue(details.ObjectValue(), "roleArn", []string{}),
+			Policy:               stringValue(details.ObjectValue(), "policy", []string{}),
+			PolicyARNs:           arrayValue(details.ObjectValue(), "policyArns", []string{}),
+			SessionName:          stringValue(details.ObjectValue(), "sessionName", []string{}),
+			WebIdentityToken:     stringValue(details.ObjectValue(), "webIdentityToken", []string{}),
+			WebIdentityTokenFile: stringValue(details.ObjectValue(), "webIdentityTokenFile", []string{}),
+		}
+		config.AssumeRoleWithWebIdentity = &assumeRole
 	}
 
 	// By default `skipMetadataApiCheck` is true for Pulumi to speed operations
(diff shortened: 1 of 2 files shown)


provider/resources.go
```diff
@@ -550,15 +572,40 @@ func validateCredentials(vars resource.PropertyMap, c shim.ResourceConfig) error
...
+			WebIdentityToken:     stringValue(details.ObjectValue(), "webIdentityToken", []string{}),
+			WebIdentityTokenFile: stringValue(details.ObjectValue(), "webIdentityTokenFile", []string{}),
+		}
+		duration, err := durationFromConfig(details.ObjectValue(), "durationSeconds", []string{})
+		if err != nil {
+			return err
+		}
+		assumeRole.Duration = duration
 
+		config.AssumeRoleWithWebIdentity = &assumeRole
```
t0yv0, line 608: I am very confused here because reading this code prior to the change it appeared that `config` only lives long enough to satisfy 'validateCredentials'. Which in turn is only used to make sure we can connect to AWS, but is not actually used to connect to AWS (which in itself is of course very suspect and error-prone, that the two methods diverge).

Are you intentional about this? That is does your change target fixing credentials validation change only without intending to change the actual behavior of the provider past the validation change? Or perhaps I'm misunderstanding how this works?

t0yv0: Ah yes.. It sounds like that's what the ticket is asking for literally.

t0yv0: No it's on me, should have read the ticket!

t0yv0:
> The remaining exception is tags where I wasn't sure about the correct conversion from PropertyMap to Go map.

Curious where do you need to process tags? I can help with this transform if need. So far the transformations are reasonable as written.

t0yv0:
This seems to track closely ./upstream code for configuring the AWS client just in Pulumi terms. An alternative way forward would be trying to export that code by patching upstream and calling it somehow, but this is quire reasonable. I think with an lightweight E2E test this is LGTM, thank you for taking it on!

thomas11:
> > The remaining exception is tags where I wasn't sure about the correct conversion from PropertyMap to Go map.
> 
> Curious where do you need to process tags? I can help with this transform if need. So far the transformations are reasonable as written.

Actually, I don't think there's a point to read the tags for credentials validation.

guineveresaenger (approved):
Thank you for updating this!

PR #3089, opened 2023-12-05
Use correct EditDirs in test

thomas11:
In preparation for pulumi/pulumi#14695 which will cause such tests to break.

### DIFF
--- a/examples/examples_nodejs_test.go
+++ b/examples/examples_nodejs_test.go
@@ -301,17 +301,18 @@ func TestAccEcr(t *testing.T) {
 }
 
 func TestAccDeleteBeforeCreate(t *testing.T) {
+	basePath := filepath.Join(getCwd(t), "delete_before_create", "mount_target")
 	test := getJSBaseOptions(t).
 		With(integration.ProgramTestOptions{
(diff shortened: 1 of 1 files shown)


examples/examples_nodejs_test.go
```diff
@@ -301,17 +301,18 @@ func TestAccEcr(t *testing.T) {
 }
 
 func TestAccDeleteBeforeCreate(t *testing.T) {
+	basePath := filepath.Join(getCwd(t), "delete_before_create", "mount_target")
```
t0yv0, line 304: Why is this needed? In go relative paths seem to resolve relative to the test. Is there something that's now requiring getCwd prefix? Can we roll back that something?

t0yv0: Ah I see = never mind. Previously "step2" was interpreted as ./step2 but that was missing. After the change it's interpreted as "$PWD/delete_before_create/mount_target/step2" which probably exists.

This makes total sense then!

I'd drop getCwd(t) for simplicity but it was there before so that's fine.

t0yv0 (approved):

PR #3169, opened 2023-12-20
Stop embedding bridge-metadata.json

t0yv0:
At this point bridge-metadata.json is a compile-time only concern as it is not needed at runtime at all. This PR makes the change obvious and stops embedding it. This should help the binary size a bit, possibly to the tune of 30MB (uncompressed):

 31M    provider/cmd/pulumi-resource-aws/bridge-metadata.json
 88K    provider/cmd/pulumi-resource-aws/runtime-bridge-metadata.json

Only runtime-bridge-metadata.json is currently needed.

### DIFF
--- a/provider/cmd/pulumi-resource-aws/main.go
+++ b/provider/cmd/pulumi-resource-aws/main.go
@@ -29,6 +29,6 @@ var pulumiSchema []byte
 
 func main() {
 	ctx := context.Background()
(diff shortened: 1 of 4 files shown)


iwahbe (approved):
Looks good.

t0yv0:
Ah of course need to fix some tests here.

Issue #3020, opened 2023-11-17
Increased RAM usage in v6.x provider series


t0yv0:
### What happened?

Investigating a related performance regression I can confirm that v6 uses more RAM than v5. For example, provisioning 1000 KMS keys in the following program:

```python
import pulumi
import pulumi_aws as aws

for i in range(0, 1000):
    aws.kms.Key(f"key-{i}",
                deletion_window_in_days=7,
                description=f"KMS key {i}")
```

On v6.7.0 I am observing 300-400MB RAM peak aws-provider use.

On 5.42.0 the same program has 230MB RAM peak. 

The statefile sizes are about the same.

### Example

See above.

### Output of `pulumi about`

CLI          
Version      3.92.0
Go Version   go1.21.3
Go Compiler  gc

Plugins
NAME    VERSION
aws     5.42.0
python  unknown

Host     
OS       darwin
Version  14.0
Arch     x86_64

This project is written in python: executable='/Users/t0yv0/bin/python3' version='3.10.13'

Current Stack: t0yv0/repro-aws-lbriggs/dev

TYPE                  URN
pulumi:pulumi:Stack   urn:pulumi:dev::repro-aws-lbriggs::pulumi:pulumi:Stack::repro-aws-lbriggs-dev
pulumi:providers:aws  urn:pulumi:dev::repro-aws-lbriggs::pulumi:providers:aws::prov
pulumi:providers:aws  urn:pulumi:dev::repro-aws-lbriggs::pulumi:providers:aws::default_5_42_0
aws:kms/key:Key       urn:pulumi:dev::repro-aws-lbriggs::aws:kms/key:Key::key-1
aws:kms/key:Key       urn:pulumi:dev::repro-aws-lbriggs::aws:kms/key:Key::key-0
aws:kms/key:Key       urn:pulumi:dev::repro-aws-lbriggs::aws:kms/key:Key::key-2
aws:kms/key:Key       urn:pulumi:dev::repro-aws-lbriggs::aws:kms/key:Key::key-6
aws:kms/key:Key       urn:pulumi:dev::repro-aws-lbriggs::aws:kms/key:Key::key-5
aws:kms/key:Key       urn:pulumi:dev::repro-aws-lbriggs::aws:kms/key:Key::key-3
aws:kms/key:Key       urn:pulumi:dev::repro-aws-lbriggs::aws:kms/key:Key::key-7
[...]

PR #3172, opened 2023-12-21
Optimize RAM usage for the provider

t0yv0:
Fixes https://github.com/pulumi/pulumi-aws/issues/3020 

RAM usage is back to 5x-ish levels although I'm sure there is still some overhead for v6.

The key issue continues to be the large schemas for WAFv2 and QuickSight types. These are not materialized into maps but hidden away behind `SchemaFunc()` to be lazily accessible. Unfortunately machinery present both upstream and in our translation layer introspects resource schemas in multiple passes, causing `SchemaFunc()` calls repeatedly.

Prior to this PR in https://github.com/pulumi/pulumi-aws/pull/3070 we introduced a cache to the SchemaFunc calls which improved startup performance in spite of repeated invocations of SchemaFunc for expensive WAFv2 and QuickSight types. Unfortunately this change ended up costing significant memory and contributing to #3020. 

What this PR does is:

- remove the cache
- optimize resources.go passes to avoid calling SchemaFunc on startup
  - AutoNaming is inlined and schema query results coded for the largest types; correctness of the cache checked by test
  - AutoAliasing is now optimized via the bridge and does not cause SchemaFunc to be called at runtime
  - Tags schema fixup is treated similarly to AutoNaming
- verify via panic injection that the provider does not all SchemaFunc at startup


patches/0037-Do-not-cache-large-schemas.patch
```diff
@@ -0,0 +1,33 @@
...
+ 	if r.SchemaFunc != nil {
+ 		old := r.SchemaFunc
+-		var once sync.Once
+-		var cache map[string]*schema.Schema
+ 		u.SchemaFunc = func() map[string]*schema.Schema {
+-			once.Do(func() {
+-				cache = markTagsAllNotComputedForSchema(rn, old())
+-			})
+-			return cache
++			return markTagsAllNotComputedForSchema(rn, old())
```
iwahbe, line 30: Instead of adding a cache (`patches/0031-Optimize-startup-performance.patch`) and then removing the cache, can you edit `patches/0031-Optimize-startup-performance.patch` to not insert the cache in the first place?

iwahbe (commented):
I've been building an [_efficient_ schema traverser](https://github.com/pulumi/pulumi-terraform-bridge/pull/1590) for `pulumi-akamai`, inspired by your auto-alias implementation. If we re-implement `AutoNaming` and `Tags` with the new traverser, I think that we will get the same benefit here without the hard coding.

If possible, I'd like to keep optimizations in the bridge, and out of providers.

---

That said, https://github.com/pulumi/pulumi-terraform-bridge/pull/1590 hasn't merged yet and this will help customers now, so I'm willing to take this PR and then follow up with the bridge improvement. Let me know if we should expedite https://github.com/pulumi/pulumi-terraform-bridge/pull/1590 or you want to merge this strategy now and follow up later.

I left one comment on the patch to address before merging.

t0yv0:
Yeah this is essentially a P1 I'm told so done is better than perfect here. I think you're right we can do better. We can and should upstream AutoName and custom tags traversal speedup to the bridge machinery as time permits.

t0yv0:
PTAL, this is now editing the same patch.

Issue #2246, opened 2022-12-05
Pulumi import or refresh adding managedPolicyArns to aws.iam.Role


Fydon:
### What happened?

When I import `aws.iam.Role` or run `pulumi up --refresh pulumi`, Pulumi will populate `managedPolicyArns` even if you use `aws.iam.PolicyAttachment`, `aws.iam.RolePolicyAttachment` or `aws.iam.RolePolicy` to attach the same set of policies. [Pulumi warns that you should not use both](https://www.pulumi.com/registry/packages/aws/api-docs/iam/role/).

It is possible to use a resource transform or setting `managedPolicyArns: []` to remove `managedPolicyArns` but then running `pulumi up --refresh` results in policies being detached and then only maybe being reattached.

[I asked about this problem on Slack](https://pulumi-community.slack.com/archives/CRH5ENVDX/p1669899547439439), but the only responses I received were from someone who says that they are only a user and so I still feel that this is a bug. They say that the notes about `managedPolicyArns` are about a user setting it and that conflict does not exist if Pulumi sets it. I have never encountered any other application where the value of a property being set in a way that would cause a problem is not a problem if the user did not explicitly set the property to that value themselves.

### Steps to reproduce

- Create an `aws.iam.Role` setting `managedPolicyArns: []` and use `aws.iam.PolicyAttachment`, `aws.iam.RolePolicyAttachment` or `aws.iam.RolePolicy` to attach the same set of policies
- Run `pulumi up --refresh pulumi`
- Check Pulumi state to confirm if managedPolicyArns is still set to the empty set

### Expected Behavior

After performing the steps, Pulumi state to continues to have managedPolicyArns is set to the empty set.

### Actual Behavior

After performing the steps, Pulumi state has managedPolicyArns is set to contain any policies already in `aws.iam.PolicyAttachment`, `aws.iam.RolePolicyAttachment` or `aws.iam.RolePolicy`.

### Output of `pulumi about`

```
CLI
Version      3.47.1
Go Version   go1.19.2
Go Compiler  gc

Plugins
NAME    VERSION
aws     5.19.0
nodejs  unknown

Host     
OS       Microsoft Windows 11 Pro
Version  10.0.22000 Build 22000
Arch     x86_64

This project is written in nodejs: executable='C:\Program Files\nodejs\node.exe' version='v16.15.0'

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/Fydon
User           Fydon
Organizations  Fydon
```

[...]

PR #3183, opened 2023-12-26
Make integration tests stricter

t0yv0:
Prior to this change, all example-based acceptance tests inherited a relaxed configuration.

```go
	return integration.ProgramTestOptions{
		ExpectRefreshChanges: true,
		SkipRefresh:          true,
		Quick:                true,
	}
```

This was masking issues with unexpected diffs during `refresh` and any possible regressions as there provider evolves. After the change, only the test that need it are configured with these relaxed options and the code is cross-correlated to root cause issues where appropriate.

### DIFF
--- a/examples/examples_dotnet_test.go
+++ b/examples/examples_dotnet_test.go
@@ -1,4 +1,5 @@
 // Copyright 2016-2017, Pulumi Corporation.  All rights reserved.
+//go:build dotnet || all
 // +build dotnet all
 
 package examples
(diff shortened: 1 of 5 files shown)


t0yv0:
Looks like a fair number of tests does not refresh cleanly due to https://github.com/pulumi/pulumi-aws/issues/2246

t0yv0:
Another big issue affecting refresh tests is https://github.com/pulumi/pulumi/issues/6235

t0yv0:
https://github.com/pulumi/pulumi-aws/issues/3190 was concealed by lax testing options.

t0yv0:
https://github.com/pulumi/pulumi-aws/blob/master/examples/cloudwatchOidcManual/index.ts#L25 cycles with non-empty preview because the token changes between program invocations. Interesting question here on whether the token should persist into the statefile at all.

iwahbe (approved):
LGTM

PR #3191, opened 2023-12-27
Unpin pulumi cli 3

t0yv0:
Before this change, we used an outdated Pulumi CLI version to test the provider against. With this change we start using the latest 3.x release like most other bridged providers. While it would be my preference to have it pinned and automatically updating via PRs to move toward [Hermetic Builds](https://github.com/pulumi/ci-mgmt/issues/738) that's more work and for now this change is a quick improvement.

### DIFF
--- a/.ci-mgmt.yaml
+++ b/.ci-mgmt.yaml
@@ -33,11 +33,6 @@ runner:
   publish: pulumi-ubuntu-8core
   buildSdk: pulumi-ubuntu-8core
 actions:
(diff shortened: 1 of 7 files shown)


VenelinMartinov (approved):

t0yv0:
Eh. I pushed a commit to the wrong branch. Apologies, fixing.

iwahbe (approved):
LGTM

PR #3225, opened 2024-01-05
Attempt to build with a smaller runner

iwahbe:
I'm hoping that the larger runners are no longer necessary since https://github.com/pulumi/pulumi/issues/14344 closed.

This is purely a cost saving measure (see https://github.com/pulumi/home/issues/3244 (internal))


t0yv0:
Run `make ci-mgmt` and commit to take effect here?

t0yv0:
It's a no-go looks like.

iwahbe:
> It's a no-go looks like.

Looks like it.

Issue #2994, opened 2023-11-15
Add provided.al2023 lambda runtime


edoardocostantinidev:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

AWS introduced a [new](https://aws.amazon.com/blogs/compute/introducing-the-amazon-linux-2023-runtime-for-aws-lambda/) lambda runtime `provided.al2023`. If you try and set a lambda with that runtime you receive the following error:
```
expected runtime to be one of [nodejs nodejs4.3 nodejs6.10 nodejs8.10 nodejs10.x nodejs12.x nodejs14.x nodejs16.x java8 java8.al2 java11 python2.7 python3.6 python3.7 python3.8 python3.9 dotnetcore1.0 dotnetcore2.0 dotnetcore2.1 dotnetcore3.1 dotnet6 nodejs4.3-edge go1.x ruby2.5 ruby2.7 provided provided.al2 nodejs18.x python3.10 java17], got provided.al2023.
```

### Affected area/feature

missing provider option for lambda runtime.

mikhailshilkov:
The upstream change has shipped two hours ago, we'll probably get it next week https://github.com/hashicorp/terraform-provider-aws/issues/34360

jbvv1:
The required upstream change was released as part of v5.26.0 of the AWS Terraform provider on November 16, 2023: https://github.com/hashicorp/terraform-provider-aws/issues/34360#issuecomment-1815494950. When will Pulumi get updated to support the `provided.al2023` runtime?

PR #3261, opened 2024-01-15
Update aws.lambda.Runtime to latest AWS support

### LINKED ISSUE #3260: aws.lambda.Runtime enum is missing many language versions
Many of the latest runtime versions supported by Lambda are not in the enum we define:

https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

https://github.com/pulumi/pulumi-aws/blob/b079b701c6caf26945cae8c510e53874943ad329/provider/resources.go#L4772-L4798

Similarly, many runtimes are deprecated, but we have not marked them as such yet.

lukehoban:
Catches our enum up to date with the latest from https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html.

Also updates the deprecated runtimes to latest, including some which have been very recently deprecated (but in a few cases, are not yet erroring on create/update - though those are coming in the next few weeks).

Also fixes CallbackFunction in TS to ensure it can accept a normal `string` as well, to ensure that users are not prevented from passing not-yet-supported enum values in the future.  This aligns with the underlying `Function` type which accepts both.

Fixes #3260.
Fixes https://github.com/pulumi/pulumi-aws/issues/2994.


iwahbe (approved):
LGTM

PR #3292, opened 2024-01-19
Upgrade terraform provider aws to v5.32.0

t0yv0:
Fixes https://github.com/pulumi/pulumi-aws/issues/3253

This PR upgrades the provider to match upstream provider version v5.32.0.

List of unusual things that defeated upgrade-provider and required manual intervention:

- The patch for elbv2 target workaround in #2517 does not apply cleanly to new upstream, had to be redone; I've tested manually that 2517 does not reproduce on top of the new provider, but need to backlog adding a test for it; the relevant ticket is in https://github.com/pulumi/pulumi-aws/issues/2749

- Upstream is gradually removing AWS Go SDK v1. This affected legacy bucket resource that Pulumi retains in spite of upstream having it removed. This is now working again but required another invasive patch to undo upstream changes.

- New "verified permissions" module needed to be registered in resources.go to support automatic module inference.

- Several new Plugin Framework based resources required patching to support tags; this is slowing us down until we can implement https://github.com/pulumi/pulumi-aws/issues/2962 

- Run test of provider shim was breaking as it didn't get all the dependencies, now fixed

- Patch from https://github.com/pulumi/pulumi-aws/pull/3290 wouldn't apply because of wrong format (`make upstream.rebase` broken) - had to reformat it

PR #3296, opened 2024-01-22
Remove VPC from tags test matrix

t0yv0:
Nothing that the TestRandomTagsCombinationsGo test periodically flakes up, as in https://github.com/pulumi/pulumi-aws/pull/3292  on testing VPC resources. I've not established the root cause but it appears to be that the VPCs allocated by independent test instances collide. This is probably just not a good resource to test tag correctness on and I think we're not losing much coverage by simply removing it. Buckets tests still test the SDKv2 based resources.

### DIFF
--- a/examples/examples_go_test.go
+++ b/examples/examples_go_test.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws/session"
 	appconfigsdk "github.com/aws/aws-sdk-go/service/appconfig"
-	ec2sdk "github.com/aws/aws-sdk-go/service/ec2"
 	s3sdk "github.com/aws/aws-sdk-go/service/s3"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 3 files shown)


iwahbe (commented):
Can we consider introducing a new sdkv2 based resources? I think of `aws.s3.Bucket` as it's own thing, since we have implemented it and it uses a different interceptor framework then upstream maintained sdkv2 resources.

When implementing our tags solution, we noticed that we could get tests passing for sdkv2 resources, but not bucket and vice versa.

t0yv0:
Yeah but  s3.NewBucketV2 is a very ordinary sdkv2 based resource right?

iwahbe (approved):

iwahbe:
> Yeah but s3.NewBucketV2 is a very ordinary sdkv2 based resource right?

yep. My bad.

t0yv0:
I'll need to put some more work into this tomorrow to fix this test.

VenelinMartinov (approved):

PR #3302, opened 2024-01-23
Attempt to build SDKs on default runners

iwahbe:
Partially fixes https://github.com/pulumi/pulumi-aws/issues/3304.


iwahbe:
Unfortunately, pulumi-aws `test (go)` is still too big to run on default runners.

Issue #2624, opened 2023-07-18
Display error with API Documentation for the AWS Lambda Function Resource


toriancrane:
When accessing the [AWS Lambda Function resource page](https://www.pulumi.com/registry/packages/aws/api-docs/lambda/function/), there is a section called "CloudWatch Logging and Permissions" near the top of the page that I believe should be nested under the "Example Usage" section. 

It is the first item on the right hand navigation menu.
![image](https://github.com/pulumi/pulumi-aws/assets/19211559/efd6c791-51c2-4c17-ba23-cb01350bfc96)

It also seems to be missing the the language selector/tabbing functionality.

![image](https://github.com/pulumi/pulumi-aws/assets/19211559/4e534712-e79e-4db2-9478-e52e00a2611b)

thomas11:
Thank you for reporting this, @toriancrane. I agree it doesn't look quite right. I've asked some folks internally and will report back.

thomas11:
Likely part of pulumi/pulumi-terraform-bridge#1291. The examples are not detected correctly, which is also why the language chooser is missing.

guineveresaenger:
Summary of [Internal Discussion](https://pulumi.slack.com/archives/C06CC07N7L2/p1707245805710099):
- we should likely take a more straightforward approach and take upstream docs in the order they appear, providing language choosers for every code snippet. Joe also had [some comments in this issue](https://github.com/pulumi/registry/issues/3498).
- this means streamlining some of the "Example Usage" logic that is scattered across the bridge and the docsgen templates.

### Findings so far:

Our assumptions for upstream docs are too specialized and do not match reality.
  - we assume code snippets only exist in an "Example Usage" section
    - in fact we hard code an expectation that an "Example Usage" section exists with that name which is potentially brittle
  - we assume if there are more examples that they are prefixed with an H3 markdown header
  - we base ordering, docsgen, metrics, and display logic on the above assumption
  
---

[...]

PR #3410, opened 2024-02-09
Patch upstream doc to be an H3 and display as part of the Example Usage section

guineveresaenger:
Partial and temporary hack for https://github.com/pulumi/pulumi-aws/issues/2624.
Adds a patch that rewrites the "CloudWatch Logging and Permission" section as a nested H3 example, which gets rid of the inline code display at the top of out [Lambda Function landing page](https://www.pulumi.com/registry/packages/aws/api-docs/lambda/function/). 

Note:
Elides display the incorrect "Lambda retries" section. A full fix for #2624 will address this and bring it back as the [description-only section that it is](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function#lambda-retries).
This is a temporary hack and should be fully addressed via #2624, after which this patch should be removed.

The unrelated patch churn seems legitimate:
 - the deleted patch was included in an earlier patch 0036
 - the changed function title in 0022 is the patch anchor and does not affect our patch content.
---

Here's a screenshot of what the page should look like in the registry (please excuse the tiny image, but you can see the improvement)
Screenshot 2024-02-09 at 11 14 59 AM

### DIFF
--- a/patches/0001-Add-TagsSchemaTrulyComputed-definition.patch
+++ b/patches/0001-Add-TagsSchemaTrulyComputed-definition.patch
@@ -1,7 +1,7 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Daniel Bradley <daniel@pulumi.com>
 Date: Fri, 4 Nov 2022 16:49:08 +0000
-Subject: [PATCH 01/40] Add TagsSchemaTrulyComputed definition
+Subject: [PATCH 01/41] Add TagsSchemaTrulyComputed definition
 
 
 diff --git a/internal/tags/tags.go b/internal/tags/tags.go
--- a/patches/0002-Conns-user-agent.patch
+++ b/patches/0002-Conns-user-agent.patch
@@ -1,7 +1,7 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Daniel Bradley <daniel@pulumi.com>
 Date: Fri, 4 Nov 2022 17:03:31 +0000
-Subject: [PATCH 02/40] Conns user agent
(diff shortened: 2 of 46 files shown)


sean1588:
@guineveresaenger  thanks this looks great! 🎉  I am not too familiar with the code and how stuff works in this repo, so I'll leave that for others to approve, but I fully support doing this patch as a quick fix.

t0yv0:
Could we avoid using a patch here?

t0yv0:
Also, terribly sorry for the nasty merge but could we land https://github.com/pulumi/pulumi-aws/pull/3411 first 🙏

t0yv0:
You could move the markdown code from upstream into the repo proper, then you could link it to the resource via .Docs:

```
		pi := tfbridge.ProviderInfo{
			P:       shimv2.NewProvider(p),
			Name:    "azurerm",
			Version: "0.0.1",
			Resources: map[string]*tfbridge.ResourceInfo{
				"azurerm_web_pubsub_custom_certificate": {
					Tok:  "azure:webpubsub/customCertificate:CustomCertificate",
					Docs: &tfbridge.DocInfo{Markdown: md},
				},
			},
		}
```

To be fair, the patch has the advantage that if upstream keeps changing this doc keeps changing also. But perhaps this doc is so important we'd rather maintain in-repo?

I can approve the patch just offering for your alternative consideration. I'm not having a good time maintaining these patches, and was thinking we need to improve this workflow anyway to make it easier.

t0yv0 (approved):

guineveresaenger:
I prefer a patch here precisely so the doc doesn't get stale. 

Let's cut the release first and I'll rework this. Have a fantastic weekend @t0yv0 !

PR #3411, opened 2024-02-09
Upgrade terraform provider aws to v5.36.0

t0yv0:
Fixes https://github.com/pulumi/pulumi-aws/issues/3408

Upgrades upstream to 5.36.0

New resources:

- controltower/landingZone.LandingZone
- opensearchingest/pipeline.Pipeline
- redshift/dataShareAuthorization.DataShareAuthorization
- securitylake/customLogSource.CustomLogSource


guineveresaenger (approved):
LGTM - was wondering if the docs replacement should be fixed via patch instead but not sure.

PR #3497, opened 2024-02-27
Cascade-replace resources when provider region changes

t0yv0:
Fixes scenario described in https://github.com/pulumi/pulumi-terraform-bridge/issues/244#issuecomment-878683259

It is now possible to change region at the provider level, in which case Pulumi will replace all resources provisioned against that provider in the new region. Unlike Pulumi, TF proper does not work in this case and changing a region causes authentication failures.


iwahbe (approved):
LGTM

I think this will need a rebase to avoid compile errors.

t0yv0:
Yes waiting a bridge update and will rebase.

t0yv0:
Debugging the failing upgrade tests yields some interesting information. The essential failing test case for the bucket upgrade has this protocol call which confuses the region as changing even though the change is presumably from no region ("") to a given region. This is because the recorded statefile does not list region of the provider. 

```
{
  "method": "/pulumirpc.ResourceProvider/DiffConfig",
  "request": {
    "urn": "urn:pulumi:test::test::pulumi:providers:aws::provider",
    "olds": {
      "version": "5.42.0"
    },
    "news": {
      "region": "us-west-2",
      "skipCredentialsValidation": "false",
      "skipMetadataApiCheck": "true",
      "skipRegionValidation": "true",
      "version": "5.42.0"
    },
    "oldInputs": {
      "version": "5.42.0"
    }
  },
  "response": {
    "replaces": [
      "region"
    ],
    "changes": "DIFF_SOME",
    "diffs": [
      "skipRegionValidation",
      "skipCredentialsValidation",
      "skipMetadataApiCheck"
    ],
    "detailedDiff": {
      "region": {
        "kind": "ADD_REPLACE",
        "inputDiff": true
      },
      "skipCredentialsValidation": {
        "inputDiff": true
      },
      "skipMetadataApiCheck": {
        "inputDiff": true
      },
      "skipRegionValidation": {
        "inputDiff": true
      }
    },
    "hasDetailedDiff": true
  },
  "metadata": {
    "kind": "resource",
    "mode": "client",
    "name": "aws"
  }
}

```

I am able to get the test to pass by deleting the recordings and re-recording the bucket test data. It appears that the new version now records region in the statefile. The same provider (5.42.0) is used as last time! Something must have changed in the way Pulumi CLI handles it.

provider/testdata/recorded/TestProviderUpgrade/bucket/5.42.0/stack.json
```diff
@@ -2,68 +2,76 @@
...
+        "modified": "2024-04-15T19:24:27.537569Z"
       },
       {
         "urn": "urn:pulumi:test::test::pulumi:providers:aws::provider",
         "custom": true,
-        "id": "3844c049-dd3a-4d9a-b4db-f476d1680642",
+        "id": "263eab09-759e-4d0c-8767-a00401d0c200",
         "type": "pulumi:providers:aws",
         "inputs": {
+          "region": "us-west-2",
```
t0yv0, line 29: The new CLI records region here. The provider version is still 5.42.0.

provider/testdata/recorded/TestProviderUpgrade/bucket/5.42.0/stack.json
```diff
@@ -2,68 +2,76 @@
   "version": 3,
   "deployment": {
     "manifest": {
-      "time": "2023-10-05T17:01:10.660615-04:00",
-      "magic": "cdb46f20d3397474c4b7038ce63a3c40fb0da4b95458ef275b6ce689aec88d17",
-      "version": "v3.86.0"
+      "time": "2024-04-15T15:24:34.499732-04:00",
+      "magic": "0e109972342b6ebd571fa6d2cffefcb95394f9c4f7bdd855463082d16126072c",
+      "version": "v3.111.1"
```
t0yv0, line 7: Last recording was on v3.86.0 of the CLI.

t0yv0:
Logged the blocker here https://github.com/pulumi/pulumi-aws/issues/3826

PR #3498, opened 2024-02-27
Update bridge to v3.76.0

t0yv0:
This is a routine manual update since the automated update started failing.


provider/tags_test.go
```diff
@@ -299,7 +299,9 @@ func TestAddingEmptyTagProducesChangeDiff(t *testing.T) {
 	    },
 	    "response": {
 	      "changes": "DIFF_SOME",
-              "hasDetailedDiff": true
+              "hasDetailedDiff": true,
+              "diffs": "*",
```
t0yv0, line 303: This now shows tags/tags_all in detailed diff, looks like an improvement.

PR #3587, opened 2024-03-07
fix: role inline policies fail to create without a policy name

### LINKED ISSUE #1949: iam.RoleInlinePolicy not working without `name` property (maybe `name` should be required)
### What happened?

I wanted to create S3FullAccessRole using `iam.Role`.

As official docs said that `name` property of interface `iam.RoleInlinePolicy` is optional, I let it optional, and only set the value of `policy` property. However, there was no creation of inlinePolicies in actual output. It seems like pulumi pass the empty string value to `policy` property even if it has proper values in the code.

When I set the value of `name` property, then it works as expected. (see below update preview)
image

### Steps to reproduce

```
import * as aws from '@pulumi/aws';

const role = new aws.iam.Role('example', {
  assumeRolePolicy: aws.iam.getPolicyDocumentOutput({
    statements: [
      {
        actions: ['sts:AssumeRole'],
        principals: [
          {
            type: '*',
            identifiers: ['*'],
          },
        ],
      },
    ],
  }).json,
  inlinePolicies: [
    {
      policy: aws.iam.getPolicyDocumentOutput({
        statements: [
          {
            actions: ['s3:*'],
            resources: ['*'],
          },
        ],
      }).json,
    },
  ],
});
```

### Expected Behavior

inlinePolicies is set to 
```
[
  {
    "name": "test",
    "policy": "{\n  \"Version\": \"2012-10-17\",\n  \"Statement\": [\n    {\n      \"Sid\": \"\",\n      \"Effect\": \"Allow\",\n      \"Action\": \"s3:*\",\n      \"Resource\": \"*\"\n    }\n  ]\n}"
  }
]
```

### Actual Behavior

inlinePolicies is set to 
```
[
  {
    "name": "",
    "policy": ""
  }
]
```

### Versions used

```
CLI          
Version      3.28.0
Go Version   go1.18
Go Compiler  gc

Plugins
NAME    VERSION
aws     4.35.0
docker  3.1.0
nodejs  unknown
tls     4.3.0

Host     
OS       darwin
Version  12.2.1
Arch     arm64
```

### Additional context

_No response_

### Contributing

corymhall:
In order to create an `inline_policy` on a role, both the `policy` and the `name` fields are required. If only one is provided then the policy will not be created (and no error will be thrown). Both properties must be optional though to allow providing empty values, i.e.

```ts
inlinePolicies: []
// or
inlinePolicies: [{}]
```

This PR sets the `name` property to be auto named and adds a transform on `inline_policy` to only create policies if a `policy` is provided.

fixes #1949

### DIFF
--- a/examples/examples_nodejs_test.go
+++ b/examples/examples_nodejs_test.go
@@ -626,7 +626,7 @@ func TestRoleInlinePolicyAutoName(t *testing.T) {
 	res, err := test.CurrentStack().Up(test.Context())
 	require.NoError(t, err)
 
-	policy := res.Outputs["inline_policy"]
+	policy := res.Outputs["inlinePolicy"]
 	value, err := json.Marshal(policy.Value)
 	require.NoError(t, err)
 
@@ -634,9 +634,9 @@ func TestRoleInlinePolicyAutoName(t *testing.T) {
 	err = json.Unmarshal(value, &inlinePolicy)
 	require.NoError(t, err)
 
-	policyEmpty := res.Outputs["inline_policy_empty"]
+	policyEmpty := res.Outputs["inlinePolicyEmpty"]
 
 	require.Equal(t, policyEmpty.Value, []interface{}{})
 	require.Regexp(t, regexp.MustCompile("testrole-*"), inlinePolicy.Name)
-	require.Equal(t, "{\n  \"Version\": \"2012-10-17\",\n  \"Statement\": [\n    {\n      \"Effect\": \"Allow\",\n      \"Action\": \"s3:GetObject\",\n      \"Resource\": \"*\"\n    }\n  ]\n}", inlinePolicy.Policy)
+	require.JSONEq(t, `{"Version": "2012-10-17", "Statement": [{"Effect": "Allow", "Action": "s3:GetObject", "Resource": "*" }]}`, inlinePolicy.Policy)
 }
--- a/examples/examples_nodejs_test.go
+++ b/examples/examples_nodejs_test.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"regexp"
 	"testing"
 	"time"
 
@@ -353,7 +354,8 @@ func TestAccIgnoreChanges(t *testing.T) {
 						assert.Equal(t, "foo", info.Deployment.Resources[2].Outputs["bucketPrefix"])
 					},
 				},
-			}})
+			},
+		})
 
 	integration.ProgramTest(t, &test)
 }
@@ -543,9 +545,9 @@ func TestRegress2555(t *testing.T) {
 func TestRegress3421(t *testing.T) {
 	test := getJSBaseOptions(t).
 		With(integration.ProgramTestOptions{
-				Dir: "regress-3421",
-				SkipRefresh: true,
-			},
+			Dir:         "regress-3421",
+			SkipRefresh: true,
+		},
 		)
 	// Disable envRegion mangling
 	test.Config = nil
@@ -581,7 +583,7 @@ func getAwsSession(t *testing.T) *session.Session {
 }
 
 func TestUpdateImportedLambda(t *testing.T) {
-	test := pulumitest.NewPulumiTest(t, "lambda-import-ts", 
+	test := pulumitest.NewPulumiTest(t, "lambda-import-ts",
 		opttest.LocalProviderPath("aws", filepath.Join(getCwd(t), "..", "bin")),
 	)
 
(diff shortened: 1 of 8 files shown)


examples/examples_nodejs_test.go
```diff
@@ -543,9 +545,9 @@ func TestRegress2555(t *testing.T) {
 func TestRegress3421(t *testing.T) {
 	test := getJSBaseOptions(t).
 		With(integration.ProgramTestOptions{
-				Dir: "regress-3421",
-				SkipRefresh: true,
-			},
+			Dir:         "regress-3421",
+			SkipRefresh: true,
```
t0yv0, line 549: We're missing something in CI to ensure checked in files are gofmt'd seems like.

corymhall: Should we update this to run on PRs? https://github.com/pulumi/pulumi-aws/blob/master/.github/workflows/lint.yml

provider/resources.go
```diff
@@ -2469,6 +2469,34 @@ func ProviderFromMeta(metaInfo *tfbridge.MetadataInfo) *tfbridge.ProviderInfo {
...
+						//   inlinePolicies: []
+						// or a list with empty objects
+						//   inlinePolicies: [{}]
+						// In both cases the role will be created _without_ any inline policies attached.
+						// If a policy is provided, then both the `policy` and the `name` fields are required.
+						// If one is provided and the other is not, then no error will be thrown and no inline policy
+						// will be created.
+						Transform: func(pv resource.PropertyValue) (resource.PropertyValue, error) {
+							inlinePolicy := []resource.PropertyValue{}
+							for _, value := range pv.ArrayValue() {
```
iwahbe, line 2483: `pv` could be a computed value, which would cause this transform to panic. We always need to do a type check (`.IsArray()`) before casting (`.ArrayValue()`) since any value could be computed.

iwahbe (commented):
This looks great. One fix we need to address before merging.

provider/resources.go
corymhall: Done!

examples/examples_nodejs_test.go
t0yv0: That'd be great! Tickets of this nature can go in https://github.com/pulumi/ci-mgmt

examples/role-inline-policy-auto-name/index.ts
```diff
@@ -0,0 +1,63 @@
...
+            identifiers: ['lambda.amazonaws.com'],
+          },
+        ],
+      },
+    ],
+  }).json,
+  inlinePolicies: [{}],
+});
+
+export const inline_policy = role.inlinePolicies[0];
```
t0yv0, line 62: Style nit: inlinePolicy casing preferable in TypeScript I think.

provider/resources.go
```diff
@@ -2469,6 +2469,36 @@ func ProviderFromMeta(metaInfo *tfbridge.MetadataInfo) *tfbridge.ProviderInfo {
 						AltTypes:  []tokens.Type{awsType(iamMod, "documents", "PolicyDocument")},
 						Transform: tfbridge.TransformJSONDocument,
 					},
+					"inline_policy": {
+						// inline_policy is an array of policy objects. It is allowed to provided an empty list
```
t0yv0, line 2473: Grammar: "allowed to be provided as an empty list" or something like this, sounds odd as written.

provider/resources.go
```diff
@@ -2469,6 +2469,36 @@ func ProviderFromMeta(metaInfo *tfbridge.MetadataInfo) *tfbridge.ProviderInfo {
...
+						//   inlinePolicies: []
+						// or a list with empty objects
+						//   inlinePolicies: [{}]
+						// In both cases the role will be created _without_ any inline policies attached.
+						// If a policy is provided, then both the `policy` and the `name` fields are required.
+						// If one is provided and the other is not, then no error will be thrown and no inline policy
+						// will be created.
+						Transform: func(pv resource.PropertyValue) (resource.PropertyValue, error) {
+							if pv.IsArray() {
```
t0yv0, line 2482: This filtering here is interesting so it looks like it will drop `{}` and `{"policy": ""}`  and proceed as if they were not there. To clarify, we certainly want this behavior?

examples/examples_nodejs_test.go
```diff
@@ -604,10 +606,37 @@ func TestUpdateImportedLambda(t *testing.T) {
...
+
+	var inlinePolicy InlinePolicy
+	err = json.Unmarshal(value, &inlinePolicy)
+	require.NoError(t, err)
+
+	policyEmpty := res.Outputs["inline_policy_empty"]
+
+	require.Equal(t, policyEmpty.Value, []interface{}{})
+	require.Regexp(t, regexp.MustCompile("testrole-*"), inlinePolicy.Name)
+	require.Equal(t, "{\n  \"Version\": \"2012-10-17\",\n  \"Statement\": [\n    {\n      \"Effect\": \"Allow\",\n      \"Action\": \"s3:GetObject\",\n      \"Resource\": \"*\"\n    }\n  ]\n}", inlinePolicy.Policy)
```
t0yv0, line 641: Consider https://pkg.go.dev/github.com/stretchr/testify/assert#JSONEq here

examples/role-inline-policy-auto-name/index.ts
```diff
@@ -0,0 +1,63 @@
...
+        principals: [
+          {
+            type: 'Service',
+            identifiers: ['lambda.amazonaws.com'],
+          },
+        ],
+      },
+    ],
+  }).json,
+  inlinePolicies: [{}],
```
t0yv0, line 59: This is a great example, I guess with this change this will be equivalent to `inlinePolicies: []`. It looks kind of clunky. To double check, the motivation for this is that we need to support programs that are out there that are already written like this, or some other motivation? For users writing new code do we want to direct them to writing `inlinePolicies: []` or even omitting  `inlinePolicies` entirely?

t0yv0 (approved):
Approving - I can see how AutoName fixes the bug as stated, was a bit less clear on the need for Transform filtering but it must have a reason. Left some questions just in case.

examples/role-inline-policy-auto-name/index.ts
corymhall: Yep that's correct. I can't think of why as a developer you would want to provide an empty policy like that, but I've learned the hard way that there will probably be someone out there that does it.

provider/resources.go
corymhall: Yeah I think this is the behavior that we want. If you provide `{"policy": ""}` it won't create the policy in AWS, but with the auto name you will end up with `{"policy": "", "name": "example-abcd"}`. This is also not valid and will end up with no policy being created in AWS (but it will show in the diff which is confusing to users).

The only valid policy is with _both_ `name` and `policy` and for backwards compatibility reasons we have to allow for users to specify `[{}]`

provider/resources.go
```diff
@@ -2469,6 +2469,36 @@ func ProviderFromMeta(metaInfo *tfbridge.MetadataInfo) *tfbridge.ProviderInfo {
...
+								inlinePolicy := []resource.PropertyValue{}
+								for _, value := range pv.ArrayValue() {
+									if value.IsObject() {
+										policy := value.ObjectValue()
+										if policy.HasValue("policy") && policy["policy"].StringValue() != "" {
```
iwahbe, line 2487: `policy["policy"]` can also be computed.

iwahbe (approved):
1 last thing to address, then LGTM!

PR #3651, opened 2024-03-18
Add EKS service principal for Node.js SDK

auvred:
`eks.amazonaws.com` service principal seems to be missing

https://docs.aws.amazon.com/eks/latest/userguide/using-service-linked-roles-eks.html#service-linked-role-permissions-eks

### DIFF
--- a/sdk/nodejs/iam/principals.ts
+++ b/sdk/nodejs/iam/principals.ts
@@ -140,6 +140,11 @@ export module Principals {
      * Service Principal for Edge Lambda
      */
     export const EdgeLambdaPrincipal: Principal = {Service: "edgelambda.amazonaws.com"};
+    
+    /**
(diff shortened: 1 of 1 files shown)


auvred:
> Unfortunately the code edited here is auto-generated and we cannot accept the change as is

I checked the file history (https://github.com/pulumi/pulumi-aws/commits/master/sdk/nodejs/iam/principals.ts), and it looks like the changes that were made there were not synced with other language SDKs. Also, I couldn't find a similar `principals` module in other SDKs.

Thus, it seems that this file is not autogenerated. Please correct me if I am wrong!

t0yv0:
You are right, my mistake, this seems to be a Node SDK overlay that is only available for Node users but is maintained manually. Your correction makes sense to include then! Thank you.

t0yv0 (approved):

Issue #3637, opened 2024-03-14
Interesting flake on TestAccDeleteBeforeCreate


t0yv0:
Interesting flake on TestAccDeleteBeforeCreate:

```  error: 1 error occurred:
  	* Retrieving AWS account details: validating provider credentials: retrieving caller identity from STS: 
  	operation error STS: GetCallerIdentity, https response error
        StatusCode: 403, RequestID: 8b1a26fa-db29-43dd-b705-869eadafa74c,
        api error ExpiredToken: The security token included in the request is expired
```

t0yv0:
https://github.com/pulumi/pulumi-aws/issues/3655

guineveresaenger:
Possibly a misconfiguration on our part?
[SO troubleshooting topicwith similar issues](https://stackoverflow.com/questions/73399249/aws-terraform-error-error-configuring-terraform-aws-provider-error-validati)
But the region is set globally for our tests.

[This cron](https://github.com/pulumi/pulumi-aws/actions/runs/8313400173/job/22749856771)  is flaking on the STS GetCallerIdentity credentials verification:
```
 Retrieving AWS account details: validating provider credentials: retrieving caller identity from STS: operation error STS: GetCallerIdentity, https response error StatusCode: 403, RequestID: <redacted>, api error ExpiredToken: The security token included in the request is expired
 ```
[This is the same error throughout on the flakes](https://github.com/pulumi/pulumi-aws/actions/runs/8331501584/job/22805609347#step:27:642)

What is interesting is that this test _never_ flakes on `master`.

I deleted the EC2 instance that failed to be deleted in our last run.

I wonder if it's an [expiration issue](https://repost.aws/knowledge-center/sts-iam-token-expired) since node tests take just a little over an hour to run?

**Comment by guineveresaenger** [comment]
UPDATE: this is now [flaking on pull requests as well](https://github.com/pulumi/pulumi-aws/actions/runs/8333514081/job/22806982878?pr=3656#step:28:496).

**Comment by t0yv0** [comment]
Flakes on master in https://github.com/pulumi/pulumi-aws/issues/3636 fwiw.. Not sure what's going on here.

**Comment by guineveresaenger** [comment]
Waiting on a runner is what is eating up most of the authenticated time:
```
2024-03-20T18:07:34.4133831Z Requested labels: ubuntu-latest
[...]

PR #3664, opened 2024-03-20
Increase parallelism to avoid credentials timeout test flake

guineveresaenger:
Potential fix for #3637.

This will have to go through CI management if it works as expected but the hope is that by increasing the parallelism, our tests will run faster so that our credentials do not time out.


guineveresaenger:
It appears that the real issue here is waiting for an available runner; not the amount of time that the tests take.

PR #3666, opened 2024-03-20
run unit tests in separate job

guineveresaenger:
This pull request adds the `provider/` unit tests as a separate job via `.ci-mgmt.yaml` ExtraTests config.

It only runs python and nodejs matrix, as there are currently no other unit tests enabled that need other languages. I'm happy to adjust this as necessary.

This refactor is meant to address the AWS credentials timeout that causes #3637.

### DIFF
--- a/.ci-mgmt.yaml
+++ b/.ci-mgmt.yaml
@@ -200,8 +200,10 @@ extraTests:
         language:
         - nodejs
 
-  unit_test:
-    name: unit_test
+  provider_test:
+    if: github.event_name == 'repository_dispatch' ||
+      github.event.pull_request.head.repo.full_name == github.repository
+    name: provider_test
     needs: build_sdk
     permissions:
       contents: read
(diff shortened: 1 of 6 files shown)


.ci-mgmt.yaml
```diff
@@ -202,3 +199,106 @@ extraTests:
...
+        name: Notify Slack
+        uses: 8398a7/action-slack@v3
+        with:
+          author_name: Failure in running ${{ matrix.language }} unit tests
+          fields: repo,commit,author,action
+          status: ${{ job.status }}
+    strategy:
+      fail-fast: false
+      matrix:
+        language:
```
t0yv0, line 302: Almost guaranteed that we will forget to edit this when adding go tests. I'd prefer a full matrix.

.github/workflows/master.yml
```diff
@@ -297,6 +297,7 @@ jobs:
       - license_check
       - go_test_shim
       - test_oidc
+      - unit_test
```
t0yv0, line 300: These are not unit tests AFAIK. These are the provider integration tests that are not examples.

t0yv0 (approved):
Looks good modulo nits.

Issue #3686, opened 2024-03-22
Understand why TestRegress1504 does not refresh cleanly


t0yv0:
### What happened?

Introducing a test with SkipRefresh https://github.com/pulumi/pulumi-aws/compare/t0yv0/fix-1504 because refresh has changes. This is typically benign but might be interesting to investigate. This is a LaunchTemplate resource. 

### Example

```typescript
import * as aws from "@pulumi/aws";
const ltemplate = new aws.ec2.LaunchTemplate(
    "my-launch-template",
    {
        imageId: "ami-0c55b159cbfafe1f0",
        instanceType:"t2.micro",
        keyName: 'key1',
        tags: {"Name": 'tag2'},
        updateDefaultVersion: true,
    });
```

Refresh diff says no changes but if pressed for details provides some:

```
No resources will be modified as part of this refresh; just your stack's state will be.
 details
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev::aws-3686::pulumi:pulumi:Stack::aws-3686-dev]
    ~ aws:ec2/launchTemplate:LaunchTemplate: (update)
        [id=lt-047219347b109768c]
        [urn=urn:pulumi:dev::aws-3686::aws:ec2/launchTemplate:LaunchTemplate::my-launch-template]
        [provider=urn:pulumi:dev::aws-3686::pulumi:providers:aws::default_6_32_0::1866b374-0cd7-4789-8430-daadd9035888]
        --outputs:--
      + securityGroupNames               : []
      + vpcSecurityGroupIds              : []

```

This fails (which is also why we noticed this in ProgramTest):

```
pulumi refresh --expect-no-changes --yes                                                                                                 ~/bugs/aws-3686
Previewing refresh (dev)

View in Browser (Ctrl+O): https://app.pulumi.com/anton-pulumi-corp/aws-3686/dev/previews/d5d294f9-a1c4-462b-8f5c-95a4782d45a4

     Type                       Name                Plan       Info
     pulumi:pulumi:Stack        aws-3686-dev                   
 ~   └─ aws:ec2:LaunchTemplate  my-launch-template  update     [diff: +securityGroupNames,vpcSecurityGroupIds]

Resources:
    ~ 1 to update
    1 unchanged

Refreshing (dev)

View in Browser (Ctrl+O): https://app.pulumi.com/anton-pulumi-corp/aws-3686/dev/updates/6

     Type                       Name                Status              Info
[...]

PR #3687, opened 2024-03-22
Fix updating tags on aws_launch_template

### LINKED ISSUE #1504: aws:ec2:LaunchTemplate update_default_version Fails on AWS Provider
I upgraded the AWS provider from 3.23 to the latest 4.5.1 recently to do a Pulumi 3.x upgrade. Noticed some code that sets `update_default_version` on this provider fails now when doing a pulumi update on the existing resources.

## Expected behavior
No changes should have been seen by the provider and updateDefaultVersion should continue to function as desired.

## Current behavior
```
  aws:ec2:LaunchTemplate (ci-general):
    error: 1 error occurred:
    	* updating urn:pulumi:ci::aws-eks-py::aws:ec2/launchTemplate:LaunchTemplate::ci-general: 1 error occurred:
    	* error modifying Launch Template (lt-<REDACTED>): InvalidParameterValue: Invalid launch template version: either '$Default', '$Latest', or a numeric version are allowed.
```

also, it doesn't even show up in the list of diffSet, so something tells me the provider is misbehaving internally here.

```
     pulumi:pulumi:Stack        aws-eks-py-ci   **failed**              1 error
 ~   ├─ aws:ec2:LaunchTemplate  ci-kiam-server  **updating failed**     [diff: -tagsAll~tags]; 1 error
 ~   └─ aws:ec2:LaunchTemplate  ci-general      **updating failed**     [diff: -tagsAll~tags]; 1 error
```

## Steps to reproduce
We are using the Python dialect. Create a `pulumi_aws.ec2.LaunchTemplate` resource and set the `update_default_version` attribute to True.

## Context (Environment)
Existing resource was present and may or may not affect the outcome of the result. This was a LaunchTemplate whose versioning updating was working just fine prior to upgrading the pulumi-aws provider to 4.5.1 and pulumi to 3.2.1.

This seems like a really old Terraform issue to be honest, that somehow resurfaced itself (regression?) in the newer AWS provider. https://github.com/hashicorp/terraform-provider-aws/issues/4655

## Affected feature
N/A

t0yv0:
Fixes #1504

Due to discrepancies between TF and Pulumi Bridged provider life cycle, editing tags on a Launch Template resource caused failure in pulumi-aws. This is now fixed by opting in this resource to the PlanResourceChange behavior that tracks TF lifecycle more closely.


provider/provider_python_test.go
```diff
@@ -30,6 +32,36 @@ func TestRegress3196(t *testing.T) {
...
 }
 
+func TestRegress1504(t *testing.T) {
+	if testing.Short() {
+		t.Skipf("Skipping test in -short mode because it needs cloud credentials")
+		return
+	}
+	test := getPythonBaseOptions(t).
+		With(integration.ProgramTestOptions{
+			// Not ideal, need investigation:
```
t0yv0, line 42: Filed https://github.com/pulumi/pulumi-aws/issues/3686

mjeffryes (approved):
I can't claim to understand the details of what adding PlanResourceChange to the resource does, but I find the test convincing.

PR #3701, opened 2024-03-25
Update auto-generated AWS managed IAM policies

iwahbe:
Updates the auto-generated managed IAM policies.

This PR is a copy of https://github.com/pulumi/pulumi-aws/pull/3652, which will hopefully play better with our testing infrastructure.


iwahbe:
> Any idea why the CI on the original pr didn’t work? I regenerated the sdk locally and found no diffs

Not sure. I've seen failures like that stem from permission errors (`GITHUB_TOKEN` # of requests expired), but I don't know what's going on.

PR #3706, opened 2024-03-25
Add support for C7a instance types

bhurlow:
I followed the flow of https://github.com/pulumi/pulumi-aws/pull/2965/files, but possible I missed one or two spots

PR #3734, opened 2024-03-26
Add support for C7a instance types

t0yv0:
Update generated files on top of #3706


t0yv0:
Failed on expired token, I think https://github.com/pulumi/ci-mgmt/pull/863 might help

corymhall (approved):

PR #3737, opened 2024-03-26
experiment to understand ci speed

t0yv0:


t0yv0:
3m:32s prerequisites job possible on the warm cache it seems: https://github.com/pulumi/pulumi-aws/actions/runs/8445164030/job/23132412156

t0yv0:
Closing, most of this has been upstreamed to ci-mgmt and the bridge. We can still try to remove P=2 compilation parallelism limits that were there to limit RAM use, though it may be tricky.

PR #3743, opened 2024-03-27
Correctly set the alt type for `aws_cloudwatch_log_resource_policy`

iwahbe:
This is a breaking change for Go and Typescript if users were reading the `PolicyDocument` field on LogResourcePolicyArgs or on LogResourcePolicy.

This was discovered as part of
https://github.com/pulumi/pulumi-terraform-bridge/pull/1758.


sdk/nodejs/sns/topicPolicy.ts
```diff
@@ -163,5 +165,5 @@ export interface TopicPolicyArgs {
     /**
      * The fully-formed AWS policy as JSON.
      */
-    policy: pulumi.Input<string>;
+    policy: pulumi.Input<string | PolicyDocument>;
```
t0yv0, line 168: This looks meaningfully better for the users.

t0yv0:
I'd be fine with merging because I think it won't break 99.9% of programs, curious on what @corymhall thinks. There is something really improving here for TypeScript users but not so much for Go, Go decides to emit `interface{}` and the typed version of the input is not discoverable.

sdk/nodejs/sns/topicPolicy.ts
iwahbe: This is clearer, but still breaking. `args.policy = new PolicyDocument(...)` is IMO strictly better. `check(args.policy)` now requires `check` to handle multiple types, which is strictly worse.

On the whole, I think it's an improvement.

sdk/go/aws/ecr/registryPolicy.go
```diff
@@ -150,13 +150,13 @@ func (RegistryPolicyState) ElementType() reflect.Type {
 
 type registryPolicyArgs struct {
 	// The policy document. This is a JSON formatted string.
-	Policy string `pulumi:"policy"`
+	Policy interface{} `pulumi:"policy"`
```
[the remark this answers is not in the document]
iwahbe: If you **write** to `RegistryPolicyArgs.Policy`, it is not breaking. If you **read** from `RegistryPolicyArgs.Policy`, then it is breaking. I believe that writes are the common case, but you can read from args.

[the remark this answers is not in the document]
t0yv0: Yeah .. I think so. We haven't discussed how strict we want to be here, but I think we shouldn't count these changes as breaking IMO.

sdk/nodejs/sns/topicPolicy.ts
t0yv0: Why is it breaking?

t0yv0:
Filed p/p enhancement around Go support.

sdk/nodejs/sns/topicPolicy.ts
t0yv0: > check(args.policy) now requires check to handle multiple types, which is strictly worse.

I think this is a listed reason, but which user programs are calling check()? Something using transforms or policy?

iwahbe: `check` is a stand-in for any read. It's the same argument as https://github.com/pulumi/pulumi-aws/pull/3743#discussion_r1541540270, maybe worded a little less clearly.

t0yv0: Gotcha. So more a theoretical possibility.

t0yv0 (approved):

PR #3744, opened 2024-03-27
Revert "run unit tests in separate job (#3666)"

t0yv0:
This reverts commit 249f81b68551648b6de6448304eee7b103ba90d5.

Looks like this change started blocking releases in https://github.com/pulumi/pulumi-aws/actions/runs/8454321289

On top of the revert, I've ran `make ci-mgmt` to adopt the latest changes from ci-mgmt.


rquitales (approved):

.ci-mgmt.yaml
```diff
@@ -199,138 +202,3 @@ extraTests:
       matrix:
         language:
         - nodejs
-
-  provider_test:
-    if: github.event_name == 'repository_dispatch' ||
-      github.event.pull_request.head.repo.full_name == github.repository
```
guineveresaenger, line 205: Pretty sure this condition is the reason why the test was skipped - this condition makes no sense on `master` and `release` builds.

might consider removing this instead? I understand we have sped up the tests via cache elsewhere, just a suggestion.

t0yv0: Yeah it's possible to simply remove it, hm.. I was going on the assumption that reverting is the fastest way to success however it seems like we have hiccups here again.

t0yv0: Yeah I'll try to go that route. It looks like we hit out of disk space. I'm pretty sure I know why - it's the excessive go dependencies of ./provider being added here.

PR #3745, opened 2024-03-27
Fix release.yml

t0yv0:
Instead of https://github.com/pulumi/pulumi-aws/pull/3744

It looks like simply reverting the change that caused us being unable to release hits disk space limits. 

Trying this change instead that avoids skipping the job so that releases are not skipped.

As part of this,  I ran `make ci-mgmt` which also inherited latest ci-mgmt changes.


guineveresaenger (approved):
let's try it! :shipit:

Issue #3752, opened 2024-03-28
rds.dataSourceEngineVersionRead panic


t0yv0:
Panic in rds.dataSourceEngineVersionRead reported by customer.

```
         panic: interface conversion: interface {} is nil, not map[string]interface {}
         
         goroutine 44 [running]:
         
         github.com/hashicorp/go-cty/cty.Value.GetAttr({{{0x1380f628?, 0xc002c1c758?}}, {0x0?, 0x0?}}, {0x11c8040c, 0x6})
         
         	/home/runner/go/pkg/mod/github.com/hashicorp/go-cty@v1.4.1-0.20200723130312-85980079f637/cty/value_ops.go:711 +0x2de
         
         github.com/hashicorp/terraform-provider-aws/internal/service/rds.criteriaSet(0xc003896800?, {0xc003896ec0?, 0x9, 0x10?})
         
         	/home/runner/work/pulumi-aws/pulumi-aws/upstream/internal/service/rds/engine_version_data_source.go:481 +0x9b
         
         github.com/hashicorp/terraform-provider-aws/internal/service/rds.dataSourceEngineVersionRead({0x1380f1c8, 0xc00383e660}, 0x0?, {0x11c1f880?, 0xc00219d500?})
         
         	/home/runner/work/pulumi-aws/pulumi-aws/upstream/internal/service/rds/engine_version_data_source.go:227 +0x645
         
         github.com/hashicorp/terraform-provider-aws/internal/provider.New.(*wrappedDataSource).Read.interceptedHandler[...].func7(0x0?, {0x11c1f880?, 0xc00219d500?})
         
         	/home/runner/work/pulumi-aws/pulumi-aws/upstream/internal/provider/intercept.go:112 +0x283
         
         github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).read(0x1380f1c8?, {0x1380f1c8?, 0xc00380fbc0?}, 0xd?, {0x11c1f880?, 0xc00219d500?})
         
         	/home/runner/go/pkg/mod/github.com/pulumi/terraform-plugin-sdk/v2@v2.0.0-20240229143312-4f60ee4e2975/helper/schema/resource.go:795 +0x7a
         
[...]

PR #3757, opened 2024-03-29
fix: rds.dataSourceEngineVersionRead panic

corymhall:
This temporarily reverts the upstream change https://github.com/hashicorp/terraform-provider-aws/commit/f8ab2903c9b0d9b30c53ad14e968bd54e2b93e30 because we cannot currently deal with `RawConfig` during read.

re #3752, re https://github.com/pulumi/pulumi-terraform-bridge/issues/1809

### DIFF
--- a/examples/rds-getengineversion/index.ts
+++ b/examples/rds-getengineversion/index.ts
@@ -1,0 +1,3 @@
+import * as aws from "@pulumi/aws";
+
+export const vs = aws.rds.getEngineVersionOutput({engine: "postgres"});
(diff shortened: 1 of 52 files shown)


examples/rds-getengineversion/index.ts
```diff
@@ -0,0 +1,3 @@
+import * as aws from "@pulumi/aws";
+
+export const vs = aws.rds.getEngineVersionOutput({engine: "postgres"});
```
mikhailshilkov, line 3: Very tangential - but should we be defining new tests in YAML these days? @t0yv0 @mjeffryes

t0yv0: Yeah I think the slight preference is for YAML, but the cross-testing advantages of having it in YAML are not yet realized anywhere I think.

PR #3770, opened 2024-04-02
Higher timeout for Go test

t0yv0:
These tests continue to flake up in CI, and saldy I don't have a precise repro but I assume this has to do with eventual consistency of tagging data. Increasing timeout/retry. I restarted this job twice 🤞


mjeffryes (approved):
Just to be clear, I see an increase in the number of retries and the delay between retries, but no changes to a timeout. I assume that you were just referring to the change to the delay, but mentioning it in case there is a timeout change that missed the diff.

PR #3792, opened 2024-04-09
Upstream v5.44.0

corymhall:
This change updates pulumi-aws to the v5.44.0 version of the upstream provider.

See upstream release notes for the detailed list of changes: [hashicorp/terraform-provider-aws@v5.44.0 (release)](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v5.44.0)


patches/0002-Conns-user-agent.patch
```diff
@@ -1,34 +0,0 @@
...
- 	awsbase "github.com/hashicorp/aws-sdk-go-base/v2"
- 	awsbasev1 "github.com/hashicorp/aws-sdk-go-base/v2/awsv1shim/v2"
- 	"github.com/hashicorp/terraform-provider-aws/internal/types"
--	"github.com/hashicorp/terraform-provider-aws/version"
- )
- 
- // ServicePackage is the minimal interface exported from each AWS service package.
-@@ -80,10 +79,10 @@ func NewSessionForRegion(cfg *aws_sdkv1.Config, region, terraformVersion string)
- 
- func StdUserAgentProducts(terraformVersion string) *awsbase.APNInfo {
```
t0yv0, line 23: This is not needed anymore? Looks like this was fixing up User Agent somewhere in one of the clients?

corymhall: ahh good catch! It looks like this configuration was moved somewhere else. I'll create another patch.

corymhall: ok I added another patch https://github.com/pulumi/pulumi-aws/pull/3792/files#diff-fec171c8e8ff659477dbbd56f2c37c3c617a42bcd723bdc592099abfa111b2c5

t0yv0: Yes it looks like an untested patch .. maybe one day we'll backfill the test for it so it'd remind us.

t0yv0:
LGTM after tests pass!

Issue #3800, opened 2024-04-10
m7i and m7i-flex series are missing from InstanceType


mikhailshilkov:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Similarly to https://github.com/pulumi/pulumi-aws/issues/1220, quite a few instance types are missing from the strongly-typed enum value.

### Affected area/feature

EC2 instance provisioning.

PR #3801, opened 2024-04-10
Add missing InstanceTypes

mikhailshilkov:
Resolves https://github.com/pulumi/pulumi-aws/issues/3800

Done with 

```
aws ec2 describe-instance-types | jq  ".InstanceTypes[]" | jq -c -r '"{Name: \"" + (.InstanceType | gsub("\\.";"_") | ./"" | first |= ascii_upcase | add) + "\", Value: \"" + .InstanceType + "\"},"' | sort
```

and then cleaning it up manually with replaces and reordering to minimize the difference.


t0yv0:
Can you add a test that runs this

```
aws ec2 describe-instance-types | jq  ".InstanceTypes[]" | jq -c -r '"{Name: \"" + (.InstanceType | gsub("\\.";"_") | ./"" | first |= ascii_upcase | add) + "\", Value: \"" + .InstanceType + "\"},"' | sort
```

Or equivalent and fails when it's out of sync? Would be awesome. Then we'll auto maintain this in-sync.

mikhailshilkov:
> Then we'll auto maintain this in-sync.

This is one step extra... We'd need a few targeted replacements to satisfy all the convensions. Probably a Go function on top of `jq`.

t0yv0:
Right on! Perhaps copilot can help write something in pure go too to avoid the JQ bit. Just a thought. I wouldn't block to merge this can backlog but would be nice.

mikhailshilkov:
Closing in favor of https://github.com/pulumi/pulumi-aws/pull/3823

PR #3814, opened 2024-04-14
Codify instance type names

mikhailshilkov:
Adds a function that calculates InstanceType Enum Name from the raw value. As far as I can tell, we populated those manually up until now.

No functional change except deprecating one value with a typo. A follow-up PR will add all missing instance types.


t0yv0:
https://github.com/pulumi/pulumi-aws/issues/3819

t0yv0 (approved):

Issue #3817, opened 2024-04-15
Flaky TestAccKmsAliasTs


t0yv0:
### What happened?

https://github.com/pulumi/pulumi-aws/issues/3811

### Example

See above

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #3818, opened 2024-04-15
Fix flaky test TestAccKmsAliasTs

flostadler:
The test is creating a KMS key alias with a static name. When multiple
instances of this test are running in the same AWS account and region they'll race
and occasionally fail because the alias already exists.

This changes the key alias to use a name prefix to guarantee uniqueness.

This fixes: https://github.com/pulumi/pulumi-aws/issues/3817

### DIFF
--- a/examples/kms-alias/index.ts
+++ b/examples/kms-alias/index.ts
@@ -23,7 +23,7 @@ const aliasAutoPrefixedWithAlias = new aws.kms.Alias("should-be-prefixed", {targ
 
 const aliasWithDedicatedName = new aws.kms.Alias("wont-be-autonamed", {
     targetKeyId: key.keyId,
-    name: "alias/my-decicated-name"
+    namePrefix: "alias/my-decicated-name-"
 }, providerOpts);
 
 export const autonamedAlias = aliasAutoPrefixedWithAlias.name;


examples/kms-alias/index.ts
```diff
@@ -23,7 +23,7 @@ const aliasAutoPrefixedWithAlias = new aws.kms.Alias("should-be-prefixed", {targ
 
 const aliasWithDedicatedName = new aws.kms.Alias("wont-be-autonamed", {
     targetKeyId: key.keyId,
-    name: "alias/my-decicated-name"
+    namePrefix: "alias/my-decicated-name-"
```
corymhall, line 26: It looks like this test might be intentionally testing setting a specific name. I can't think of a good reason for that though so this change looks good to me!

corymhall (approved):

examples/kms-alias/index.ts
flostadler: I did think about generating a random name and using that for the `name` parameter, but I also couldn't think of a reason to keep it that way.

Alternatively we could test all three cases (auto-generated name, name, namePrefix)

t0yv0 (approved):

Issue #3680, opened 2024-03-22
Make enums discoverable


t0yv0:
### What happened?

```
managed_policy_arns=[aws.iam.ManagedPolicy.AWS_LAMBDA_BASIC_EXECUTION_ROLE])
```

Documentation lists this field as string-typed https://www.pulumi.com/registry/packages/aws/api-docs/iam/role/#managedpolicyarns_nodejs and makes it hard to find the enumeration with good values to send in.

### Example

```
import json
import pulumi
import pulumi_aws as aws
import pulumi_aws_apigateway as apigateway

# An execution role to use for the Lambda function
role = aws.iam.Role("role", 
    assume_role_policy=json.dumps({
        "Version": "2012-10-17",
        "Statement": [{
            "Action": "sts:AssumeRole",
            "Effect": "Allow",
            "Principal": {
                "Service": "[lambda.amazonaws.com](http://lambda.amazonaws.com/)",
            },
        }],
    }),
    managed_policy_arns=[aws.iam.ManagedPolicy.AWS_LAMBDA_BASIC_EXECUTION_ROLE])

# A Lambda function to invoke
fn = aws.lambda_.Function("fn",
    runtime="python3.9",
    handler="handler.handler",
    role=role.arn,
    code=pulumi.FileArchive("./function"))

# A REST API to route requests to HTML content and the Lambda function
api = apigateway.RestAPI("api",
  routes=[
    apigateway.RouteArgs(path="/", local_path="www"),
    apigateway.RouteArgs(path="/date", method=apigateway.Method.GET, event_handler=fn)
  ])

# The URL at which the REST API will be served.
pulumi.export("url", api.url)
```

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #3832, opened 2024-04-17
Add AWSBackupServiceRolePolicyForBackup to ManagedPolicies

jesperseverinsen:


corymhall (changes requested):
@jesperseverinsen thanks for the PR! It looks like the `ManagedPolicy` is preferred over this `ManagedPolicies` object, and this policy already exists there

https://github.com/pulumi/pulumi-aws/blob/6623fc6a2abe85ad276ad213b674f46cc788a4b6/provider/iam_managed_policy.go#L5

We also have an issue open to make these types more discoverable. https://github.com/pulumi/pulumi-aws/issues/3680

jesperseverinsen:
@corymhall I wonder how I didn't notice that ManagedPolicies and ManagedPolicy were two different things, too much tunnel vision I suppose. It's all good, I'll close this PR then.

PR #3838, opened 2024-04-17
Fix panic in aws.rds.Proxy

### LINKED ISSUE #3835: `rds.resourceProxy` Panic
Panic in `rds.resourceProxy` reported by a customer. Will upload full log separately.

Top level error 
```
         [stdout]  2024-04-17 14:09:45 INFO     panic: interface conversion: interface {} is nil, not map[string]interface {} 
         [stdout]  2024-04-17 14:09:45 INFO      
         [stdout]  2024-04-17 14:09:45 INFO     goroutine 270 [running]: 
         [stdout]  2024-04-17 14:09:45 INFO      
         [stdout]  2024-04-17 14:09:45 INFO     github.com/hashicorp/terraform-provider-aws/internal/service/rds.resourceProxy.SimpleSchemaSetFunc.func1({0x0?, 0x0?}) 
```

t0yv0:
Fixes #3835 by reverting the upstream change until the upstream issue is root-caused and resolved. Passing an empty object to auths field of `aws.rds.Proxy` should no longer panic.

### DIFF
--- a/provider/provider_nodejs_test.go
+++ b/provider/provider_nodejs_test.go
@@ -75,6 +75,6 @@ func TestRegress3835(t *testing.T) {
 		opttest.YarnLink("@pulumi/aws"),
 	}
 	test := pulumitest.NewPulumiTest(t, dir, options...)
-	upResult := test.Up()
-	t.Logf("#%v", upResult.Summary)
+	result := test.Preview()
+	t.Logf("#%v", result.ChangeSummary)
 }
--- a/provider/provider_nodejs_test.go
+++ b/provider/provider_nodejs_test.go
@@ -63,3 +63,18 @@ func TestRegress3094(t *testing.T) {
 	upResult := test.Up()
 	t.Logf("#%v", upResult.Summary)
 }
+
+func TestRegress3835(t *testing.T) {
+	skipIfShort(t)
+	dir := filepath.Join("test-programs", "regress-3835")
+	cwd, err := os.Getwd()
+	require.NoError(t, err)
+	providerName := "aws"
+	options := []opttest.Option{
+		opttest.LocalProviderPath(providerName, filepath.Join(cwd, "..", "bin")),
+		opttest.YarnLink("@pulumi/aws"),
+	}
+	test := pulumitest.NewPulumiTest(t, dir, options...)
+	upResult := test.Up()
+	t.Logf("#%v", upResult.Summary)
+}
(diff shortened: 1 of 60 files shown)


provider/provider_nodejs_test.go
```diff
@@ -63,3 +63,18 @@ func TestRegress3094(t *testing.T) {
...
+	dir := filepath.Join("test-programs", "regress-3835")
+	cwd, err := os.Getwd()
+	require.NoError(t, err)
+	providerName := "aws"
+	options := []opttest.Option{
+		opttest.LocalProviderPath(providerName, filepath.Join(cwd, "..", "bin")),
+		opttest.YarnLink("@pulumi/aws"),
+	}
+	test := pulumitest.NewPulumiTest(t, dir, options...)
+	upResult := test.Up()
```
corymhall, line 78: I feel like we should test `preview` instead here because based on the integration test failure, this is actually invalid code that will fail on deployment. Seems like the real issue is https://github.com/pulumi/pulumi-terraform-bridge/pull/1881 causing the preview value to be incorrect.

provider/test-programs/regress-3835/index.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const example = new aws.rds.Proxy("example", {
+  name: "example",
+  debugLogging: false,
+  engineFamily: "MYSQL",
+  idleClientTimeout: 1800,
+  requireTls: true,
+  roleArn: proxyRole.arn,
+  vpcSubnetIds: [subnet.id, anotherSubnet.id],
+  auths: [{}],
```
corymhall, line 38: It looks like the integration test is failing because this actually is invalid. If you want to actually run an `up` instead of just `preview` you can change this to

```ts
const secret = new aws.secretsmanager.Secret("secret", {});
const example = new aws.rds.Proxy("example", {
  debugLogging: false,
  engineFamily: "MYSQL",
  idleClientTimeout: 1800,
  requireTls: true,
  roleArn: proxyRole.arn,
  vpcSubnetIds: [subnet.id, anotherSubnet.id],
  auths: pulumi
    .all([secret.arn])
    .apply((sec): aws.types.input.rds.ProxyAuth[] => {
      return [
        {
          iamAuth: "DISABLED",
          authScheme: "SECRETS",
          secretArn: sec[0],
        },
      ];
    }),
  tags: {
    Name: "example",
    Key: "value",
  },
});
```

Although if we ever fix https://github.com/pulumi/pulumi-terraform-bridge/pull/1881 then we would not reproduce it with this test.

corymhall (approved):

provider/provider_nodejs_test.go
t0yv0: Good point. It reproduces on preview.

provider/test-programs/regress-3835/index.ts
t0yv0: Yes let's try the test with an unknown like you suggest. This regression test will probably be short-lived but I just wanted some confirmation that the patch edit is working.

t0yv0:
Test failures fixed on master; rebasing.

PR #3843, opened 2024-04-18
Add CODEBUILD_CONFIG_AUTO_DISCOVER environment variable of CodeBuild example

flostadler:
AWS started setting an environment variable by default to auto discover reports (CODEBUILD_CONFIG_AUTO_DISCOVER) when creating CodeBuild projects.
This led to `pulumi preview` showing a diff after the initial deployment.

This change will fix the test, but the in behavior  from AWS will lead to users ending up with a drift when creating CodeBuild projects until they re-run `pulumi up` once.

This fixes https://github.com/pulumi/pulumi-aws/issues/3839


t0yv0 (approved):
You can log the problem into Issues and if people run into it it will get upvoted and we can look at compensating.

Issue #2447, opened 2023-03-31
explicit provider secrets are shown as plaintext in CLI output


phillipedwards:
### What happened?

Explicit providers sourcing plaintext values from the environment for accessKey, secretKey, token are showing up in the CLI output (pre and up) as plaintext.

### Expected Behavior

Regardless if plaintext values are provided to provider or not, they should be marked as secrets (outputs masked).

### Steps to reproduce

Create an aws typescript project with the following:

```
const awsProvider = new aws.Provider("aws-provider", {
        accessKey: process.env["AWS_ACCESS_KEY_ID"],
        secretKey: process.env["AWS_SECRET_ACCESS_KEY"],
        token: process.env["AWS_SESSION_TOKEN"],
        region: <aws.Region>process.env["AWS_REGION"],
    });
```
set the applicable environment variables with plaintext values. Those plaintext values will show up in the CLI output.

### Output of `pulumi about`

@pulumi/aws@5.13.0

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

lukehoban:
Related to https://github.com/pulumi/pulumi/issues/10319.

phillipedwards:
My understanding is if a provider input is marked as `Sensitive` in the upstream provider, inputs should automatically be marked as Secrets, but it seems that is not true? @lukehoban

rquitales:
@phillipedwards That was also my understanding. I'll take a deeper look into this.

t0yv0:
Looks like we still don't have a good fix here, escalating so our team takes a look. Apologies for the wait.

t0yv0:
@phillipedwards do you have a reference for "showing up in the CLI output (pre and up) as plaintext." are you perchance running Pulumi with some parameters? on a very simple `pulumi up` I'm currently not seeing these values exposed on stdout/stderr on the latest version of the provider.

However! I do see the values exposed as plaintext in the statefile which is very bad.

I seem to be able to easily fix that behavior by configuring  the following config options as secret in resources.go:

```
			"access_key": {
				Secret: tfbridge.True(),
			},
			"secret_key": {
				Secret: tfbridge.True(),
			},
			"token": {
				Secret: tfbridge.True(),
			},
```

They are not marked sensitive upstream but we can make them secret in the pulumi projection. Doing so encrypts them in the state both in output and in input position, so state is then clear. 

[...]

PR #3863, opened 2024-04-24
Marks accessKey, secretKey and token options as Secret

t0yv0:
This makes sure that any values passed to these provider configuration options are stored encrypted in the state file.

Fixes https://github.com/pulumi/pulumi-aws/issues/2447

### DIFF
--- a/provider/configure_test.go
+++ b/provider/configure_test.go
@@ -104,11 +104,11 @@ func TestCheckConfigFastWithCustomEndpoints(t *testing.T) {
 	  },
 	  "response": {
 	    "inputs": {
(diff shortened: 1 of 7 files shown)


t0yv0:
This could be thought of as a configuration change so skipping a test (verified manually). Could try working on one though if desired.

t0yv0:
Looks like some over-eager golden tests are failing. I will have a look tomorrow.

Could be nice to have automation that tries to auto-correct PR that failed golden tests and propose another PR.

corymhall (approved):

Issue #3887, opened 2024-05-02
Regression - 6.33.0 AWS Batch job definition - 'Job definition name is required'


Raggey:
### What happened?

Upgraded to latest version and getting an error when updating a batch job definition

### Example

`aws:batch:JobDefinition (dev-clustering-clustering-job-defn):
    error: 1 error occurred:
        * updating urn:pulumi:dev::clustering::aws:batch/jobDefinition:JobDefinition::dev-clustering-clustering-job-defn: 1 error occurred:
        * deleting Batch Job Definition (): ClientException: Error executing request, Exception : Job definition name is required.`

### Output of `pulumi about`

`CLI          
Version      3.114.0
Go Version   go1.22.2
Go Compiler  gc

Plugins
KIND      NAME    VERSION
resource  aws     6.33.0
resource  docker  4.5.0
language  python  unknown

Host     
OS       ubuntu
Version  22.04
Arch     x86_64
`

### Additional context

Reverting to 6.32.0 resolves the issue, tested on 6.32.0, 3.13. 

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

Raggey:
> Hi @Raggey thanks for posting this and I'm sorry you're hitting this regression. Could you add a quick sample program that configures JobDefinition and hits the issue? Much appreciated.

Hi @t0yv0 ,

Have managed to narrow down exactly how to re-create it. (Sorry should have posted this in the original issue)

Starting with the example from the docs:

```import json

import pulumi_aws as aws

test = aws.batch.JobDefinition(
    "test",
    name="my_test_batch_job_definition",
    type="container",
    container_properties=json.dumps(
        {
            "command": [
                "ls",
                "-la",
            ],
            "image": "busybox",
            "resourceRequirements": [
                {
                    "type": "VCPU",
                    "value": "1",
                },
                {
                    "type": "MEMORY",
                    "value": "512",
                },
            ],
            "volumes": [
                {
                    "host": {
                        "sourcePath": "/tmp",
                    },
                    "name": "tmp",
                }
            ],
            "environment": [
                {
                    "name": "VARNAME",
                    "value": "VARVAL",
                }
            ],
            "mountPoints": [
                {
                    "sourceVolume": "tmp",
                    "containerPath": "/tmp",
                    "readOnly": False,
                }
            ],
            "ulimits": [
                {
[...]

PR #3888, opened 2024-05-02
Fixes errors and resource cycling issues for aws.batch.JobDefinition

t0yv0:
Recent changes in hashicorp/terraform-provider-aws#37111 introduced a Diff customizer that leads to errors in the Pulumi version of the provider due to discrepancies in the order in which Diff customizer functions are applied between Pulumi and Terraform. This change fixes the problem by applying the experimental PlanResourceChange flag to the affected resource.

Fixes https://github.com/pulumi/pulumi-aws/issues/3887

A regression test is included.

See also: https://github.com/pulumi/pulumi-terraform-bridge/issues/1785


flostadler (approved):
LGTM

PR #3891, opened 2024-05-03
Upgrade dependencies affecting example rendering

t0yv0:
This change fixes examples that rendered incorrectly before that involved null, Kubernetes and local providers.

### DIFF
--- a/sdk/java/src/main/java/com/pulumi/aws/ec2/SecurityGroup.java
+++ b/sdk/java/src/main/java/com/pulumi/aws/ec2/SecurityGroup.java
@@ -295,7 +295,7 @@ import javax.annotation.Nullable;
  * import com.pulumi.aws.ec2.SecurityGroupArgs;
  * import com.pulumi.command.local.Command;
  * import com.pulumi.command.local.CommandArgs;
- * import com.pulumi.null.resource;
+ * import com.pulumi.null.Resource;
  * import com.pulumi.null.ResourceArgs;
  * import com.pulumi.resources.CustomResourceOptions;
  * import java.util.List;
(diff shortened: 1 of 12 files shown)


sdk/java/src/main/java/com/pulumi/aws/ec2/SecurityGroup.java
```diff
@@ -335,7 +335,10 @@
  *                 .build());
  * 
  *         var exampleResource = new Resource(&#34;exampleResource&#34;, ResourceArgs.builder()        
- *             .triggers(%!v(PANIC=Format method: runtime error: invalid memory address or nil pointer dereference))
+ *             .triggers(Map.of(&#34;rerun_upon_change_of&#34;, StdFunctions.join(JoinArgs.builder()
```
t0yv0, line 338: One PANIC removed

sdk/java/src/main/java/com/pulumi/aws/iot/IotFunctions.java
```diff
@@ -48,8 +50,12 @@ public final class IotFunctions {
      *         final var example = IotFunctions.getEndpoint();
      * 
      *         var agent = new Pod(&#34;agent&#34;, PodArgs.builder()        
-     *             .metadata(%!v(PANIC=Format method: runtime error: invalid memory address or nil pointer dereference))
-     *             .spec(%!v(PANIC=Format method: runtime error: invalid memory address or nil pointer dereference))
```
t0yv0, line 52: Another one

iwahbe (approved):
PANIC in docs is not the best way to surface a missing plugin dependency...

LGTM

Issue #3880, opened 2024-04-30
wafv2 rules are modified on every up.   


gmarkowski:
### What happened?

I created a new wafv2.WebAcl. Everything gets built properly but subsequent `pulumi up`  commands output differences in that are unexpected. I also see some changes doubled. See `cloudwatchMetricsEnabled` and `metricName` in the output below.

### Example

```typescript
const mainAcl = new aws.wafv2.WebAcl("main-acl", {
    scope: "REGIONAL",
    defaultAction: {
        allow: {},
    },
    visibilityConfig: {
        cloudwatchMetricsEnabled: true,
        metricName: "myWebAclMetrics",
        sampledRequestsEnabled: false,
    },
    captchaConfig: {
        immunityTimeProperty: {
            immunityTime: 1800
        }
    },
    tokenDomains: [],
    rules: [
        {
            action: {
                block: {},
            },
            name: "IPAllowRule",
            priority: 0,
            statement: {
                ipSetReferenceStatement: {
                    arn: allowedIpSet.arn,
                },
            },
            visibilityConfig: {
                cloudwatchMetricsEnabled: true,
                metricName: "IPAllowRule",
                sampledRequestsEnabled: true,
            },
        },
```

### Output of `pulumi about`

```
Do you want to perform this update? details
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:poc::MY-COMPANY::pulumi:pulumi:Stack::MY-COMPANY-poc]
    ~ aws:wafv2/webAcl:WebAcl: (update)
        [id=ded6d29d-5318-4e98-aa14-740f31dbc4f4]
        [urn=urn:pulumi:poc::MY-COMPANY::aws:wafv2/webAcl:WebAcl::main-acl]
        [provider=urn:pulumi:poc::MY-COMPANY::pulumi:providers:aws::default_6_31_0::911e0d16-9224-481c-8e0c-62b88c6aa82b]
      ~ rules: [
          ~ [0]: {
                  ~ action          : {
                      + __defaults: []
                      ~ allow     : {
                          + __defaults: []
                        }
                    }
                  ~ name            : "IPAllowRule" => "IPAllowRule"
                  ~ priority        : 0 => 0
                  ~ statement       : {
                      + __defaults             : []
                      ~ ipSetReferenceStatement: {
                          + __defaults: []
                        }
                    }
                  ~ visibilityConfig: {
                      + __defaults              : []
[...]

PR #3897, opened 2024-05-06
wafv2.WebAcl: tweak Set identity to suppress resource cycling

t0yv0:
Fixes https://github.com/pulumi/pulumi-aws/issues/3880

Deep investigation in https://github.com/pulumi/pulumi-terraform-bridge/pull/1917 - the root cause bottoms out at inconsistent set element identity computed by the bridge as well as Terraform proper; there is some slight discrepancy in how the bridged providers work compared to pure TF that makes the result poor in Pulumi, however even under TF there are two different values computed for what the user perceives as the same data. This is a surgical change to that makes the set identity indifferent to whether block.custom_response is encoded as missing or an empty list, which works around the problem.


t0yv0:
Closing in favor of https://github.com/pulumi/pulumi-terraform-bridge/pull/1927

t0yv0:
This was actually fundamentally incorrect per recent conversation - @corymhall  - excluding elements from Set function will make changes to these elements always result in no-diff plans which is ultimately incorrect.

PR #3898, opened 2024-05-06
Upstream v5.48.0

t0yv0:
Upgrades upstream to v5.48.0

Fixes https://github.com/pulumi/pulumi-aws/issues/3890


t0yv0:
Looks like the regression test is catching an actual problem. We introduced custom retry for KMS condition on lambda invocation and it has regressed because upstream is moving from Go SDK v1 to Go SDK V2:

https://github.com/pulumi/pulumi-aws/pull/3234

t0yv0:
Fixed the patch to use AWS SDKv2 and trying this again.

t0yv0:
aws_bedrockagent_agent_knowledge_base_association mapping to bedrock/agentAgentKnowledgeBaseAssociation.AgentAgentKnowledgeBaseAssociation is unfortunate, perhaps can do better here.

t0yv0:
TestAccWebserverComp flake.

corymhall (approved):

t0yv0:
I'm running into some difficulties with moving bedrockanalysis resources namely https://github.com/pulumi/pulumi-terraform-bridge/issues/1937

Going to backlog this and take the change later.

PR #3910, opened 2024-05-08
Fix aws.s3.Bucket tag drift detection

### LINKED ISSUE #3674: Bucket(V1): pulumi refresh does not detect tag changes
### What happened?

Using `aws.s3.Bucket()` resource and setting a tag, if I change the tag outside of Pulumi and do a `pulumi refresh` the change is not detected.
Similarly, if I remove the tag outside of pulumi, a refresh does not detect the tag is gone.

### Example

1. `pulumi new aws-python`
2. Modify code to add a `tags={"mytag": "myvalue"}` block to the bucket.
3. `pulumi up`
4. Edit or remove the tag in AWS
5. `pulumi refresh` and see that no change is detected.

### Output of `pulumi about`

CLI          
Version      3.111.0
Go Version   go1.22.1
Go Compiler  gc

Plugins
NAME    VERSION
aws     6.27.0
python  unknown

Host     
OS       darwin
Version  13.6.3
Arch     x86_64

This project is written in python: executable='/Users/mitch/Downloads/fxgt-workshop/test-tag-refresh/venv/bin/python3' version='3.10.11'

Current Stack: MitchGerdisch/test-tag-refresh/dev

TYPE                      URN
pulumi:pulumi:Stack       urn:pulumi:dev::test-tag-refresh::pulumi:pulumi:Stack::test-tag-refresh-dev
pulumi:providers:aws      urn:pulumi:dev::test-tag-refresh::pulumi:providers:aws::default_6_27_0
aws:s3/bucketV2:BucketV2  urn:pulumi:dev::test-tag-refresh::aws:s3/bucketV2:BucketV2::my-bucket

Backend        
Name           pulumi.com
Token type     personal

Dependencies:
NAME        VERSION
pip         24.0
pulumi_aws  6.27.0
setuptools  69.2.0
wheel       0.43.0

### Additional context

If I use `aws.s3.BucketV2` the refresh works as expected.

### Contributing

t0yv0:
Make sure that `pulumi refresh` detects changes to bucket tagging.

The implementation restores aws_s3_legacy_bucket resource participation in the generic tags interceptor.

Fixes #3674

### DIFF
--- a/examples/examples_go_test.go
+++ b/examples/examples_go_test.go
@@ -94,6 +94,11 @@ func TestTagsCombinationsGo(t *testing.T) {
 			tagsState{DefaultTags: map[string]string{}, ResourceTags: map[string]string{"x": "s", "y": ""}},
 			tagsState{DefaultTags: map[string]string{"x": ""}, ResourceTags: map[string]string{}},
 		},
+		{
+			"regress 3",
+			tagsState{DefaultTags: map[string]string{"x": "", "y": "s"}, ResourceTags: map[string]string{"x": "s", "y": "s"}},
+			tagsState{DefaultTags: map[string]string{}, ResourceTags: map[string]string{"x": "", "y": "s"}},
+		},
 	}
 
 	for i, tc := range testCases {
--- a/patches/0038-Restore-legacy-bucket.patch
+++ b/patches/0038-Restore-legacy-bucket.patch
@@ -76,6 +76,54 @@ index deadff7071..6310023478 100644
  func (client *AWSClient) S3ConnURICleaningDisabled(ctx context.Context) *s3_sdkv1.S3 {
  	config := client.S3Conn(ctx).Config
  	config.DisableRestProtocolURICleaning = aws_sdkv2.Bool(true)
+diff --git a/internal/provider/provider.go b/internal/provider/provider.go
+index 9b4ebd5a46..0e7a94e470 100644
+--- a/internal/provider/provider.go
++++ b/internal/provider/provider.go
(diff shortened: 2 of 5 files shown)


t0yv0:
The randomized tagging tests flagged legacy-bucket not behaving correctly which is very relevant to this change. Specifically the randomized state transition defying the expectation is: 

```
     {"defaultTags":{"x":"","y":"s"},"resourceTags":{"x":"s","y":"s"}}
 --> {"defaultTags":{},"resourceTags":{"x":"","y":"s"}

      expected: map[string]string{"x":"", "y":"s"}
      actual  : map[string]string{"x":"s", "y":"s"}
```

We have seen flakes here before due to tag retrieval eventual consistency but I'm going to double check since this very well may be an actual problem.

t0yv0:
Doesn't reproduce locally, must have been a flake.

corymhall (approved):
@t0yv0 I'm assuming that adding the legacy bucket as a `ServicePackage` is what enrolls it into the generic tags? Is there any other implications of doing this, I'm not very familiar with what `ServicePackage` does.

t0yv0:
I think yes that's it. https://github.com/hashicorp/terraform-provider-aws/blob/master/internal/provider/provider.go#L39 walks through all ServicePackage defined resources and https://github.com/hashicorp/terraform-provider-aws/blob/master/internal/provider/provider.go#L329 installs the interceptor. It looks like it does a few checks. I think it's reasonable for the legacy bucket to participate here.

flostadler (commented):
LGTM

Issue #3905, opened 2024-05-07
Panic in batch job definition in 6.33.1


t0yv0:
### What happened?

Customer report:

```
panic: interface conversion: interface {} is nil, not map[string]interface {}
    goroutine 2639 [running]:
    github.com/hashicorp/terraform-provider-aws/internal/service/batch.needsJobDefUpdate(0x1400266c3c0)
    	/home/runner/work/pulumi-aws/pulumi-aws/upstream/internal/service/batch/job_definition.go:537 +0x1004
    github.com/hashicorp/terraform-provider-aws/internal/service/batch.jobDefinitionCustomizeDiff({0x11282d880?, 0x11e7ab5c0?}, 0x1400266c3c0, {0xe?, 0x140020009c0?})
    	/home/runner/work/pulumi-aws/pulumi-aws/upstream/internal/service/batch/job_definition.go:452 +0x44
    github.com/hashicorp/terraform-provider-aws/internal/service/batch.ResourceJobDefinition.Sequence.func23({0x113b73488, 0x14002680030}, 0x1400266c3c0, {0x1139133c0, 0x140020009c0})
    	/home/runner/go/pkg/mod/github.com/pulumi/terraform-plugin-sdk/v2@v2.0.0-20240229143312-4f60ee4e2975/helper/customdiff/compose.go:69 +0x80
    github.com/hashicorp/terraform-provider-aws/internal/provider.New.(*wrappedResource).CustomizeDiff.func5({0x113b73488?, 0x140040117d0?}, 0x1400266c3c0, {0x1139133c0, 0x140020009c0})
    	/home/runner/work/pulumi-aws/pulumi-aws/upstream/internal/provider/intercept.go:186 +0x68
    github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.schemaMap.Diff(0x1400201def0, {0x113b73488, 0x140040117d0}, 0x14003546dd0, 0x1400277fae0, 0x14001187d70, {0x1139133c0, 0x140020009c0}, 0x0)
    	/home/runner/go/pkg/mod/github.com/pulumi/terraform-plugin-sdk/v2@v2.0.0-20240229143312-4f60ee4e2975/helper/schema/schema.go:698 +0x3b8
[...]

PR #3921, opened 2024-05-10
Fix empty retry_strategy of Batch JobDefinition causing panics

flostadler:
Computing whether a Batch JobDefinition has a diff panics if the `retry_strategy` is empty. This is a valid configuration because all properties of `retry_strategy` are optional.

This patch fixes that by adding the missing nil checks for the retry strategy.

What's notable is that having an empty `retry_strategy` will cause a perma diff. `MaxItems=1` blocks with all optional attributes lead to permanent diffs if the user provides an empty block.

Fixes https://github.com/pulumi/pulumi-aws/issues/3905


t0yv0 (approved):
Thank you! Looks clean.

PR #3923, opened 2024-05-10
Dependencies: terraform converter v1.0.17

t0yv0:
This upgrades examples pipeline to the latest terraform converter v1.0.17.

Examples are added since more examples convert on the latest version.


sdk/nodejs/ec2/getNatGateways.ts
```diff
@@ -9,6 +9,26 @@ import * as utilities from "../utilities";
...
+ * import * as aws from "@pulumi/aws";
+ *
+ * const ngws = aws.ec2.getNatGateways({
+ *     vpcId: vpcId,
+ *     filters: [{
+ *         name: "state",
+ *         values: ["available"],
+ *     }],
+ * });
+ * const ngw = .map(__index => (aws.ec2.getNatGateway({
```
t0yv0, line 28: The .map does not look right.

flostadler (approved):

sdk/nodejs/ec2/getNatGateways.ts
t0yv0: This is https://github.com/pulumi/pulumi/issues/15859 which is a known issue, linked it to this failure but not blocking on it.

PR #3932, opened 2024-05-13
fix: sns topic creation fails in non-standard partitions

### LINKED ISSUE #3926: Improper partition used when generating SNS Topic ARN
### What happened?

In US Gov partitions, the provider generates the wrong ARN for the `GetTopicAttributes` call. This results in the `pulumi_aws.sns.Topic()` resource from failing to create when the lookup happens.

### Example

When using a provider in the US Gov (or China I'm sure too) creation of the topic fails:

```test_topic = aws.sns.Topic("test_tjf_1")```

### Output of `pulumi about`

```
pulumi about
CLI          
Version      3.112.0
Go Version   go1.22.1
Go Compiler  gc

Plugins
NAME        VERSION
aws         6.32.0
datadog     4.28.0
kubernetes  4.11.0
postgresql  3.11.0
python      unknown
random      4.16.1

Host     
OS       darwin
Version  13.6.6
Arch     arm64

This project is written in python: executable='/Users/tomfarvour/.local/share/virtualenvs/pulumi-FBQqJsZg/bin/python3' version='3.11.7'
```

### Additional context

This patch, which I presume patches the upstream TF provider, assumes it can use `aws` as the partition.

https://github.com/pulumi/pulumi-aws/blob/ba6524913c1a4b5b254bcc20465ed59b83b92b34/patches/0052-non-idempotent-sns-topic-creation.patch#L23-L28

### Contributing

corymhall:
Set the AWS partition dynamically instead of hardcoding the default `aws` partition.

I'm not sure there is a good way to test this since we do not have credentials for other partitions

fixes #3926

### DIFF
--- a/patches/0052-non-idempotent-sns-topic-creation.patch
+++ b/patches/0052-non-idempotent-sns-topic-creation.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] non-idempotent sns topic creation
 
 
 diff --git a/internal/service/sns/topic.go b/internal/service/sns/topic.go
-index 17c78c35c1..0d2e35b695 100644
+index 17c78c35c1..5ee99c0c44 100644
 --- a/internal/service/sns/topic.go
 +++ b/internal/service/sns/topic.go
 @@ -9,6 +9,7 @@ import (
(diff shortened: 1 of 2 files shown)


farvour:
> I'm not sure there is a good way to test this since we do not have credentials for other partitions

Yeah they are specialty partitions. We plan to use this as soon as it merges. Hoping to avoid having to compile Pulumi from source to test this if that's what you're waiting on. Generally speaking, if you're using the `aws.Client` provider's `Partition` attribute, and it's working for you and returning 'aws', then it'll likely properly return 'aws-us-gov' or 'aws-cn' in the others.

flostadler (approved):
LGTM

t0yv0:
Thank you!

PR #3933, opened 2024-05-13
Explorations: rewire legacy bucket backend

t0yv0:
BucketV2 and Bucket underlying schemata are still very close and almost identical, since they have a common ancestry, with BucketV2 continuing to evolve and Bucket lagging behind. This experiment attempts to exploit this by redirecting the implementation details of handling Bucket CRUD to the BucketV2 implementation.

This change is ideally a refactor invisible to the user but given a substantial API surface need to be careful not to regress.

- removed s3legacy and recreated the legacy bucket against v2 code
- encountering some bugs, we might need to remove default: "private" from "acl"

```
* putting S3 Bucket (my-bucket-eb276e3) ACL: operation error S3: PutBucketAcl, https response error StatusCode:
       400, RequestID: GMTPEW43STHM9DWM, HostID:
       /NJMasmaMuMN64OXgC+h58N1ozOYowaC0dP2I9rb2z7NPRZ83BWrabjNk0SIqYniyPCJe/iD/oI=, api error
       AccessControlListNotSupported: The bucket does not allow ACLs
```
   
- logging (loggings?) has changed from Set to List with MaxItems=1 there might be a slight change to compensate for
- there were a lot of custom TypeSet Set functions specified in the legacy, why were they removed in BucketV2, is it OK to match or we can roll without them
- generally might need to add lifecycle and provider upgrade tests targeting each attribute like "loggings"
- removes a fair bit of Pulumi-maintained custom code, need to itemize and ensure that was redundant

### DIFF
--- a/patches/0057-Fix-empty-retry_strategy-of-Batch-JobDefinitions-cau.patch
+++ b/patches/0057-Fix-empty-retry_strategy-of-Batch-JobDefinitions-cau.patch
@@ -12,7 +12,7 @@ This patch fixes that by adding the missing nil checks for the
 retry strategy.
 
 diff --git a/internal/service/batch/job_definition.go b/internal/service/batch/job_definition.go
-index d1b5102162..a6a9c190c8 100644
+index d1b5102162..6f0d4976e7 100644
 --- a/internal/service/batch/job_definition.go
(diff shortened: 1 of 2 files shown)


t0yv0:
Opting against this route.

Issue #3903, opened 2024-05-07
TestAccWebserverComp flake


t0yv0:
See https://github.com/pulumi/pulumi-aws/issues/3895

```
aws:ec2:Instance (web-server-app):
  error: 1 error occurred:
  	* creating urn:pulumi:p-it-fv-az1019--webserver--0adc9e13::webserver_comp::aws:ec2/instance:Instance::web-server-app: 1 error occurred:
  	* waiting for EC2 Instance (i-032319ee5436b513c) create: timeout while waiting for state to become 'running' (last state: 'pending', timeout: 10m0s)
 ```
 
Could be a need for higher timeout, or could be a manifestation of test account health.

t0yv0:
one more: https://github.com/pulumi/pulumi-aws/issues/3902

flostadler:
one more: https://github.com/pulumi/pulumi-aws/issues/3906

flostadler:
Another two occurrences: https://github.com/pulumi/pulumi-aws/issues/3925

This is way too flaky now. I'm gonna dig into why it's starting to fail to often, seems like the startup of the EC2 instance isn't  properly working because it doesn't go to running state

flostadler:
My hunch is that this is caused by a combination of a very small instance type `t2.nano` and a very old AMI (deprecated since 2022, no security patches anymore...).
All the test failures mention the `t2.nano` instance failing to become healthy so I'm gonna try switching that for a t2 micro with a more recent AMI.

flostadler:
I upgraded the AMI of the failing test to AL2023. Let's see if this fixes the test

flostadler:
It happened again: Just upgrading the AMI was not enough. I'll change the test to only start a t2.micro instance

t0yv0:
Sadness. Thanks for looking into this. Feel free to also skip the test while we investigate since the disturbance from flakiness outweighs the coverage we get out of it.

PR #3937, opened 2024-05-14
Upgrade AMI of TestAccWebserverComp test to AL2

flostadler:
The test `TestAccWebserverComp` used a AL1 AMI from 2016 and started to be flaky. AL1 is deprecated and no longer supported. Upgrading it to the latest AL2 AMI might make the test more stable

Closes https://github.com/pulumi/pulumi-aws/issues/3903


VenelinMartinov (approved):

examples/webserver-comp/webserver.ts
```diff
@@ -29,14 +28,23 @@ let group = new aws.ec2.SecurityGroup("web-secgrp-comp", {
     ],
 }, providerOpts);
 
+const ami = aws.ec2.getAmiOutput({
+    owners: ["amazon"],
+    mostRecent: true,
+    filters: [{
+        name: "name",
+        values: ["amzn2-ami-hvm-*-x86_64-ebs"],
```
corymhall, line 36: Should we use the latest al2023 AMIs instead? I think it is `al2023-ami-kernel-default-x86_64`

corymhall (changes requested):

examples/webserver-comp/webserver.ts
flostadler: Yeah that's a good suggestion! They're supported even longer (2025 vs 2028)

flostadler: Done! :)

corymhall (approved):

PR #3958, opened 2024-05-17
Upgrade instances for TestAccWebserverComp test

flostadler:
The TestAccWebserverComp still occasionally flaked after upgrading the AMI. This changes the instances to a newer generation and changes the `nano` instance to a `small` to ensure the long boot times aren't caused by running out of CPU credits

Fixes: https://github.com/pulumi/pulumi-aws/issues/3903


corymhall (approved):
🤞🏻

Issue #1692, opened 2021-11-07
Reconsider the default aws:skipMetadataCheck setting


jaxxstorm:
In #1288 we disabled the metadata check in order to speed Pulumi up. 

Unfortunately, this breaks Pulumi when running with an EC2 instance role, meaning lots of scenarios like self hosted CI/CD runners will not work.

It's possible of course to fix this by setting the environment variable `AWS_EC2_METADATA_DISABLED=false` or `pulumi config set aws:skipMetadataApiCheck false`, but the it's extremely difficult to figure that out unless you inherently know about this setting. This default setting seems to be violating the [principal of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) and is causing a headache for those users who are setting Pulumi up in CI/CD.

To rectify this, we need to either:

1. Provide concrete information back to the user about how to fix this. This might involve intercepting the AWS SDK calls and injecting information about why they're getting this information
2. Consider setting the default value back to `false`

it might be possible to intercept the error message, which currently looks a bit like this:

```
Error: NoCredentialProviders: no valid providers in chain
    caused by: EnvAccessKeyNotFound: failed to find credentials in the environment.
    SharedCredsLoad: failed to load profile, .
    EC2RoleRequestError: no EC2 instance role found
    caused by: RequestCanceled: EC2 IMDS access disabled via AWS_EC2_METADATA_DISABLED env var
``` 

But this comes from the AWS SDK, so we'd need to trap the error

SharpEdgeMarshall:
ATM is not possible to override through ENV VAR, for some reason setting `AWS_EC2_METADATA_DISABLED=false` doesn't work.
We had to change all our Providers and add `aws:skipMetadataApiCheck false`

seunggs:
Hi @SharpEdgeMarshall - I'm running into this issue as I'm trying to implement Pulumi Automation API. Is the only solution then to create an explicit AWS provider with { skipMetadataApiCheck: false } in the CustomResourcesOptions? Or is there a simpler solution?

SharpEdgeMarshall:
@seunggs there are no simpler solutions that i know atm sorry.

seunggs:
No worries - I just misunderstood the solution. I didn't know you meant adding skipMetadataApiCheck: false as Pulumi config. That's a simple solution! :) Thanks for sharing this.

SharpEdgeMarshall:
Pay attention that if you put it in the config it works only for the
default provider.

[...]

PR #1288, opened 2021-01-05
Setting Provider defaults to remove slow calls to AWS STS and Metadata

stack72:
Fixes: #873

* `skipCredentialsValidation` now defaults to `true`.
* `skipGetEc2Platforms` now defaults to `true`.
* `skipMetadataApiCheck` now defaults to `true`.
* `skipRegionValidation` now defaults to `true`.

It's important to note that this does not require the replacement of either default or named provider:

```
pulumi up --yes
Previewing update (dev)

View Live: https://app.pulumi.com/stack72/provider-change/dev/previews/cb4c2428-a6a2-40fe-aabc-814509e782a4

     Type                     Name                 Plan
 +   pulumi:pulumi:Stack      provider-change-dev  create
 +   ├─ pulumi:providers:aws  my-provider          create
 +   ├─ aws:s3:Bucket         stack72-bucket       create
 +   └─ aws:s3:Bucket         stack72-bucket-2     create

Resources:
    + 4 to create

Updating (dev)

View Live: https://app.pulumi.com/stack72/provider-change/dev/updates/13

     Type                     Name                 Status
 +   pulumi:pulumi:Stack      provider-change-dev  created
 +   ├─ pulumi:providers:aws  my-provider          created
 +   ├─ aws:s3:Bucket         stack72-bucket-2     created
 +   └─ aws:s3:Bucket         stack72-bucket       created

Outputs:
    bucket1Name: "stack72-bucket-3021538"
    bucket2Name: "stack72-bucket-2-e5e84ca"

Resources:
    + 4 created

Duration: 19s


~/code/provider-change
pulumi up
Previewing update (dev)

View Live: https://app.pulumi.com/stack72/provider-change/dev/previews/93b423fc-c366-4c44-a852-b2c613221470

     Type                     Name                 Plan       Info
     pulumi:pulumi:Stack      provider-change-dev
 ~   └─ pulumi:providers:aws  my-provider          update     [diff: +skipCredentialsValidation,skipGetEc2Platforms,skipMetadataApiCheck,skipRegionValidation]

Resources:
    ~ 1 to update
    3 unchanged

Do you want to perform this update? yes
Updating (dev)

View Live: https://app.pulumi.com/stack72/provider-change/dev/updates/14

     Type                     Name                 Status      Info
     pulumi:pulumi:Stack      provider-change-dev
[...]

Issue #873, opened 2020-02-11
AWS metadata and STS API calls are slow


joeduffy:
During the configuration of the AWS provider, it calls into multiple AWS APIs to check various endpoints and identity metadata. Times vary quite a bit depending on the speed of your network, however it's not uncommon for these to add up to **10-15 seconds** of lag before an update even begins running.

Here are the specific calls:

* https://github.com/hashicorp/aws-sdk-go-base/blob/91fe5671bc1bd585d180bba8954ef56d48bef22b/awsauth.go#L270-L294
* https://github.com/hashicorp/aws-sdk-go-base/blob/91fe5671bc1bd585d180bba8954ef56d48bef22b/session.go#L104-L109
* https://github.com/hashicorp/aws-sdk-go-base/blob/91fe5671bc1bd585d180bba8954ef56d48bef22b/session.go#L131-L139
* https://github.com/hashicorp/aws-sdk-go-base/blob/91fe5671bc1bd585d180bba8954ef56d48bef22b/session.go#L141-L158

Note that you can set config variables to skip this logic:

* `pulumi config set aws:skipCredentialsValidation true`
* `pulumi config set aws:skipMetadataApiCheck true`
* `pulumi config set aws:skipRequestingAccountId true`

For some reason, the provider seems to call these APIs _twice_, one of which ignores the config. @lukehoban posited that this could be due to the way we do a prepass over configuration to validate and check for defaults. If so, that seems like it's a bug that we do it without having first applied the configuration.

Note also that you can set `AWS_METADATA_TIMEOUT=0` which shortens the timeouts of the AWS metadata API calls and does have a small noticeable effect.

I don't know precisely what to do here, but we could consider setting our own defaults differently than the underlying Terraform provider. I don't know enough about what those APIs are doing -- it appears, for instance, that the metadata API check is determining whether the update is happening from within an AWS data center (though why the code needs to know that, I'm not quite sure).

uLan08:
I encountered an issue somewhat related to this. I was trying to spin up the basic example from `pulumi new aws-javascript` but `pulumi up` was just hanging indefinitely. The most I waited was probably 5 mins.

Setting `pulumi config set aws:skipCredentialsValidation true` fixed it for me,

Here are some of my environment info in case you need them.
```
▶ node -v
v10.16.3

▶ pulumi version
v1.10.1

"@pulumi/pulumi": "^1.0.0",
"@pulumi/aws": "^1.0.0",
[...]

PR #3960, opened 2024-05-17
Do not skip metadata API check by default

t0yv0:
This PR explores reverting the default `aws:skipMetadataApiCheck=false` setting to enable the provider to be able to seamlessly authenticate against an IMDS(v2) endpoints in the AWS environment. It appears that doing so no longer slows down the provider startup time perceptibly.  The way I tested the speed delta was by measuring local empty preview of an AWS s3 Bucket using AWS_PROFILE authentication with local <-> us-east-1 there is no perceptible difference.

Fixes: https://github.com/pulumi/pulumi-aws/issues/1692

An integration test is added that exercises `pulumi preview` on an EC2 instance with IMDSv2 and asserts that the provider can authenticate successfully. 

Background:

- https://github.com/pulumi/pulumi-aws/issues/873
- https://github.com/pulumi/pulumi-aws/pull/1288

### DIFF
--- a/provider/configure_test.go
+++ b/provider/configure_test.go
@@ -110,7 +110,6 @@ func TestCheckConfigFastWithCustomEndpoints(t *testing.T) {
 	      "s3UsePathStyle": "true",
 	      "secretKey": "*",
 	      "skipCredentialsValidation": "true",
-	      "skipMetadataApiCheck": "true",
 	      "skipRegionValidation": "true",
 	      "skipRequestingAccountId": "true",
 	      "version": "6.5.0"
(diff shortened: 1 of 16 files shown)


t0yv0:
I've got the newly written test pass locally but it unfortunately is tripping up some other tests in the suite, will need another pass. Also as a side note in my VA -> us-east-1 setup I'm only getting 2MB/s from CopyFile and only slightly more up to 2.5MB/s from uploading to S3, this is a bit disappointing:

```
  helper-bucket:
    type: aws:s3:Bucket

  provider-obj:
    type: aws:s3:BucketObjectv2
    properties:
      bucket: ${helper-bucket.bucket}
      source:
        fn::fileAsset: ./pulumi-resource-aws
      etag:
        fn::invoke:
          function: std:filemd5
          arguments:
            input: ./pulumi-resource-aws
          return: result
```

t0yv0:
Disregard the above comment, had to troubleshoot and fix local home network

t0yv0:
I've only tested  IMDS(v2) not IMDS(v1) but I have no reasons to believe that reverting to defaults wouldn't fix that as well.

corymhall (approved):

PR #3970, opened 2024-05-21
fix URL in "learn more" text

tgummerer:
This URL has updated, let's use the new URL in the help text here.


VenelinMartinov (approved):
Thanks for fixing!

Issue #2206, opened 2022-11-09
Artificial limiting of Lambda Function creation


jaxxstorm:
### What happened?

The underlying bridged provider artificially limits the number of lambda functions that can be uploaded in parallel.

The cause of this is https://github.com/hashicorp/terraform/issues/9364 and the fix was implemented in here:

To resolve this, some logic was implemented to create a `mutex.Lock` was added to the lambda function upload, which slows down the provisioning of lambdas. You can see the effect of this here:

https://github.com/hashicorp/terraform/pull/9667/files

### Steps to reproduce

```
import pulumi
import pulumi_aws as aws

 for i in range(50):
     name = f'pulumi-{str(i).rjust(2,"0")}'
     aws.lambda_.Function(
         name,
         aws.lambda_.FunctionArgs(
             name=name,
             code=pulumi.FileArchive("hello-world.zip"),
             role="arn:aws:iam::616138583583:role/lbriggs-lambda-test",
             handler="hello-world.handler",
             runtime="nodejs12.x",
         ),
     )
```

### Expected Behavior

It creates the functions in parallel

### Actual Behavior

This takes around 4/5 minutes and you can observe the creation slowing down and operating in serial

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

lukehoban:
Note that this issue is tracking us making a decision on whether to make a different decision than upstream on this behaviour - so is not "awaiting-upstream".  

The current locking incurs a very significant performance penalty in cases like the one above, and provides no way to work around. Whereas the original issue that this locking was introduced to fix is much rarer, and more within the control of the user who can apply `--parallel` to limit the parallelism (and thus concurrent memory usage).

mikhailshilkov:
We may want to prioritize a decision here. #3740 reported the same issue here, however, StapStart+Publish settings make the code updates much longer (minutes instead of seconds), so the whole UX get much worse.

flostadler:
I'm leaning towards taking a different decision than upstream and consequently removing the serialization for lambda upload.

The underlying issue that was fixed upstream (https://github.com/hashicorp/terraform/issues/9364) only surfaces when dealing with many (10+) Lambdas in a memory constrained environment. Additionally, this fix is ~8 years old. Worth checking if this is still reproducible. For example, the Lambda resource in upstream was switched to the AWS SDK v2 since then.

[...]

PR #3976, opened 2024-05-23
Change Lambdas to create and update in parallel

flostadler:
The root cause behind why the serialization of Lambda creation/update was introduced upstream is excessive memory usage (see https://github.com/hashicorp/terraform/issues/9364).
After investigation we found that this is caused by the HTTP request logging middleware. It logs the lambda archive as a base64 encoded string. In order to do so, multiple copies of the body are created in memory, which leads to memory bloating.
This change fixes that by redacting the body in the logs for the Create/Update Lambda calls.

The PR introduces two patches. One removes the Lambda serialization and the other fixes the HTTP request logging middleware for the Lambda `CreateFunction` and `UpdateFunctionCode` operations.
After this, Lambdas are created/updated in parallel and don't suffer from excessive memory usage. Users can still limit the parallelism with the CLI flag `--parallel` if they wish so.

Relates to #2206


patches/0060-Create-Logging-Middleware-for-Lambda-service-that-do.patch
```diff
@@ -0,0 +1,158 @@
...
++// ID is the middleware identifier.
++func (r *wrappedRequestResponseLogger) ID() string {
++	return "PULUMI_AWS_RequestResponseLogger"
++}
++
++func NewWrappedRequestResponseLogger(wrapped middleware.DeserializeMiddleware) middleware.DeserializeMiddleware {
++	return &wrappedRequestResponseLogger{wrapped: wrapped}
++}
++
++//go:linkname decomposeHTTPResponse github.com/hashicorp/aws-sdk-go-base/v2.decomposeHTTPResponse
```
flostadler, line 57: I wanna highlight this. If we don't do this, we need to fork aws-sdk-go-base or not log the response at all.

In case upstream changes the symbol, we catch it during compile time

mjeffryes:
This could be a nice win for our users, but it might be worth proposing an upstream PR before making a patch.

flostadler:
> Details

I'm already working on preparing an upstream PR, but I'd propose that we still go forward with the patch for the time being because this change will touch two different repos upstream (hashicorp/aws-sdk-go-base & hashicorp/terraform-provider-aws).
My hunch is that this will take a substantial amount of time to get merged in (even small patches for panics take weeks). And in case it goes faster we can remove the patch again. Wdyt?

patches/0060-Create-Logging-Middleware-for-Lambda-service-that-do.patch
```diff
@@ -0,0 +1,158 @@
...
+@@ -34,6 +35,19 @@ func (p *servicePackage) NewClient(ctx context.Context, config map[string]any) (
+ 		if endpoint := config[names.AttrEndpoint].(string); endpoint != "" {
+ 			o.BaseEndpoint = aws_sdkv2.String(endpoint)
+ 		}
++
++		// Switch out the terraform http logging middleware with a custom logging middleware that does not log the
++		// lambda code. Logging the lambda code leads to memory bloating because it allocates a lot of copies of the
++		// body
++		o.APIOptions = append(o.APIOptions, func(stack *middleware.Stack) error {
++			loggingMiddleware, err := stack.Deserialize.Remove("TF_AWS_RequestResponseLogger")
```
flostadler, line 148: If the name of the logging middleware ever changes it'll fail during the integration tests

patches/0059-Parallelize-Lambda-Function-resource-operations.patch
```diff
@@ -0,0 +1,47 @@
...
+ 	FunctionVersionLatest = "$LATEST"
+-	mutexKey              = `aws_lambda_function`
+ 	listVersionsMaxItems  = 10000
+ )
+ 
+@@ -481,11 +480,6 @@ func resourceFunctionCreate(ctx context.Context, d *schema.ResourceData, meta in
+ 	}
+ 
+ 	if v, ok := d.GetOk("filename"); ok {
+-		// Grab an exclusive lock so that we're only reading one function into memory at a time.
```
t0yv0, line 28: Nice.

patches/0060-Create-Logging-Middleware-for-Lambda-service-that-do.patch
```diff
@@ -0,0 +1,158 @@
...
++}
++
++func NewWrappedRequestResponseLogger(wrapped middleware.DeserializeMiddleware) middleware.DeserializeMiddleware {
++	return &wrappedRequestResponseLogger{wrapped: wrapped}
++}
++
++//go:linkname decomposeHTTPResponse github.com/hashicorp/aws-sdk-go-base/v2.decomposeHTTPResponse
++func decomposeHTTPResponse(ctx context.Context, resp *http.Response, elapsed time.Duration) (map[string]any, error)
++
++func (r *wrappedRequestResponseLogger) HandleDeserialize(ctx context.Context, in middleware.DeserializeInput, next middleware.DeserializeHandler,
```
t0yv0, line 60: Guessing this was inlined and edited from upstream ? I think that's great and worth doing, just would be helpful to indicate that in the comment, and highlight the edits, just in case. Since we forked it the upstream changes to this handler will not propagate anymore but that sounds ok.

patches/0060-Create-Logging-Middleware-for-Lambda-service-that-do.patch
t0yv0: Nice !!

t0yv0 (approved):
Thank you! This is massively helpful. Apologies for the long review cycles here, I should have gotten to this much sooner.

t0yv0:
It fixes https://github.com/pulumi/pulumi-aws/issues/2206 right or there's some work remaining?

flostadler:
> It fixes #2206 right or there's some work remaining?

It fixes it! The added test would fail without the patches

PR #4027, opened 2024-06-05
Fix rds.Instance v5 to v6 migration

t0yv0:
Enabling PlanResourceChange=true on v6 of pulumi-aws provider regresses on replacing rds.Instance. It appears that the problem here is that the code has not been completely removing the deprecated "name" property from the old state. This is now fixed and covered by tests.


provider/resources.go
```diff
@@ -3094,6 +3094,7 @@ func ProviderFromMeta(metaInfo *tfbridge.MetadataInfo) *tfbridge.ProviderInfo {
 				},
 				TransformFromState: func(ctx context.Context, state resource.PropertyMap) (resource.PropertyMap, error) {
 					if _, ok := state["dbName"]; ok {
+						delete(state, "name")
```
t0yv0, line 3097: It seems that under PlanResourceChange, having "name" hanging around confuses the diff to think it's changing from "something" to "" and that should be a replacement.

VenelinMartinov (approved):
looks good

iwahbe (commented):
LGTM

PR #4028, opened 2024-06-05
Refactor TestS3BucketObjectDeprecation

t0yv0:
This test was over-fitting on some implementation details and blocking the rollout of PlanResourceChange.

The refactor moves it to a full integration test that should hopefully be more robust.


iwahbe (approved):

flostadler (approved):
LGTM (Seems like there's an unused import though)

PR #4062, opened 2024-06-12
Simplify CI hooks

danielrbradley:
Broken into 3 smaller commits for review.

## 1. Fix invalid matrix condition in prerequisites

Prerequisites does not run in a matrix, so there's no need to check for dotnet.

## 2. Move clearing disk space in custom provider_test job

Removes the need to install dotnet for a second time.

## 3.Remove scripts & unshallow

- Set version for consistency with other jobs even though we likely won't use it.
- We're not running anything from the scripts repo in these test jobs. The only previous use of the scripts was for working tree clean checks, which we only perform during the build_sdk step, and should now use the `pulumi/git-status-check-action@v1` action for instead.
- Remove unshallow clone as we no longer need to fetch all tags to calculate the version via pulumictl.

### DIFF
--- a/.ci-mgmt.yaml
+++ b/.ci-mgmt.yaml
@@ -80,8 +80,9 @@ extraTests:
       with:
         ref: ${{ env.PR_COMMIT_SHA }}
         submodules: true
-    - name: Unshallow clone for tags
-      run: git fetch --prune --unshallow --tags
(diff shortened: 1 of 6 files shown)


flostadler (approved):
Nice! LGTM

t0yv0 (approved):

PR #4086, opened 2024-06-17
Directly inject `aws:rds:Instance.name` into the schema

iwahbe:
This bypasses the need to call `shim.SchemaMap.Set` which has been removed in the latest version of the bridge. Since `"name"` is *never* witnessed by the underlying provider, it should not matter that the provider doesn't have the phony "name" field in it's map anymore.

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/2101


corymhall (approved):
Nice!

PR #4095, opened 2024-06-21
fix!: unable to create fsx for OpenZfs Multi-AZ file system

### LINKED ISSUE #3034: OpenZfsFileSystem.subnetIds is a string, not a string array
### What happened?

Trying to create an OpenZfsFileSystem with multiple subnet ids.

### Example

The example in the docs will not compile:

https://www.pulumi.com/registry/packages/aws/api-docs/fsx/openzfsfilesystem/

### Output of `pulumi about`

CLI          
Version      3.76.1
Go Version   go1.20.6
Go Compiler  gc

Plugins
NAME    VERSION
nodejs  unknown

Host     
OS       amazon
Version  2
Arch     x86_64

This project is written in nodejs: executable='/usr/bin/node' version='v18.18.2'

Additional info:
```
    "@pulumi/aws": "^6.7.0",
    "@pulumi/pulumi": "^3.76.1",
```

### Additional context

The type of `subnetIds` here should be `string[]`, not `string`. The example code on the same page also expects this to be the case.

https://github.com/pulumi/pulumi-aws/blob/a40c588e1160a51301619512299a0650bdc028bb/sdk/nodejs/fsx/openZfsFileSystem.ts#L148

### Contributing

### LINKED ISSUE #3106: Unable to create Fsx for OpenZFS Multi-AZ file system
### What happened?

When trying to create Fsx filesystem for OpenZFS with deployment type: Multi-AZ, I received the following error:
```
  aws:fsx:OpenZfsFileSystem (fsxOpenZfs):
    error: 1 error occurred:
        * creating FSx for OpenZFS File System: BadRequest: 1 validation error detected: Value '[subnet-xxxxxxxxxxxxxxxxx,
           subnet-xxxxxxxxxxxxxxxxx]' at 'subnetIds' failed to satisfy constraint: Member must satisfy constraint: 
           [Member must have length less than or equal to 24, Member must have length greater than or equal to 15, 
           Member must satisfy regular expression pattern: ^(subnet-[0-9a-f]{8,})$]
```

### Example

```

fsx_open_zfs = fsx.OpenZfsFileSystem(
                        "fsxOpenZfs",
                        subnet_ids="subnet-xxxxxxxxxxxxxxxxx,subnet-xxxxxxxxxxxxxxxxx",
                        deployment_type="MULTI_AZ_1",
                        throughput_capacity=320,
                        automatic_backup_retention_days=7,
                        storage_capacity=500,        
                        preferred_subnet_id="subnet-xxxxxxxxxxxxxxxxx",
                        disk_iops_configuration=fsx.OpenZfsFileSystemDiskIopsConfigurationArgs(
                            iops=3000,
                            mode="USER_PROVISIONED"
                        ),
                        root_volume_configuration=fsx.OpenZfsFileSystemRootVolumeConfigurationArgs(
                            data_compression_type="LZ4",
                            record_size_kib=128,
                        ),
                        weekly_maintenance_start_time="2:00:00"
                    ),
```

### Output of `pulumi about`

CLI          
Version      3.94.2
Go Version   go1.21.4
Go Compiler  gc

Plugins
NAME    VERSION
aws     6.13.0
kubernetes    4.5.4

Host     
OS       darwin
Version  14.1
Arch     arm64

This project is written in python: executable='/Library/Frameworks/Python.framework/Versions/3.11/bin/python3' version='3.11.5'

Backend        
Name           << REDACTED >>
URL            s3://<< REDACTED >>
User           xxxxxxxx
Organizations  xxxxxxxx
Token type     personal

Dependencies:
NAME            VERSION
@pulumi/aws     6.13.0
@pulumi/pulumi  3.94.2

### Additional context
Passing a list instead of a string result in a type error.
Passing only single subnet to the subnetIds resulted in another error/;

```
  aws:fsx:OpenZfsFileSystem (fsxOpenZfs):
    error: 1 error occurred:
        * creating FSx for OpenZFS File System: BadRequest: Exactly 2 subnet IDs are required for MULTI_AZ_1.
```

### Contributing

corymhall:
This PR removes the `MaxItemsOne` config for the `subnetIds` property. Originally this was a `MaxItemsOne` property in upstream, but it was removed when Multi-AZ support was added by AWS. In Terraform removing `MaxItemsOne` is not a breaking change, but for Pulumi it is.

Because it looks like there is very low usage of this resource, and the change enables a (maybe) more common use case (Multi-AZ filesystem) we have decided to take the breaking change now.

In order to upgrade users will only need to update their code so that `subnetIds` is now a list. For example,

From
```ts
const test = new aws.fsx.OpenZfsFileSystem("test", {
    storageCapacity: 64,
    subnetIds: test1.id,
    deploymentType: "SINGLE_AZ_1",
    throughputCapacity: 64,
});
```

To
```ts
const test = new aws.fsx.OpenZfsFileSystem("test", {
    storageCapacity: 64,
    subnetIds: [test1.id],
    deploymentType: "SINGLE_AZ_1",
    throughputCapacity: 64,
});
```

Because we are including the `TransformFromState` function, users should not need to make any changes to the state themselves.

BREAKING CHANGE: `fsx.OpenZfsFileSystem.subnetIds` now accepts a list instead of a string

closes #3106, closes #3034


provider/provider_test.go
```diff
@@ -110,3 +211,112 @@ func maxDuration(dur time.Duration, t *testing.T, test func(t *testing.T)) {
...
+			delete(resourcePlan, "seed")
+
+			if goal, ok := resourcePlan["goal"]; ok {
+				g := goal.(map[string]interface{})
+				if g["type"] == "pulumi:providers:aws" {
+					if inputDiff, ok := g["inputDiff"]; ok {
+						idiff := inputDiff.(map[string]interface{})
+						if adds, ok := idiff["adds"]; ok {
+							a := adds.(map[string]interface{})
+							delete(a, "region")
```
flostadler, line 281: Why do we delete the region, that shouldn't change, right?

provider/provider_yaml_test.go
```diff
@@ -305,6 +305,83 @@ func TestNonIdempotentSnsTopic(t *testing.T) {
...
+
+	var (
+		providerName    string = "aws"
+		baselineVersion string = "6.41.0"
+	)
+	cwd, err := os.Getwd()
+	assert.NoError(t, err)
+	workdir := t.TempDir()
+
+	firstProgram := []byte(fmt.Sprintf(pulumiYaml, "subnetIds: ${MySubnet.id}"))
```
flostadler, line 342: Why don't we do that in `testProviderCodeChanges`? Would be quite a bit of duplication if every test needs to do this

provider/provider_yaml_test.go
```diff
@@ -305,6 +305,83 @@ func TestNonIdempotentSnsTopic(t *testing.T) {
...
+		assertpreview.HasNoChanges(t, res)
+
+		upResult := pulumiTest.Up()
+		fmt.Printf("stdout: %s \n", upResult.StdOut)
+		fmt.Printf("stderr: %s \n", upResult.StdErr)
+	})
+
+	// test that we can deploy a new filesystem with a list of subnetIds
+	// we use a test with a snapshot since this test is only useful the first time, once
+	// we know it works it should continue to work.
```
flostadler, line 370: This assumes there's no changes in AWS and the terraform layers, right?

provider/resources.go
```diff
@@ -2277,10 +2278,25 @@ compatibility shim in favor of the new "name" field.`)
 			"aws_fsx_ontap_file_system":             {Tok: awsResource(fsxMod, "OntapFileSystem")},
 			"aws_fsx_ontap_storage_virtual_machine": {Tok: awsResource(fsxMod, "OntapStorageVirtualMachine")},
 			"aws_fsx_ontap_volume":                  {Tok: awsResource(fsxMod, "OntapVolume")},
-			"aws_fsx_openzfs_file_system":           {Tok: awsResource(fsxMod, "OpenZfsFileSystem")},
-			"aws_fsx_openzfs_snapshot":              {Tok: awsResource(fsxMod, "OpenZfsSnapshot")},
-			"aws_fsx_openzfs_volume":                {Tok: awsResource(fsxMod, "OpenZfsVolume")},
-			"aws_fsx_windows_file_system":           {Tok: awsResource(fsxMod, "WindowsFileSystem")},
+			"aws_fsx_openzfs_file_system": {
+				Tok: awsResource(fsxMod, "OpenZfsFileSystem"),
+				TransformFromState: func(ctx context.Context, state resource.PropertyMap) (resource.PropertyMap, error) {
```
t0yv0, line 2283: Curious if you found this necessary? If I recall correctly the bridge might be already applying this transformation generically to permit lenient state reads to compensate for MaxItems=1 mismatches. If it doesn't at the moment we could consider moving it there anyway.

provider/provider_test.go
```diff
@@ -82,6 +89,100 @@ func testProviderUpgrade(t *testing.T, dir string, opts *testProviderUpgradeOpti
...
+type testProviderCodeChangesOptions struct {
+	firstProgram         []byte
+	secondProgram        []byte
+	firstProgramOptions  []opttest.Option
+	secondProgramOptions []opttest.Option
+}
+
+// testProviderCodeChanges tests two different runs of a pulumi program. This allow you to run
+// pulumi up with an initial program, change the code of the program and then run another pulumi command
+func testProviderCodeChanges(t *testing.T, opts *testProviderCodeChangesOptions) *pulumitest.PulumiTest {
```
flostadler, line 101: Nice!

flostadler (approved):
lgtm

provider/provider_test.go
```diff
@@ -82,6 +89,100 @@ func testProviderUpgrade(t *testing.T, dir string, opts *testProviderUpgradeOpti
...
+		assert.NoError(t, err)
+		tmpPlanFile := filepath.Join(workdir, "plan.json")
+
+		pulumiTest.Preview(optpreview.Plan(tmpPlanFile))
+
+		if equal := planEqual(t, planFile, tmpPlanFile); equal {
+			return
+		}
+
+		pulumiTest.T().Log("Plan is not equal, re-running up")
```
t0yv0, line 178: This is https://www.pulumi.com/docs/concepts/update-plans/ right?

provider/resources.go
corymhall: Yes it was necessary. Without it it showed a resource replacement.

provider/provider_test.go
corymhall: It can change between what you set it to locally and what is set in CI. I ended up changing the test to set the region explicitly, but I thought to keep this in.

provider/provider_test.go
corymhall: Yes!

provider/provider_test.go
```diff
@@ -82,6 +89,100 @@ func testProviderUpgrade(t *testing.T, dir string, opts *testProviderUpgradeOpti
 	assertpreview.HasNoReplacements(t, result)
 }
 
+type testProviderCodeChangesOptions struct {
```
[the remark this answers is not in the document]
corymhall: Yes I think we could upstream this since I think it is a generic enough use case. And yes those are the two main differences.

provider/provider_test.go
```diff
@@ -82,6 +89,100 @@ func testProviderUpgrade(t *testing.T, dir string, opts *testProviderUpgradeOpti
...
+	return secondTest
+}
+
+// pulumiUpWithSnapshot will only run the up portion of the test if the plan has changed since the
+// last time the test was run.
+//
+// This should be used when the plan is a good representation of what you are testing. Sometimes
+// there are issues where the plan is consistent, but the apply fails. In those cases a snapshot test is not
+// a good fit.
+func pulumiUpWithSnapshot(t *testing.T, pulumiTest *pulumitest.PulumiTest) {
```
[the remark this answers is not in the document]
corymhall: I'm experimenting with something here. I think it may be useful to upstream, but I'm not sure yet since I think it would be easy to footgun yourself. I'm trying to figure out ways we can short circuit the test.

I think it is useful for the type of test where everything that you want to test exists in the plan snapshot. In those cases we can short circuit the test and only actually run the test when the plan has changed. In this case I'm testing that subnets as a list is able to be deployed and the snapshot contains this information.

In other cases the snapshot won't contain enough information, for example if the test depends on the value of unknowns the plan won't have those values so will never show a difference.

provider/provider_yaml_test.go
corymhall: which piece are you talking about? The `[]byte()` part? Everything else is specific to this test.

provider/provider_test.go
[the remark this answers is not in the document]
t0yv0: So you're looking for an optimization that doesn't run `pulumi up` when the plan hasn't changed, but does run it if there's changes in plans? This can be interesting. I worry that plan support is not yet solid enough to trust for this, but it can be interesting.

t0yv0 (approved):

provider/provider_yaml_test.go
flostadler: Sorry, yeah the casting part. Looking again it's definitely not a lot of duplication. So rather just a nit. Feel free to ignore

provider/provider_test.go
flostadler: But wouldn't change more in this case? E.g. `ARNs` or `URLs`

corymhall: Hmm that's true. I"ll remove it.

PR #4108, opened 2024-06-25
Enable TypedDict input types for the Python SDK

julienp:
Enable TypedDict input types for the Python SDK. The only manual change is to set the InputTypes flag in the python language features of the schema. This is done here https://github.com/pulumi/pulumi-aws/pull/4108/commits/25a4cec9ac577b74d56bbad600479032ec86559d. This flag is currently opt in, but will default to classes-and-dicts in the future.

I ran `make tfgen build_sdks` afterwards.

Epic https://github.com/pulumi/pulumi/issues/12689


sdk/python/pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
   name = "pulumi_aws"
   description = "A Pulumi package for creating and managing Amazon Web Services (AWS) cloud resources."
-  dependencies = ["parver>=0.2.1", "pulumi>=3.0.0,<4.0.0", "semver>=2.8.1"]
+  dependencies = ["parver>=0.2.1", "pulumi>=3.0.0,<4.0.0", "semver>=2.8.1", "typing-extensions>=4.11; python_version < \"3.11\""]
```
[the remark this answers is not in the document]
julienp: There are no changes required in the Pulumi Python SDK to support TypedDicts, the changes are purely in the Provider SDKs. The reason for this is that TypedDicts are purely a typechecking concept and don't exist at runtime in Python. Previously you could already pass in dictionaries instead of args classes, but these were always untyped. Now they are typed.

Previously inputs looked like his

```python
# pulumi.InputType = Union[T, Mapping[str, Any]]

Optional[pulumi.Input[pulumi.InputType['AnalyzerConfigurationArgs']]]
```
and with this change, we essentially inline the `InputType` Union with a stricter type:
```python
# class AnalyzerConfigurationArgsDict(TypedDict):
#    unused_access: NotRequired[pulumi.Input['AnalyzerConfigurationUnusedAccessArgsDict']]
# ...

Optional[pulumi.Input[Union['AnalyzerConfigurationArgs', 'AnalyzerConfigurationArgsDict']]]
```

At runtime, when you pass in a concrete instance of a dictionary for an input, it is just a standard `dict`.

sdk/nodejs/ec2/securityGroup.ts
```diff
@@ -176,8 +176,8 @@ import * as utilities from "../utilities";
  *         input: exampleAwsVpcEndpoint.securityGroupIds,
  *     }).then(invoke => invoke.result),
  * }});
- * const exampleResourceProvisioner0 = new command.local.Command("exampleResourceProvisioner0", {create: `            aws ec2 modify-vpc-endpoint --vpc-endpoint-id ${exampleAwsVpcEndpoint.id} --remove-security-group-ids ${_default.id}
- * `}, {
+ * const exampleResourceProvisioner0 = new command.local.Command("exampleResourceProvisioner0", {create: _default.then(_default => `            aws ec2 modify-vpc-endpoint --vpc-endpoint-id ${exampleAwsVpcEndpoint.id} --remove-security-group-ids ${_default.id}
+ * `)}, {
```
danielrbradley, line 180: This diff looks a bit odd and it appears to be being reset during the CI build. Perhaps something was wrong during your local build?

julienp: Yeah I am a little confused where this is coming from, I pushed again without this.

danielrbradley (approved):

PR #4120, opened 2024-06-26
chore: run upstream provider-lint

### LINKED ISSUE #4110: Run upstream linter after applying patches
As reported in [Improper partition used when generating SNS Topic ARN· pulumi-aws/3926](https://github.com/pulumi/pulumi-aws/issues/3926), one of our local patches of the upstream TF provider introduced a hardcoded partition into the ARN, which prevented the resource from working with US Gov or China aws partitions. There's actually a linter in the upstream TF provider (https://github.com/hashicorp/terraform-provider-aws/tree/main/.ci/providerlint) with a rule for this exact problem (https://github.com/hashicorp/terraform-provider-aws/tree/main/.ci/providerlint/passes/AWSAT005). 

As a remediation, we'd like to start running the upstream linter on upstream after our patches to hopefully catch other problems like this in the future.

corymhall:
This adds a step for running the upstream `provider-lint` make target.

As part of this I had to fix some of the patches which violated some
lint rules.

**0009-Add-ECR-credentials_data_source.patch**
- `ForceNew` does not apply to data sources

**0032-DisableTagSchemaCheck-for-PF-provider.patch**
- Schema have to have a `Type`
- Also needed to add a ignore for `S013` which forces `Computed`,
  `Optional` or `Required` to be set. Looks like it can't recognize the
  `tagsComputed` var

**0034-Fail-fast-when-PF-resources-are-dropped.patch**
- Added a lint ignore for a rule which doesn't allow panics

**0050-Normalize-retentionDays-in-aws_controltower_landing_.patch**
- This test doesn't actually need a region or partition so replacing
  with a placeholder

closes #4110


patches/0050-Normalize-retentionDays-in-aws_controltower_landing_.patch
```diff
@@ -151,7 +151,7 @@ index 0000000000..a8bb57939e
 +	actual, err := resourceLandingZoneNormalizeManifest(`
 +	{
 +	  "governedRegions": [
-+	    "ap-southeast-2"
++	    "REGION"
```
flostadler, line 154: Does the upstream linter catch hardcoded regions/partitions?

patches/0032-DisableTagSchemaCheck-for-PF-provider.patch
```diff
@@ -67,6 +67,31 @@ index 0000000000..f790acb4e2
...
+index 35202ebd58..28995620f2 100644
+--- a/internal/provider/provider_tagcheck.go
++++ b/internal/provider/provider_tagcheck.go
+@@ -18,9 +18,18 @@ func schemaMapForTagsChecking(ctx context.Context, r *schema.Resource, tagsCompu
+ 	switch flag := flag.(type) {
+ 	case bool:
+ 		if flag {
++			//lintignore:S013
+ 			return map[string]*schema.Schema{
+-				names.AttrTags:    &schema.Schema{Computed: tagsComputed},
```
flostadler, line 80: That file is created in this [patch](https://github.com/pulumi/pulumi-aws/blob/8771ec64099fe5c257fd343e0f9a976a5458a84a/patches/0030-Optimize-startup-performance.patch#L29). Imho we should correct it there, otherwise we'll make it harder to maintain that patch

patches/0050-Normalize-retentionDays-in-aws_controltower_landing_.patch
corymhall: Yep!

patches/0032-DisableTagSchemaCheck-for-PF-provider.patch
corymhall: oh woops, I picked the wrong one! Good catch

corymhall: Updated!

flostadler (approved):
nice!

PR #4136, opened 2024-06-28
Readme update

interurban:
Adding Pulumi logo and developer resources links to the readme


pulumi-aws.sln
iwahbe, line 1: I don't think we want to check in this file.

iwahbe (commented):
The readme changes look good.

t0yv0 (approved):

Issue #3727, opened 2024-03-25
PATCH 0047-securitylake_subscriber-tags_all-patch.patch


t0yv0:
Tracking 0047-securitylake_subscriber-tags_all-patch.patch

Pertains to tags re-interpretation.

corymhall:
merged into `7.0.0-alpha`

Issue #3725, opened 2024-03-25
PATCH 0045-Do-not-Compute-tags_all-of-aws_bedrock_provisioned_m.patch


t0yv0:
Tracking Do-not-Compute-tags_all-of-aws_bedrock_provisioned_m.patch

Pertains to tags re-interpretation.

corymhall:
merged into `7.0.0-alpha`

Issue #3721, opened 2024-03-25
PATCH 0040-Patch-osis_pipeline-tags-flags.patch


t0yv0:
Tracking Patch-osis_pipeline-tags-flags.patch

More patches needed for tags re-interpretation.

corymhall:
merged into `7.0.0-alpha`

Issue #3717, opened 2024-03-25
PATCH Fix-tags_all-Computed-for-PF-resources.patch


t0yv0:
More schema changes pertaining to tags re-interpretation in pulumi-aws.

corymhall:
merged into `7.0.0-alpha`

PR #4151, opened 2024-07-01
Automate removing Computed on tags_all for PF resources

t0yv0:
Fixes https://github.com/pulumi/pulumi-aws/issues/2962

The change is related to https://github.com/pulumi/pulumi-aws/issues/3709 and most of it is folded into that patch; with this PR the tags_all attribute is changing to be non-Computed as before, but the change is no longer a source-level change but is a quick runtime pass instead. This removes the burden of patching the provider repeatedly for this concern. 

Obsolete patches are removed.

Fixes https://github.com/pulumi/pulumi-aws/issues/3711
Fixes https://github.com/pulumi/pulumi-aws/issues/3714
Fixes https://github.com/pulumi/pulumi-aws/issues/3715
Fixes https://github.com/pulumi/pulumi-aws/issues/3717
Fixes https://github.com/pulumi/pulumi-aws/issues/3721
Fixes https://github.com/pulumi/pulumi-aws/issues/3725
Fixes https://github.com/pulumi/pulumi-aws/issues/3727

All removed patches:

patches/0029-Fix-markTagsAllNotComputedForResources-to-recognize-.patch 
patches/0032-DisableTagSchemaCheck-for-PF-provider.patch 
patches/0033-Run-scripts-patch_computed_only.sh-to-patch-eks-pod_.patch 
patches/0035-Fix-tags_all-Computed-for-PF-resources.patch 
patches/0039-Patch-osis_pipeline-tags-flags.patch 
patches/0041-Do-not-Compute-tags_all-of-aws_bedrock_provisioned_m.patch 
patches/0043-securitylake_subscriber-tags_all-patch.patch  
patches/0044-Patch-tags-ComputedOnly-for-m2-resources.patch
patches/0053-Patch-tags-ComputedOnly-on-bedrockagent-and-other-mo.patch
patches/0055-Fix-tags_all-Computed-for-aws_datazone_domain.patch
patches/0058-Fix-tags_all-Computed-for-PF-resources.patch
patches/0059-Fix-tags_all-Computed-for-PF-resources.patch
patches/0060-Fix-tags_all-Computed-for-PF-resources.patch
patches/0062-Patch-tags-ComputedOnly-for-appfabric-networkmonitor.patch


patches/0027-Do-not-compute-tags_all-at-TF-level.patch
```diff
@@ -336,21 +203,185 @@ index 0bc956a747..e0636e99ce 100644
...
++import (
++	"context"
++
++	"github.com/hashicorp/terraform-plugin-framework/function"
++	"github.com/hashicorp/terraform-plugin-framework/provider"
++	"github.com/hashicorp/terraform-plugin-framework/resource"
++	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
++)
++
++type transformedFrameworkProvider struct {
```
t0yv0, line 223: There is an interesting Go problem here I'd love eyes on cc @blampe  - the intent is to override the response of provider.Provider interface at some depth that is all I'd like to do is change this response :

```
provider.Resources(ctx)[17 /*arbitrary index*/]().Schema(ctx, request, response)
```

Applying the technique of embedding interfaces into structs and overriding.

However..

There is an open set of extra interfaces like resource.ResourceWithModifyPlan though that the framework might be probing for. 

The danger with struct wrappers it would seem is that the wrapped provider or resource will no longer pass a type test on these interfaces and possibly lead to invalid behavior? I'm implementing all these interfaces here that I could find, but there still is the risk that new interfaces will be added upstream and this code will need updates and fail quietly. Is there anything better we can do?

corymhall (approved):

patches/0027-Do-not-compute-tags_all-at-TF-level.patch
flostadler: I'm a bit worried about adding this without running the full set of upstream tests.
I don't think we have enough coverage in our existing tests to catch potential problems here.

flostadler (commented):
Change looks good, but I'm a bit worried about merging it right now without running upstream tests in CI

t0yv0:
That's reasonable, we could build https://github.com/pulumi/pulumi-aws/issues/3022 first and then rearrange things so that the provider modifications happen so that upstream tests exercise our patches, it could be very valuable here.

t0yv0:
Notes from team discussion: we can try running some upstream tests over PF, but probably targeting all modules is too hard; we'll target a few modules, exclude upstream tags tests if needed, and unblock this.

corymhall:
Closing since this was removed in v7

PR #4228, opened 2024-07-15
chore: fix flaky test

corymhall:
This test has been failing recently due to a resource with the same name
already existing. I think we have been running into a race condition
between different jobs.

Remove the hardcoded name.


flostadler (approved):
nice!

PR #4232, opened 2024-07-16
chore: testing out nodejs example test splitting

corymhall:


.github/workflows/run-acceptance-tests.yml
```diff
@@ -53,6 +53,41 @@ jobs:
     secrets: inherit
     with:
       version: ${{ needs.prerequisites.outputs.version }}
+  combine_build_artifacts:
```
t0yv0, line 56: Perhaps you know this and are just experimenting but FYI these are auto-generated files, and the team is in the middle of refactoring these workflows elsewhere.

corymhall: yeah I'm just experimenting

PR #4285, opened 2024-07-22
Rewrite GameLift regression test into TypeScript

t0yv0:
It appears that running .NET tests is not fully supported in the release.yml workflow unfortunately. 

Fixes https://github.com/pulumi/pulumi-aws/issues/4275


provider/provider_dotnet_test.go
```diff
@@ -17,7 +17,7 @@ func TestGameLift(t *testing.T) {
 		return
 	}
 
-	ptest := pulumiTest(t, filepath.Join("test-programs", "gamelift-csharp"))
+	ptest := pulumiTest(t, filepath.Join("test-programs", "gamelift-typescript"))
```
corymhall, line 20: Can we move this to the `nodejs` tests?

t0yv0: Good catch, my mistake. Will do.

corymhall (approved):

PR #4298, opened 2024-07-24
chore: TestIMDSAuth is very slow

corymhall:
The `TestIMDSAuth` test frequently takes ~45 minutes to run. Looking into it more it looks like the `command:remote:CopyFile` takes the bulk of the time and I am not sure why. Copying the file to S3 via `BucketObjectV2` and then using the userData to copy the file from S3 during instance start up seems to be significantly faster

The test went from ~40minutes to ~4 minutes.

### DIFF
--- a/provider/provider_yaml_test.go
+++ b/provider/provider_yaml_test.go
@@ -9,6 +9,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"math/rand"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


t0yv0:
Maybe it was installing the command provider and getting throttled? @thomas11 any other guesses?

t0yv0 (approved):

thomas11:
> Maybe it was installing the command provider and getting throttled? @thomas11 any other guesses?

`CopyFile` is deprecated in favor of `CopyToRemote`. One difference is that the latter has concurrent writes enabled for SFTP. That should make a noticeable difference.

Downloading the Command provider could also be a noticeable chunk.

PR #4310, opened 2024-07-26
Add basic instructions for upstream upgrades

t0yv0:
While we improve automation and documentation in the bridge for the upstream upgrade process, here are some notes on the de-facto process of doing these upgrades.


corymhall (approved):
Nice!

PR #4348, opened 2024-08-06
Upgrade CallbackFunction default runtime to Node 20.x

### LINKED ISSUE #4134: Default Node runtime for lambdaMixins is out of LTS (and marked as EOL)
Lambda mixins (magic functions) default to running Node v16 which has been EOL for 9 months.

https://github.com/pulumi/pulumi-aws/blob/896ae14c62ab8527923fcda571d0e3187a440d01/sdk/nodejs/lambda/lambdaMixins.ts#L415C38-L415C56

I discovered this by trying to use `fetch` within an `snsTopic.onEvent` handler. The lambda threw an error at runtime that fetch wasn't defined. Of course, fetch was only adding in node v18.

Would probably be wise to update the default to the current LTS (v18).

t0yv0:
This is a breaking change for users of Node automatically provisioned lambdas, moving them from the Node 16.x to Node 20.x runtime without an ability to opt out.
 
Prior to this change, Lambda functions provisioned in Node with CallbackFunction and helper methods such as Bucket.onObjectRemoved, Bucket.onObjectCreated or sns.Topic.onEvent used Node 16.x runtime by default. Per [lambda-runtimes.html](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-supported) this is now deprecated by AWS. After the change, the new default runtime is Node 20.x.

Fixes #4134 

The change is supported by strengthening a test to utilize `fetch()` API that was made available in 18.x and exercise the callback function at runtime.


sdk/nodejs/lambda/lambdaMixins.ts
```diff
@@ -406,13 +406,11 @@ export class CallbackFunction<E, R> extends LambdaFunction {
...
         // about with our own values.  This ensures that clients can pass future supported
         // lambda options without us having to know about it.
-        // The default version for Lambda functions now is NodeJS16. As of April 30 2022, Node12
-        // is EOL/ nodeJS 14 is only in maintenance mode so it's best to upgrade to Node16
         const functionArgs = {
             ...args,
             code: code,
             handler: serializedFileNameNoExtension + "." + handlerName,
-            runtime: args.runtime || Runtime.NodeJS16dX,
+            runtime: args.runtime || Runtime.NodeJS18dX,
```
corymhall, line 413: Should we take the opportunity to just upgrade to 20x?

corymhall (approved):
Just 1 suggestion

flostadler (changes requested):
Even though it's a breaking change I think not doing it would be a greater issue for our users.
Function creates using node 16 will be blocked on Feb 28, 2025 and updates will be blocked on Mar 31, 2025.

We should really directly upgrade to node 20 though. The node 18 runtime is still using AL2 under the hood (see [AWS docs](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-supported)) which will be deprecated in June 2025. There's a high chance they'll deprecate Node 18 as well in the near future.

flostadler:
We should mark the Node 16.x runtime as deprecated as well: https://github.com/pulumi/pulumi-aws/blob/master/provider/resources.go#L4541

sdk/nodejs/lambda/lambdaMixins.ts
```diff
@@ -406,13 +406,11 @@ export class CallbackFunction<E, R> extends LambdaFunction {
         // Copy over all option values into the function args.  Then overwrite anything we care
```
flostadler, line 406: Can't put the comment on that line because it's outside of the PR, but we should also update the docs to reflect the new default: https://github.com/pulumi/pulumi-aws/blob/7461d6834b90936a38e80cc4bac2c10ff97499a0/sdk/nodejs/lambda/lambdaMixins.ts#L227

sdk/nodejs/lambda/lambdaMixins.ts
t0yv0: You are right, https://github.com/pulumi/pulumi-aws/issues/4134 was suggesting 18 as the current LTS but https://endoflife.date/nodejs  suggests 20 is the current LTS so that should be reasonable.

flostadler (approved):
Nice!

t0yv0:
PTAL @flostadler addressed the feedback.

> Even though it's a breaking change I think not doing it would be a greater issue for our users.

I suspect just a slightly unexpected upgrade. Just calling it out for the release notes, still doing a minor release on this.

mikhailshilkov:
This PR has been shipped in [release v6.49.0](https://github.com/pulumi/pulumi-aws/releases/tag/v6.49.0).

PR #4352, opened 2024-08-07
Schematize lambda.CallbackFunction overlay and add docs for it

### LINKED ISSUE #4351: Document CallbackFunction resource overlay
The CallbackFunction resource overlay isn't documented in the registry right now. Schematize it and create docs for it.

flostadler:
The CallbackFunction resource overlay was not present in the registry before. This change adds the schema for it and adds the docs for the registry.
The schema is mostly based on the `lambda.Function` resource with modifications to adhere to the typescript class here: https://github.com/pulumi/pulumi-aws/blob/7461d6834b90936a38e80cc4bac2c10ff97499a0/sdk/nodejs/lambda/lambdaMixins.ts#L245

### How to review?
- Check out this PR locally
- Check out https://github.com/pulumi/registry and run the following in the root of that repository:
  - Follow the setup steps here: https://github.com/pulumi/registry?tab=readme-ov-file#installing-dependencies
  - `go install github.com/pulumi/registry/tools/resourcedocsgen@master`
  - `resourcedocsgen docs --schemaFile ../pulumi-aws/provider/cmd/pulumi-resource-aws/schema.json --docsOutDir content/registry/packages/aws/api-docs --packageTreeJSONOutDir static/registry/packages/navs/aws.json --version 6.0.0-alpha.0+dev` (you might have to tweak the path to the schema.json to match your dir layout)
  - `make serve`
- Visit http://localhost:1313/registry/packages/aws/api-docs/lambda/callbackfunction

Fixes #4351


provider/types.go
```diff
@@ -0,0 +1,416 @@
...
+package provider
+
+import (
+	"maps"
+
+	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+)
+
+// extraTypes augment the types defined by the upstream AWS provider
+var extraTypes = map[string]schema.ComplexTypeSpec{
```
flostadler, line 24: I moved this to its own file to make it easier to work with and add additional extraTypes sources (like the overlays)

t0yv0: Thank you so much, I dread working in the super long resources.go.

provider/overlays.go
```diff
@@ -0,0 +1,191 @@
...
+
+var arnToken = fmt.Sprintf("#/types/%s", awsTypeDefaultFile(awsMod, "ARN"))
+
+// resourceOverlays augment the resources defined by the upstream AWS provider
+var resourceOverlays = map[string]schema.ResourceSpec{}
+
+// typeOverlays augment the types defined by the upstream AWS provider
+var typeOverlays = map[string]schema.ComplexTypeSpec{}
+
+//go:embed pkg/overlays/examples/callbackFunction.md
```
t0yv0, line 39: This is a moot point for just one file but if we're doing it a lot we can avoid shipping this into pulumi-resource-aws binary to avoid bulking it up. Trouble with resources.go code is that it's evaluated twice, once for tfgen and once for runtime startup in the provider, so it's a little tricky, but can be done with os.ReadFile that only runs for `make tfgen` case.

provider/pkg/overlays/examples/callbackFunction.md
```diff
@@ -0,0 +1,216 @@
...
+- `AWSXrayWriteOnlyAccess`
+
+### Lambda Function Handler
+
+You can provide the JavaScript function used for the Lambda Function's Handler either directly by setting the `callback` input property or instead specify the `callbackFactory`, which is a Javascript function that will be called to produce the callback function that is the entrypoint for the AWS Lambda.
+Using `callbackFactory` is useful when there is expensive initialization work that should only be executed once. The factory-function will be invoked once when the final AWS Lambda module is loaded. It can run whatever code it needs, and will end by returning the actual function that Lambda will call into each time the Lambda is invoked.
+
+The callback function should follow the following signature: `(event: E, context: Context, callback: (error?: Error | string | null, result?: R) => void) => Promise<R> | void`. It is recommended to use an async function, otherwise the Lambda execution will run until the `callback` parameter is called and the event loop is empty. See [Define Lambda function handler in Node.js](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html) for additional details.
```
t0yv0, line 24: Consider avoiding inline code gory details in favor of a code example (or just entirely drop).

provider/pkg/overlays/examples/callbackFunction.md
```diff
@@ -0,0 +1,216 @@
+A CallbackFunction is a special type of `aws.lambda.Function` that can be created out of an actual JavaScript function instance. The Pulumi compiler and runtime work in tandem to extract your function, package it up along with its dependencies, upload the package to AWS Lambda, and configure the resulting AWS Lambda resources automatically.
+
+The JavaScript function may capture references to other variables in the surrounding code, including other resources and even imported modules. The Pulumi compiler figures out how to serialize the resulting closure as it uploads and configures the AWS Lambda. This works even if you are composing multiple functions together.
+
+See [Function Serialization](https://www.pulumi.com/docs/concepts/inputs-outputs/function-serialization/) for additional details on this process.
+
+If no IAM Role is specified, CallbackFunction will automatically use the following managed policies:
```
t0yv0, line 7: Is this important information to have upfront, or perhaps reference section at the bottom?

t0yv0 (approved):
Very nice.

provider/overlays.go
flostadler: IMO we should opt for the simpler option here to avoid bugs and trouble down the road. Even if we had multiple files it wouldn't be a lot of storage size we're adding. The markdown file for the callbackFunction has ~9KB, so we'd need a lot of them before this has meaningful impact on the overall size of the provider (~794MB right now).

t0yv0: You are right, I concede the point.. The provider binary is too large but we need to look holistically as to what to do with that, this is not really the root problem here.

t0yv0 (approved):

mikhailshilkov:
This PR has been shipped in [release v6.49.0](https://github.com/pulumi/pulumi-aws/releases/tag/v6.49.0).

PR #4356, opened 2024-08-09
[TESTING]: bridge generate shorter type names

corymhall:
PR to test out the changes in https://github.com/pulumi/pulumi-terraform-bridge/pull/2290

Only checking in the java sdk to keep the file changes to a minimum

NOTE: the types that are removed have been merged with one of the renamed types since the underlying type is the same


sdk/java/src/main/java/com/pulumi/aws/autoscaling/inputs/PolicyPredictiveScalingConfigurationMetricArgs.java
```diff
@@ -63,9 +63,9 @@ public Output<String> namespace() {
         return this.namespace;
     }
 
-    private PolicyPredictiveScalingConfigurationMetricSpecificationCustomizedScalingMetricSpecificationMetricDataQueryMetricStatMetricArgs() {}
+    private PolicyPredictiveScalingConfigurationMetricArgs() {}
 
-    private PolicyPredictiveScalingConfigurationMetricSpecificationCustomizedScalingMetricSpecificationMetricDataQueryMetricStatMetricArgs(PolicyPredictiveScalingConfigurationMetricSpecificationCustomizedScalingMetricSpecificationMetricDataQueryMetricStatMetricArgs $) {
+    private PolicyPredictiveScalingConfigurationMetricArgs(PolicyPredictiveScalingConfigurationMetricArgs $) {
```
t0yv0, line 68: So nice..

t0yv0:
Actually other languages could be curious to see.

PR #4384, opened 2024-08-19
Start running upstream tests for waf and sqs

t0yv0:
Adds a CI job to run upstream tests on every PR. To start, the following services are tested:

- sqs
- waf

Tags related tests are excluded as they fail.

Related: pulumi/pulumi-aws#3022


t0yv0:
Early results: it appears feasible to run the full matrix of upstream acc tests. There are some successes and a lot of failures. The common failures:

1. tags handling

```
== NAME  TestAccRedshiftCluster_tags
    cluster_test.go:476: Step 3/4 error: Check failed: Check 2/4 error: aws_redshift_cluster.test: Attribute 'tags.%' expected "2", got "1"
=== NAME  TestAccRedshiftClusterSnapshot_tags
```

To fix this we would need to either revisit tags-editing patches to be conditional on an env var so we can run acctests modulo the tags patches, or supply a regex that rejects _tags tests.

2. VPC limit reached - we might need some cleanup in the devsandbox account

3.  service-specific issues like the following, we can scope out these services from the matrix to start out: 

```
=== RUN   TestAccMacie2_serial/Account/disappears
    account_test.go:183: Step 1/1 error: Error running apply: exit status 1

        Error: enabling Macie Account: ConflictException: Macie has already been enabled
```

upstream-tools/package.json
```diff
@@ -8,7 +8,7 @@
     "strip-links": "tf-patch strip-links --cwd ../upstream",
     "global-replace": "tf-patch global-replace --cwd ../upstream",
     "replace": "tf-patch replace --cwd ../upstream",
-    "apply": "yarn pre-replace && yarn strip-links && yarn global-replace"
```
flostadler, line 11: nice! I wonder if we should reformat our patches as well.

Right now when working on a patch we'd most likely get the upstream tests complaining about formatting, right?

.github/workflows/run-upstream-tests.yml
```diff
@@ -0,0 +1,71 @@
...
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-region: ${{ env.AWS_REGION }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          role-duration-seconds: 3600
+          role-session-name: aws@githubActions
+          role-to-assume: ${{ secrets.AWS_CI_ROLE_ARN }}
+      - name: Test ${{ matrix.service }}
+        run: |
+          cd upstream
+          TF_ACC=1 make testacc GO_VER=go PKG=${{ matrix.service }} ACCTEST_PARALLELISM=6 TESTARGS="-skip 'tags|.*/.*/Tags'"
```
flostadler, line 71: Should we document that somewhere so folks know how to run this locally?

flostadler (approved):
Niceeee!

.github/workflows/run-upstream-tests.yml
```diff
@@ -0,0 +1,71 @@
...
+env:
+  PR_COMMIT_SHA: ${{ github.event.client_payload.pull_request.head.sha }}
+  AWS_REGION: us-west-2
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  TF_APPEND_USER_AGENT: pulumi
+
+# This should cancel any previous runs of the same workflow on the same branch which are still running.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
corymhall, line 33: We probably need this, but it will also lead to more orphaned resources if tests are cancelled before the destroy step.

corymhall (approved):
Nice!

Since most of these upstream tests should have sweepers, maybe we should add a backlog item for running the sweepers on a schedule for the services we're testing.

t0yv0:
I've added https://github.com/pulumi/home/issues/3603 to track the options we have for running sweepers.

upstream-tools/package.json
t0yv0: I've started going that route, but with this addition it's no longer necessary to do that, just busy-work, so I think we do not need to reformat the patches.

.github/workflows/run-upstream-tests.yml
t0yv0: Added a note 🙏

.github/workflows/run-upstream-tests.yml
t0yv0: Added a link, upstream has good docs on this.

upstream-tools/package.json
flostadler: But if I'm currently rebasing the patches or working on one, this tool wouldn't be executed. Right? Afaik it's only executed on `make upstream`

In this case we'd most likely have formatting issues?:
```
./upstream.sh check_out
./upstream.sh rebase -i
```

PR #4425, opened 2024-09-03
Upstream 5.65.0

t0yv0:
Upgrade upstream to v5.65.0

Fixes https://github.com/pulumi/pulumi-aws/issues/4422

New Data Source: aws_shield_protection
New Resource: aws_glue_catalog_table_optimizer

Upgrade to Go 1.23.0 following upstream.


patches/0037-Restore-legacy-bucket.patch
```diff
@@ -5,54 +5,1179 @@ Subject: [PATCH] Restore legacy bucket
...
-index 87a0092938..b4265d86f7 100644
+index 0d8ccda18d..ee2ff5cbfc 100644
 --- a/go.mod
 +++ b/go.mod
-@@ -256,6 +256,7 @@ require (
+@@ -1,6 +1,6 @@
+ module github.com/hashicorp/terraform-provider-aws
+ 
+-go 1.23.0
++go 1.22.6
```
flostadler, line 15: This change seems a bit odd to me.
Maybe I'm reading it wrong, but aren't we downgrading the go version and a lot of the other dependencies here? Why is that needed for the Bucket patch?

t0yv0: It might be an artifact of my local tooling versions that's not needed. I'll have a look.

.github/workflows/aws-upstream-tests.yml
```diff
@@ -1,75 +0,0 @@
-name: run-upstream-tests
```
flostadler, line 1: I guess we'll also need to update the make target [here](https://github.com/pulumi/pulumi-aws/blob/3ff4fa4ddc54bcef6d043e7c62e0e881c6ecdaa5/Makefile#L192). It's deleting all workflows

flostadler (approved):
LGTM
Only the `aws-upstream-tests` make target needs some fixing

.github/workflows/aws-upstream-tests.yml
t0yv0: I'll update. Good callout.

Issue #4424, opened 2024-09-03
Update aws lambda runtime constants


LazaroOnline:
Update the aws.lambda runtime constants to add the new versions available.
In this repo's file:
https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/lambda/runtimes.ts

In particular related to NodeJs I can only see up to "`nodejs12.x`", and at the moment of this writing there is up until "`nodejs20.x`", so it is missing:
"`nodejs14.x`"
"`nodejs16.x`"
"`nodejs18.x`"
"`nodejs20.x`"

For the other  languages I imagine it is also outdated.

The "`@deprecated`"  comment should also be added to the relevant versions.

LazaroOnline:
I was using those constants from this code from `import * as aws from "@pulumi/aws";`
`aws.lambda.NodeJS12dXRuntime`

But I just discovered there is another file that is up to date here: 
`aws.lambda.Runtime.NodeJS20dX`

Now the question is what is the difference and if the old one can be removed.

t0yv0:
Glad you found the updated list. The SDK retains the older enum values for backwards-compat, as the doc comment says:

```/** The NodeJS12dXRuntime constant is deprecated, use Runtime.NodeJS12dX instead. */
export let NodeJS12dXRuntime:    Runtime = "nodejs12.x";
```

t0yv0:
Looks like these were marked with `@deprecated` but this was removed in https://github.com/pulumi/pulumi-aws/pull/1194

I will follow up here with the team, thanks for calling this to our attention.

PR #4426, opened 2024-09-03
Annotate Node runtime constants with @deprecated

t0yv0:
Fixes https://github.com/pulumi/pulumi-aws/issues/4424

Literals like lambda.Java8Runtime are marked formally @deprecated with this change. While they do work, the preferred form is lambda.Runtime.Java8. The preferred form also has the up-to-date literals.


t0yv0:
Having followed up on this, we'll be able to do this at the next major release but not earlier. Tracking https://github.com/pulumi/pulumi-aws/issues/4648

PR #4449, opened 2024-09-11
T0yv0/full quicksight schema

t0yv0:

PR #4458, opened 2024-09-13
Upstream v5.67.0

t0yv0:
This change upgrades upstream to v5.67.0.

Fixes https://github.com/pulumi/pulumi-aws/issues/4435
Fixes https://github.com/pulumi/pulumi-aws/issues/4454

**v5.66.0 changes:**

Notable upstream additions:

New resources:

- datazone/environment.Environment

New functions:

- glue/getRegistry.getRegistry
- organizations/....getOrganizationalUnitDescendantOrganizationalUnits
- quicksight/getAnalysis.getAnalysis
- quicksight/getQuicksightAnalysis.getQuicksightAnalysis

Unfortunately the aws_quicksight_analysis datasource cannot yet be fully supported in the provider because excessively large type hierarchies in Quicksight cause issues for Pulumi SDKs. This is the same problem that makes "definition" fields unavailable for Quicksight resources.  A solution is being explored in https://github.com/pulumi/pulumi-aws/pull/4449 but in the meantime the "definition" field is not available yet in Pulumi for the given datasource. 

There also was a regression https://github.com/hashicorp/terraform-provider-aws/issues/39320 in v5.66.0 that tripped pulumi-aws testing so the matching pulumi-aws release is skipped.

**v5.67.0 changes:**

- cloudformation/stackInstances.StackInstances
- codebuild/fleet.Fleet
- computeoptimizer/enrollmentStatus.EnrollmentStatus
- computeoptimizer/recommendationPreferences.RecommendationPreferences
- costoptimizationhub/enrollmentStatus.EnrollmentStatus
- costoptimizationhub/preferences.Preferences
- datazone/assetType.AssetType
- lambda/functionRecursionConfig.FunctionRecursionConfig

New functions:

- codebuild/getFleet.getFleet

Unfortunately one of the upstream features that introduced support for IAM role chaining is projecting as a breaking change to Pulumi types. Since our policy is to prioritize backwards compatibility, this feature has been backed out and is not yet available in the upcoming pulumi-aws release. 

See: https://github.com/hashicorp/terraform-provider-aws/issues/22728

Reminder opened to track this in https://github.com/pulumi/pulumi-aws/issues/4459


t0yv0:
Looks like assumeRole is now assumeRoles (plural), which is a breaking change for Pulumi. Looking at options quickly. 

```
SError: ⨯ Unable to compile TypeScript:
Error: index.ts(28,5): error TS2345: Argument of type '{ assumeRole: { roleArn: pulumi.Output<string>; }; }' is not assignable to parameter of type 'ProviderArgs'.
  Object literal may only specify known properties, but 'assumeRole' does not exist in type 'ProviderArgs'. Did you mean to write 'assumeRoles'?
```

t0yv0:
Coming from here: https://github.com/hashicorp/terraform-provider-aws/commit/faa9b8e72c43a08d77c6d1e123b1ad7063dcc69a

patches/0037-Restore-legacy-bucket.patch
```diff
@@ -5,319 +5,508 @@ Subject: [PATCH] Restore legacy bucket
 
 
 diff --git a/go.mod b/go.mod
-index 0d8ccda18d..85b3de0453 100644
+index 5b8c85d7d3..a145fa7f16 100644
```
flostadler, line 8: This patch diff is hard to review. Couldn't spot anything wrong though

flostadler (approved):
This was a wild ride! Good job!

patches/0037-Restore-legacy-bucket.patch
t0yv0: We should probably take upstream/go.mod churn out of patches and try to see if we can accept the unmodified upstream go.mod.

t0yv0: https://github.com/pulumi/pulumi-aws/issues/4465

PR #4467, opened 2024-09-17
WIP: try to fix dangling references in the schema

t0yv0:


provider/overlays.go
```diff
@@ -188,4 +188,11 @@ func init() {
 	resourceOverlays[callbackFunctionTok] = callbackFunction
 
 	typeOverlays[codePathOptionsTok] = codePathOptions
+
+	typeOverlays["aws:index/aRN:ARN"] = schema.ComplexTypeSpec{
```
t0yv0, line 192: This is an attempt to register an overlay-only type with the schema, hoping that Node SDK would be able to recognize references and other SDK would ignore it.

provider/resources.go
```diff
@@ -873,14 +873,18 @@ compatibility shim in favor of the new "name" field.`)
...
+			f.InputProperties["logGroup"].OneOf[1].Ref = "#/resources/aws:cloudwatch/logGroup:LogGroup"
 		},
 
 		Config: map[string]*tfbridge.SchemaInfo{
 			"assume_role": {
 				MaxItemsOne: tfbridge.True(),
 			},
 			"region": {
-				Type: awsTypeDefaultFile(awsMod, "Region"),
+				Type: "aws:index/Region:Region",
```
t0yv0, line 887: This now points to the official enum for region. Highlighing here because region is one of the most impactful potentially breaking changes from the lot.

provider/resources.go
```diff
@@ -1235,7 +1239,7 @@ compatibility shim in favor of the new "name" field.`)
 						AltTypes: []tokens.Type{awsResource(ec2Mod, "PlacementGroup")},
 					},
 					"enabled_metrics": {
-						Elem: &tfbridge.SchemaInfo{Type: awsType(autoscalingMod, "metrics", "Metric")},
+						Elem: &tfbridge.SchemaInfo{Type: awsType(autoscalingMod, "Metric", "Metric")},
```
t0yv0, line 1242: A dangling reference is replaced with an actual reference to a real enum here, as one possibility to fix the problem. The other possibility would be to introduce a IsOverlay entry for the previous dangling reference.

sdk/go/aws/sns/topic.go
```diff
@@ -163,7 +164,7 @@ type Topic struct {
 	// The message archive policy for FIFO topics. More details in the [AWS documentation](https://docs.aws.amazon.com/sns/latest/dg/message-archiving-and-replay-topic-owner.html).
 	ArchivePolicy pulumi.StringPtrOutput `pulumi:"archivePolicy"`
 	// The ARN of the SNS topic, as a more obvious property (clone of id)
-	Arn pulumi.StringOutput `pulumi:"arn"`
+	Arn aws.ARNOutput `pulumi:"arn"`
```
t0yv0, line 167: Breaking

sdk/go/aws/sns/topic.go
```diff
@@ -259,7 +260,7 @@ type topicState struct {
 	// The message archive policy for FIFO topics. More details in the [AWS documentation](https://docs.aws.amazon.com/sns/latest/dg/message-archiving-and-replay-topic-owner.html).
 	ArchivePolicy *string `pulumi:"archivePolicy"`
 	// The ARN of the SNS topic, as a more obvious property (clone of id)
-	Arn *string `pulumi:"arn"`
+	Arn *aws.ARN `pulumi:"arn"`
```
t0yv0, line 263: Breaking

sdk/nodejs/provider.ts
```diff
@@ -68,10 +66,6 @@ export class Provider extends pulumi.ProviderResource {
      * The profile for API operations. If not set, the default profile created with `aws configure` will be used.
      */
     public readonly profile!: pulumi.Output<string | undefined>;
-    /**
-     * The region where AWS operations will take place. Examples are us-east-1, us-west-2, etc.
-     */
-    public readonly region!: pulumi.Output<Region | undefined>;
```
t0yv0, line 74: It's suspect that this got removed. It should not have been removed probably.

sdk/nodejs/provider.ts
```diff
@@ -214,7 +208,7 @@ export interface ProviderArgs {
     /**
      * The region where AWS operations will take place. Examples are us-east-1, us-west-2, etc.
      */
-    region?: pulumi.Input<Region>;
+    region?: pulumi.Input<enums.Region>;
```
t0yv0, line 211: Slightly breaking - the type is compatible but it moved.

PR #4476, opened 2024-09-19
chore: introduce stale issue workflow

corymhall:
There are two workflows that we want to implement

1. Stale issue workflow
- If an issue has the `response-requested-label` for `days-before-stale`
  then add the `stale-issue-label` and comment with the `stale-issue-message`
- If an issue has the `stale-issue-label` for `days-before-close` then close the issue

2. Ancient issue workflow (currently disabled)
- If an issue has been open for `days-before-ancient` then add then `ancient-issue-message`
  and add the `stale-issue-label` (which starts the Stale issue workflow)

The workflow introduced in this PR is the stale issue workflow. A separate ancient issue workflow will be introduced in a separate PR.

### DIFF
--- a/.github/workflows/aws-close-stale-issues.yml
+++ b/.github/workflows/aws-close-stale-issues.yml
@@ -53,4 +53,5 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         loglevel: DEBUG
         # Set dry-run to true to not perform label or close actions.
-        dry-run: false
+        # TODO: switch to false after some testing
+        dry-run: true
--- a/.github/workflows/aws-close-stale-issues.yml
+++ b/.github/workflows/aws-close-stale-issues.yml
@@ -1,0 +1,62 @@
+name: "Close Stale Issues"
+
+on:
+  workflow_dispatch:
+  schedule:
+    # runs every day at midnight
+  - cron: "0 0 * * *"
+
+# This job currently has two workflows
+# 1. Stale issue workflow
+# - If an issue has the `response-requested-label` for `days-before-stale`
+#   then add the `stale-issue-label` and comment with the `stale-issue-message`
+# - If an issue has the `stale-issue-label` for `days-before-close` then close the issue
+# 2. Ancient issue workflow (currently disabled)
+# - If an issue has been open for `days-before-ancient` then add then `ancient-issue-message`
+#   and add the `stale-issue-label` (which starts the Stale issue workflow)
+jobs:
+  cleanup:
+    # this workflow will always fail in forks; bail if this isn't running in the upstream
+    if: github.repository == 'pulumi/pulumi-aws'
+    permissions:
+      issues: write
+      contents: read
+    runs-on: ubuntu-latest
+    name: Stale issue job
+    steps:
+    - uses: aws-actions/stale-issue-cleanup@v6
+      with:
+        # Types of issues that will be processed
+        issue-types: issues
+
+        # Setting messages to an empty string will cause the automation to skip
+        # that category
+        stale-issue-message: This issue has not received a response in a while. If you want to keep this issue open, please leave a comment below and auto-close will be canceled.
+
+        # TODO: turn on ancient-issue workflow once we are ready
+        # ancient-issue-message: This issue has not received any attention in 1 year. If you want to keep this issue open, please leave a comment below and auto-close will be canceled.
+
+        # These labels are required
+        stale-issue-label: closing-soon
+        exempt-issue-labels: no-autoclose
+        response-requested-label: awaiting-feedback
+
+        # Don't set closed-for-staleness label to skip closing very old issues
+        # regardless of label
+        closed-for-staleness-label: closed-for-staleness
+
+        # Issue timing
+        days-before-stale: 5
+        days-before-close: 10
+        days-before-ancient: 365
+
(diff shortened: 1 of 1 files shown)


.github/workflows/aws-close-stale-issues.yml
```diff
@@ -0,0 +1,62 @@
...
+        # ancient-issue-message: This issue has not received any attention in 1 year. If you want to keep this issue open, please leave a comment below and auto-close will be canceled.
+
+        # These labels are required
+        stale-issue-label: closing-soon
+        exempt-issue-labels: no-autoclose
+        response-requested-label: awaiting-feedback
+
+        # Don't set closed-for-staleness label to skip closing very old issues
+        # regardless of label
+        closed-for-staleness-label: closed-for-staleness
```
corymhall, line 46: I think we need to change this to something that the pulumi-bot accepts. Should it be `resolution/no-repro`? Or should we introduce a new one `resolution/stale`?

t0yv0: I like a new resolution since we do not have too many of those yet.

.github/workflows/aws-close-stale-issues.yml
```diff
@@ -0,0 +1,62 @@
...
+        # These labels are required
+        stale-issue-label: closing-soon
+        exempt-issue-labels: no-autoclose
+        response-requested-label: awaiting-feedback
+
+        # Don't set closed-for-staleness label to skip closing very old issues
+        # regardless of label
+        closed-for-staleness-label: resolution/stale
+
+        # Issue timing
```
t0yv0, line 48: Should this be 90 or 180 days for starters? Per proposal.

.github/workflows/aws-close-stale-issues.yml
```diff
@@ -0,0 +1,62 @@
...
+    name: Stale issue job
+    steps:
+    - uses: aws-actions/stale-issue-cleanup@v6
+      with:
+        # Types of issues that will be processed
+        issue-types: issues
+
+        # Setting messages to an empty string will cause the automation to skip
+        # that category
+        stale-issue-message: This issue has not received a response in a while. If you want to keep this issue open, please leave a comment below and auto-close will be canceled.
```
flostadler, line 34: Should we add some information about the days since the last interaction.
I.e. "This issue has not received a response in the last X days..."

.github/workflows/aws-close-stale-issues.yml
```diff
@@ -0,0 +1,62 @@
...
+
+        # TODO: turn on ancient-issue workflow once we are ready
+        # ancient-issue-message: This issue has not received any attention in 1 year. If you want to keep this issue open, please leave a comment below and auto-close will be canceled.
+
+        # These labels are required
+        stale-issue-label: closing-soon
+        exempt-issue-labels: no-autoclose
```
t0yv0, line 41: Should this include `p1`, `customer-feedback`, `impact/regression`.

Per your proposal it sounds like it should also be `awaiting-upstream` as we wanted to have an automated check there for whether the upstream has been closed or not to gate staling it out?

.github/workflows/aws-close-stale-issues.yml
```diff
@@ -0,0 +1,62 @@
...
+        # If you don't want to mark a issue as being ancient based on a
+        # threshold of "upvotes", you can set this here. An "upvote" is
+        # the total number of +1, heart, hooray, and rocket reactions
+        # on an issue.
+        minimum-upvotes-to-exempt: 5
+
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        loglevel: DEBUG
+        # Set dry-run to true to not perform label or close actions.
+        dry-run: false
```
flostadler, line 62: Might actually be worth it to run it with dry-run for the first couple of days to check whether it's actually doing what we want

corymhall: I'm not sure, the worst that happens is some issues get closed which we can see and re-open. Otherwise we have to remember to look at the logs of each run everyday.

.github/workflows/aws-close-stale-issues.yml
corymhall: Given that this workflow only targets `awaiting-feedback` issues I think we don't need to wait so long.

.github/workflows/aws-close-stale-issues.yml
corymhall: Added!

.github/workflows/aws-close-stale-issues.yml
corymhall: This only affects issues labeled with `awaiting-feedback` which I think should affect all issues regardless of label. We can always add the `no-autoclose` to issues that we require a response on, but want to leave open even if we never get a response (would we ever want that?)

.github/workflows/aws-close-stale-issues.yml
flostadler: It's not something I'm too worried about, but what if it wrongly closes too many issues and we have to spend time manually cleaning it up?
What if we use dry-run for the first run and then switch it off?

corymhall: That's fair

t0yv0 (approved):

corymhall:
Sounds like we don't want to do this

t0yv0:
Pending discussion, will revisit.

PR #4478, opened 2024-09-19
Remove explicit dependency on core SDK

justinvp:
This change removes the explicitly specified dependency on the core SDK from the schema so that the SDK generator can emit the minimum required version that's actually required for the generated provider SDK.

### DIFF
--- a/sdk/dotnet/Pulumi.Aws.csproj
+++ b/sdk/dotnet/Pulumi.Aws.csproj
@@ -45,7 +45,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Pulumi" Version="3.*" />
+    <PackageReference Include="Pulumi" Version="[3.66.1.0,4)" />
   </ItemGroup>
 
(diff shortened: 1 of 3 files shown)


t0yv0:
Sorry we missed this Justin, having a look.

sdk/dotnet/Pulumi.Aws.csproj
```diff
@@ -45,7 +45,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Pulumi" Version="3.*" />
+    <PackageReference Include="Pulumi" Version="[3.66.1.0,4)" />
```
t0yv0, line 48: Beautiful.

t0yv0:
TestRandomTagsCombinationsGo/test-9 looks like a flake.

t0yv0 (approved):
Looks great. This may be a little stale needs another make tfgen round + rebase which we have not automated yet. I'll do that tomorrow.

PR #4480, opened 2024-09-20
Upgrade pulumi-terraform-bridge to v3.91.0

iwahbe:
This PR was generated via `$ upgrade-provider pulumi/pulumi-aws --kind=bridge`.

---

- Updating Java Gen version from 0.15.0 to 0.16.1.
- Upgrading pulumi-terraform-bridge from v3.90.1-0.20240911150014-7864bb06caa7 to v3.91.0.
- Upgrading pulumi-terraform-bridge/pf from v0.43.1-0.20240911150014-7864bb06caa7 to v0.44.0.


iwahbe:
@t0yv0 @flostadler Is it safe to assume that you had set aws to use an upstream bridge version, instead of a unmerged branch of the bridge?

flostadler:
> @t0yv0 @flostadler Is it safe to assume that you had set aws to use an upstream bridge version, instead of a unmerged branch of the bridge?

Yeah, we set it to https://github.com/pulumi/pulumi-terraform-bridge/commit/7864bb06caa76ac4adc8a7f44f8c8762f9a428cc which was released as part of v3.91.0

t0yv0 (approved):
Looks like some internals of invoke handling changed but all good!

Issue #4465, opened 2024-09-16
Reduce upstream/go.mod patch churn


t0yv0:
When dependencies change upstream, it often generates patch conflicts over upstream/go.mod. Need to double-check if any modifications are actually necessary. If they are not necessary we can switch to using unmodified upstream/go.mod. If they are indeed necessary we can look at reducing the conflicts by some automation (e.g. automatically apply go mod tidy).

Issue #4362, opened 2024-08-13
Warnings emitted when importing aws:lb:Listener, One or more imported inputs failed to validate


Zaid-Ajaj:
Similar to #4361 importing `aws:lb:Listener` works but emits warnings:
```
  aws:lb:Listener (listener):
    warning: One or more imported inputs failed to validate. This is almost certainly a bug in the `aws` provider. The import will still proceed, but you will need to edit the generated code after copying it into your program.
    warning: aws:lb/listener:Listener resource 'listener' has a problem: expected default_action.0.order to be in the range (1 - 50000), got 0. Examine values at 'listener.defaultActions[0].order'.

```

flostadler:
Thanks for reporting this @Zaid-Ajaj! I have a hunch that this is nils/default values for primitives returned by the AWS SDK not being handled correctly.

I tried reproducing this, but couldn't trigger this exact error. But another one:
```
aws:lb/listener:Listener resource 'testitest' has a problem: expected default_action.0.forward.0.stickiness.0.duration to be in the range (1 - 604800), got 0. Examine values at 'testitest.defaultActions[0].forward.stickiness.enabled'.
```

This is the Listener I tried importing:
```
const listener = new aws.lb.Listener("my-listener", {
    loadBalancerArn: loadbalancer.arn,
    port: 80,
    defaultActions: [{
        type: "forward",
        targetGroupArn: targetGroup.arn,
    }],
});
```

I'm quite sure this line here is triggering the initial issue. It's converting the pointer to `Order` to an int32, which will yield 0 if `Order` is nil:
https://github.com/hashicorp/terraform-provider-aws/blob/3724def77810f619392ecb79fd3e40915c0b0214/internal/service/elbv2/listener.go#L977

Similarly this here will trigger the issue I was able to find because it coerces the pointers for stickiness `duration` & `enabled` config into primitives:
https://github.com/hashicorp/terraform-provider-aws/blob/3724def77810f619392ecb79fd3e40915c0b0214/internal/service/elbv2/listener.go#L1170-L1171

We'll need to fix the upstream handling of optional values in read for elbv2 Listener.

flostadler:
Correcting the code of the imported resource and running `pulumi up` correctly aligns the state.
This is in line with the behavior in upstream. e.g. see `Step 4 - Import` here https://github.com/hashicorp/terraform-provider-aws/issues/37211#issuecomment-2127911378.

[...]

PR #4571, opened 2024-09-27
Upgrade upstream to v5.69.0

t0yv0:
Upstream is updated to [v5.69.0](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v5.69.0)

There are no new resources or functions but the update is expected to propagate bug fixes. Notably this update also propagates updating to `Go 1.23.1` for the provider builds.

Fixes https://github.com/pulumi/pulumi-aws/issues/4563
Fixes https://github.com/pulumi/pulumi-aws/issues/4465
Fixes https://github.com/pulumi/pulumi-aws/issues/4362

Taking this opportunity to remove modifications from upstream/go.mod so that the Pulumi provider uses a consistent Go tool-chain and dependency set with upstream.


rquitales (approved):
LGTM. Update follows the playbook and there are no schema changes.

provider/cmd/pulumi-resource-aws/schema.json
```diff
@@ -275029,7 +275182,7 @@
...
+            "description": "\n\n## Import\n\nUsing `pulumi import`, import IAM Roles using the `name`. For example:\n\n```sh\n$ pulumi import aws:iam/role:Role developer developer_name\n```\n",
```
t0yv0, line 275185: We missed this, curiously this is not good.

Issue #4573, opened 2024-09-27
AWS IAM and other services don´t work anymore (TLS issue)


timlukastlt:
### Describe what happened

Pulumi is attempting to make a request to ```iam.amazonaws.com``` but uses TLS 1.0. Since AWS only supports higher versions of TLS, the request is blocked, as shown in the image below:
![Image](https://github.com/user-attachments/assets/d419b57f-1044-4730-bb6b-4ab2a894034b)

For sts as an example the requests from Pulumi are working because is uses TLS3. ![Image](https://github.com/user-attachments/assets/ea3f57e0-1d9a-4460-bde0-46597731d7aa)

### Sample program

```python

import pulumi
import pulumi_aws as aws

example = aws.iam.get_role(name="example-iam-role")

```

### Log output

The programm just shows creating... for one hour (55-65 minutes)
After that it shows that the connection to iam.amazonaws.com was not possible.

### Affected Resource(s)

We got this error for iam

### Output of `pulumi about`

```
CLI         
Version     3.134.1
Go Version   go1.23.1
Go Compiler  gc

Plugins
KIND      NAME    VERSION
resource  aws     6.53.0
language  python  unknown

Host    
OS       debian
Version  12.7
Arch     x86_64

This project is written in python: executable='/app/venv/bin/python' version='3.9.20'

Backend       
Name           3fd5c982165b
URL            s3://our-bucket/test123
User           root
Organizations 
Token type     personal

Dependencies:
NAME        VERSION
pip         23.0.1
pulumi_aws  6.53.0
setuptools  58.1.0
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

flostadler:
I'm sorry you're running into this @timlukastlt. I'm starting to investigate it and will provide updates as soon as possible!

flostadler:
@timlukastlt I'm not able to reproduce this on my end, but I'm strongly convinced that this was caused by upgrading the provider to use Go 1.23.
This Go update introduced a [minor change to the crypto/tls standard library package](https://tip.golang.org/doc/go1.23#cryptotlspkgcryptotls) that seemed to trip up the AWS firewalls.
The upstream provider was affected by this [same problem](https://github.com/hashicorp/terraform-provider-aws/issues/39311). They fixed it by disabling this experimental key exchange mechanism.
We should do the same on our end as a short term workaround.


[...]

Issue #4582, opened 2024-09-30
Go 1.23: Additional key exchange mechanism X25519Kyber768Draft00 causes AWS Network Firewall to drop packets


flostadler:
The AWS Provider was upgraded to Go 1.23 in [v6.51.0](https://github.com/pulumi/pulumi-aws/releases/tag/v6.51.0), which introduced a [minor change to the crypto/tls standard library package](https://tip.golang.org/doc/go1.23#cryptotlspkgcryptotls):

> The experimental post-quantum key exchange mechanism X25519Kyber768Draft00 is now enabled by default when [Config.CurvePreferences](https://tip.golang.org/pkg/crypto/tls#Config.CurvePreferences) is nil. The default can be reverted by adding tlskyber=0 to the GODEBUG environment variable.

This experimental key exchange mechanism seems to trip up the AWS firewall.

The upstream provider was affected by this [same problem](https://github.com/hashicorp/terraform-provider-aws/issues/39311). They fixed it by disabling this experimental key exchange mechanism.
We should do the same on our end as a short term workaround.

flostadler:
Causes https://github.com/pulumi/pulumi-aws/issues/4573

flostadler:
This is now fixed in https://github.com/pulumi/pulumi-aws/releases/tag/v6.54.1. I'm gonna keep the issue open to keep an eye on a possible long term fix that upstream is working on with AWS.

PR #4583, opened 2024-09-30
Disable experimental post-quantum key exchange mechanism X25519Kyber768Draft00

flostadler:
The AWS Provider was upgraded to Go 1.23 in v6.51.0, which introduced a change
to the crypto/tls standard library package. It enabled the post-quantum
key exchange mechanism `X25519Kyber768Draft00` by default. This experimental key
exchange mechanism is causing errors in the AWS firewall.
As a short term workaround this change disables the experimental key exchange mechanism.

Upstream maintainers and AWS are in touch to work on a long-term fix.

Fixes https://github.com/pulumi/pulumi-aws/issues/4573
Relates to https://github.com/pulumi/pulumi-aws/issues/4582

### DIFF
--- a/provider/go.mod
+++ b/provider/go.mod
@@ -2,6 +2,11 @@ module github.com/pulumi/pulumi-aws/provider/v6
 
 go 1.23.1
 
+// Disable experimental post-quantum key exchange mechanism X25519Kyber768Draft00
+// This was causing errors with AWS Network Firewall
+// https://github.com/pulumi/pulumi-aws/issues/4582
+godebug tlskyber=0
+
 require (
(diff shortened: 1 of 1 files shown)


iwahbe (approved):
Is it possible to get a test in here?

flostadler:
> Is it possible to get a test in here?

I was not able to trigger it myself. Our existing test didn't catch it either.

This is a tricky issue because it seems to be triggered by the length of the [TLS ClientHello message](https://datatracker.ietf.org/doc/html/rfc8446#section-4.1.2). This new key-exchange mechanism causes the length to increase, in certain circumstances this seems to trigger an error in the AWS firewalls.

Issue #4568, opened 2024-09-27
efs.get_file_system_output returns MaxItems=1 panic error


itsx:
Hello,

for a specific EFS configuration, calling  `efs.get_file_system_output` returns a "MaxItems=1" panic error. 

**Setup:** 

Latest Pulumi SDK for Python
- pulumi                    **3.134.1**
- pulumi_aws            **6.53.0**

**Error:**
```
panic: fatal: A failure has occurred: Unexpected multiple elements in array with MaxItems=1
    goroutine 113 [running]:
    github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.failfast(...)
        /home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.126.0/go/common/util/contract/failfast.go:23
    github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.Failf({0x156119b0?, 0xc001a00640?}, {0x0?, 0xc0069d5110?, 0x1527ade0?})
        /home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.126.0/go/common/util/contract/fail.go:32 +0xcb
    github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge.MakeTerraformOutput.func1({0x1774d5b0, 0xc001c73380}, {0x1286b8a0, 0xc007f39410}, {0x1774d4f8, 0xc001a00640}, 0xc00407e9a0, 0x0, 0x1)
        /home/runner/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.88.0/pkg/tfbridge/schema.go:1175 +0xe74
    github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge.MakeTerraformOutput({0x176985d8?, 0xc0069d5110?}, {0x1774d5b0?, 0xc001c73380?}, {0x1286b8a0?, 0xc007f39410?}, {0x1774d4f8, 0xc001a00640}, 0x0?, 0x0, ...)
        /home/runner/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.88.0/pkg/tfbridge/schema.go:1214 +0xaa
[...]

Issue #4588, opened 2024-10-01
Lost descriptions


t0yv0:
### Describe what happened

It appears we lost resource descriptions from schema and registry

https://www.pulumi.com/registry/packages/aws/api-docs/iam/role/

I have noticed that running `make tfgen` adds descriptions to the checked-in schema.json.

This seems to be a very short description:

```
jq -r ".resources[\"aws:iam/role:Role\"].description" $(git ls-files "**schema.json")                                                                       

## Import

Using `pulumi import`, import IAM Roles using the `name`. For example:

```sh
$ pulumi import aws:iam/role:Role developer developer_name
```

There is a lot more information that used to be available.

### Sample program

See above.

### Log output

N/A

### Affected Resource(s)

Presumably a lot of resources.

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
On 6.53.0 it had:

```
anton@anton-mbp-m3> jq -r ".resources[\"aws:iam/role:Role\"].description" $(git ls-files "**schema.json")   | head                                                                    ~/code/pulumi-aws2
Provides an IAM role.

> **NOTE:** If policies are attached to the role via the `aws.iam.PolicyAttachment` resource and you are modifying the role `name` or `path`, the `force_detach_policies` argument must be set to `true` and applied before attempting the operation otherwise you will encounter a `DeleteConflict` error. The `aws.iam.RolePolicyAttachment` resource (recommended) does not have this requirement.

> **NOTE:** If you use this resource's `managed_policy_arns` argument or `inline_policy` configuration blocks, this resource will take over exclusive management of the role's respective policy types (e.g., both policy types if both arguments are used). These arguments are incompatible with other ways of managing a role's policies, such as `aws.iam.PolicyAttachment`, `aws.iam.RolePolicyAttachment`, and `aws.iam.RolePolicy`. If you attempt to manage a role's policies by multiple means, you will get resource cycling and/or errors.

> **NOTE:** We suggest using explicit JSON encoding or `aws.iam.getPolicyDocument` when assigning a value to `policy`. They seamlessly translate configuration to JSON, enabling you to maintain consistency within your configuration without the need for context switches. Also, you can sidestep potential complications arising from formatting discrepancies, whitespace inconsistencies, and other nuances inherent to JSON.

[...]

PR #4590, opened 2024-10-01
Add pluralized lifecycle_policies to EFS file system data source

flostadler:
The `lifecycle_policy` output of the `efs.getFileSystem` data source has `MaxItemsOne` hardcoded on the Pulumi side and this triggers panics when there is more than one lifecycle policy on the resource.

By adding a pluralized lifecycle_policies output and fixing the singular version to only return at most one element we're able to fix this panic without introducing breaking changes. To incentivize users to move to the pluralized version I've marked the singular version as deprecated.

During the next major release we can drop the patch (I'll create a tracking ticket once this PR goes in) and the auto-aliasing should correctly drop `MaxItemsOne` as well. At that point we're tracking upstream behavior again.

Fixes https://github.com/pulumi/pulumi-aws/issues/4568


provider/cmd/pulumi-resource-aws/schema.json
```diff
@@ -275175,7 +275175,7 @@
...
...!= nil {\n\t\t\treturn err\n\t\t}\n\t\treturn nil\n\t})\n}\n```\n```java\npackage generated_program;\n\nimport com.pulumi.Context;\nimport com.pulumi.Pulumi;\nimport com.pulumi.core.Output;\nimport com.pulumi.aws.iam.Role;\nimport com.pulumi.aws.iam.RoleArgs;\nimport java.util.List;\nimport java.util.ArrayList;\nimport java.util.Map;\nimport java.io.File;\nimport java.nio.file.Files;\nimport java.nio.file.Paths;\n\npublic class App {\n    public static void main(String[] args) {\n        Pulumi.run(App::stack);\n    }\n\n    public static void stack(Context ctx) {\n        var example = new Role(\"example\", RoleArgs.builder()\n            .name(\"yak_role\")\n            .assumeRolePolicy(instanceAssumeRolePolicy.json())\n            .managedPolicyArns()\n            .build());\n\n    }\n}\n```\n```yaml\nresources:\n  example:\n    type: aws:iam:Role\n    properties:\n      name: yak_role\n      assumeRolePolicy: ${instanceAssumeRolePolicy.json}\n      managedPolicyArns: []\n```\n\u003c!--End PulumiCodeChooser --\u003e\n\n## Import\n\nUsing `pulumi import`, import IAM Roles using the `name`. For example:\n\n```sh\n$ pulumi import aws:iam/role:Role developer developer_name\n```\n",
```
flostadler, line 275178: This change relates to: https://github.com/pulumi/pulumi-aws/issues/4588

t0yv0:
Beautiful. I think if you add a patch ticket to track this with a ref to this PR and mark it as impact/breaking we'll be sure to revisit for v7 release.

t0yv0 (approved):

PR #4637, opened 2024-10-11
fixed syntax error in doc comments of topicSubscription.ts

hunterjsb:
## Changes
Fixed two typos that appear in the[ pulumi docs](https://www.pulumi.com/registry/packages/aws/api-docs/sns/topicsubscription/) for `TopicSubscription`. 

## Notes
The two variables, `sns-topic-policy` & `sns-queue-policy`, are kebab case which is not valid in typescipt. After the assignment, however, [subsequent references](https://github.com/pulumi/pulumi-aws/pull/4637/files#diff-81a590bb85aa5fad4c30c8c5c1c7f39b989b219455b53f459c444ee3dd480bdbL140) to there variables were already snake case, so I changed them to snake case.

### DIFF
--- a/sdk/nodejs/sns/topicSubscription.ts
+++ b/sdk/nodejs/sns/topicSubscription.ts
@@ -69,7 +69,7 @@ import {Topic} from "./index";
  *     region: "us-east-1",
  *     "role-name": "service/service",
  * };
- * const sns-topic-policy = aws.iam.getPolicyDocument({
+ * const sns_topic_policy = aws.iam.getPolicyDocument({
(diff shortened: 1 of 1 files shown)


t0yv0:
Thank you for spotting this problem! I have opened https://github.com/pulumi/pulumi-aws/issues/4639 to track. 

Unfortunately these examples are generated automatically and therefore they cannot be edited directly (see the WARNING in the beginning of the file). We will need to find and fix the bug in the intermediate pipeline leading to this result.

PR #4638, opened 2024-10-14
Upstream v5.71.0

t0yv0:
Upstream is updated to v5.71.0

New resources:

backup/restoreTestingPlan.RestoreTestingPlan
backup/restoreTestingSelection.RestoreTestingSelection
datazone/userProfile.UserProfile
pinpoint/smsvoicev2ConfigurationSet.Smsvoicev2ConfigurationSet
route53/profilesAssociation.ProfilesAssociation
route53/profilesProfile.ProfilesProfile
route53/profilesResourceAssociation.ProfilesResourceAssociation

New functions:

amp/getDefaultScraperConfiguration.getDefaultScraperConfiguration
bedrock/getInferenceProfile.getInferenceProfile
bedrock/getInferenceProfiles.getInferenceProfiles
elasticache/getServerlessCache.getServerlessCache
route53/getProfilesProfiles.getProfilesProfiles

Fixes https://github.com/pulumi/pulumi-aws/issues/4630


patches/0034-Fix-tags_all-Computed-for-PF-resources.patch
```diff
@@ -18,18 +18,18 @@ index b6d8c71ef7..f78c4502b0 100644
...
  			},
- 			names.AttrTags:    tftags.TagsAttribute(),
--			names.AttrTagsAll: tftags.TagsAttributeComputedOnly(),
-+			names.AttrTagsAll: tftags.TagsAttribute(),
- 			"training_metrics": schema.ListAttribute{
- 				CustomType: fwtypes.NewListNestedObjectTypeOf[customModelTrainingMetricsModel](ctx),
- 				Computed:   true,
+ 			names.AttrTags:       tftags.TagsAttribute(),
+-			names.AttrTagsAll:    tftags.TagsAttributeComputedOnly(),
++			names.AttrTagsAll:    tftags.TagsAttribute(),
```
t0yv0, line 29: Patch conflict here. I think simply the surrounding code changed upstream, so all this does is makes sure the patch still cleanly applies, while the real edits are removing ComputedOnly.

provider/resources.go
```diff
@@ -5862,4 +5862,10 @@ func setupComputedIDs(prov *tfbridge.ProviderInfo) {
 	prov.Resources["aws_iam_user_policies_exclusive"].ComputeID = func(ctx context.Context, state resource.PropertyMap) (resource.ID, error) {
 		return attr(state, "userName"), nil
 	}
+	prov.Resources["aws_backup_restore_testing_plan"].ComputeID = func(ctx context.Context, state resource.PropertyMap) (resource.ID, error) {
+		return attr(state, "name"), nil
+	}
+	prov.Resources["aws_backup_restore_testing_selection"].ComputeID = func(ctx context.Context, state resource.PropertyMap) (resource.ID, error) {
+		return attr(state, "restoreTestingPlanName", "name"), nil
```
flostadler, line 5869: Isn't this the other way around? i.e. `name:restore_testing_plan_name`

flostadler (commented):
Left one question, looks good otherwise

provider/resources.go
flostadler: I.e. that's how terraform defines the ID: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/backup_restore_testing_selection#import

t0yv0:
@flostadler we didn't quite always align the Pulumi ID with TF import IDs but it is safe to do so and makes sense, I implemented your suggestion, PTAL.

provider/resources.go
```diff
@@ -5865,7 +5868,11 @@ func setupComputedIDs(prov *tfbridge.ProviderInfo) {
...
 		return attr(state, "name"), nil
 	}
-	prov.Resources["aws_backup_restore_testing_selection"].ComputeID = func(ctx context.Context, state resource.PropertyMap) (resource.ID, error) {
-		return attr(state, "restoreTestingPlanName", "name"), nil
+	prov.Resources["aws_backup_restore_testing_selection"].ComputeID = func(
+		ctx context.Context, state resource.PropertyMap,
+	) (resource.ID, error) {
+		// The ${name}:${restoreTestingPlanName} is the format used by TF import, so we are going to use it for
+		// the Pulumi ID as well.
+		return attrWithSeparator(state, ":", "name", "restoreTestingPlanName"), nil
```
flostadler, line 5876: I wasn't even that worried about the separator, mostly the order. But it's fine nevertheless!

flostadler (approved):
LGTM

provider/resources.go
t0yv0: Might as well, I think.

PR #4640, opened 2024-10-14
Fix minimal schema computation

t0yv0:
The machinery for computing the AWS minimal schema is not working quite right. Looking at correcting the issues. 

- [x] fixed CI workflow to check that recomputing the minimal schema is a no-op
- [x] fixed gzip encoding issues with recomputing the minimal schema in go generate
- [x] recomputed the minimal schema to match the latest

### DIFF
--- a/.github/workflows/aws-minimal-schema.yml
+++ b/.github/workflows/aws-minimal-schema.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   check_minimal_schema:
-    runs-on: unbuntu-latest
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v4
--- a/.github/workflows/aws-minimal-schema.yml
+++ b/.github/workflows/aws-minimal-schema.yml
@@ -7,6 +7,17 @@ on:
   pull_request:
     paths-ignore:
     - CHANGELOG.md
+  workflow_dispatch: {}
+  push:
+    branches:
+    # check integration branch
+    - master
+    paths-ignore:
+    - "**.md"
+    # check releases and pre-releases
+    tags-ignore:
+    - sdk/*
+
 
 jobs:
   check_minimal_schema:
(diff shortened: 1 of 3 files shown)


.github/workflows/aws-minimal-schema.yml
```diff
@@ -10,6 +10,7 @@ on:
 
 jobs:
   check_minimal_schema:
+    runs-on: ubuntu-latest
```
t0yv0, line 13: This passes now. I wonder though, this is not part of required checks I think. SO it it fails it will display in red but we can miss it. Are there any quick fixes? 

@flostadler you also thought this should run on release?

flostadler: We should be able to add it to the required checks, right?

I think we should add it to the master workflow as well to guard against faulty merges (e.g. when conflict resolution goes rogue).

provider/cmd/pulumi-resource-aws/generate.go
```diff
@@ -65,10 +66,18 @@ func compressAndVersionSchemaFile(opts compressAndVersionSchemaFileOptions) erro
...
+		err := file.Close()
+		contract.AssertNoErrorf(err, "Failed closing file")
+	}()
 	var w io.Writer = file
 	if opts.gzip {
-		w = gzip.NewWriter(file)
+		gzipWriter := gzip.NewWriter(file)
+		w = gzipWriter
+		defer func() {
+			err := gzipWriter.Close()
```
flostadler, line 78: So the issue was that the writer wasn't flushed before?

flostadler (approved):
LGTM

provider/cmd/pulumi-resource-aws/generate.go
t0yv0: Yes and it made garbled data.

.github/workflows/aws-minimal-schema.yml
t0yv0: Added to master and releases. I am confused if we manage branch protection through UI or centralized stack in ci-mgmt, but I'll ask. We should be able to add later on to branch protections.

provider/cmd/pulumi-resource-aws/generate.go
flostadler: Got it!

PR #4647, opened 2024-10-16
Upstream v5.72.0

t0yv0:
Upgrades upstream to [v5.72.0](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v5.72.0)

New resources:

- iam/rolePolicyAttachmentsExclusive.RolePolicyAttachmentsExclusive

In this release upstream has introduced deprecations for AWS IAM Role inline_policy and managed_policy_arns arguments. These deprecations are not yet propagated into the Pulumi provider (https://github.com/pulumi/pulumi-aws/issues/4654 tracks the context). 

Fixes https://github.com/pulumi/pulumi-aws/issues/4642


t0yv0:
```
      /**
  +     * Set of exclusive IAM managed policy ARNs to attach to the IAM role. If this attribute is not configured, Pulumi will ignore policy attachments to this resource. When configured, Pulumi will align the role's managed policy attachments with this set by attaching or detaching managed policies. Configuring an empty set (i.e., `managedPolicyArns = []`) will cause Pulumi to remove _all_ managed policy attachments.
  +     *
        * @deprecated The managedPolicyArns argument is deprecated. Use the aws.iam.RolePolicyAttachmentsExclusive resource instead.
        */
```

There's new deprecations rolling out, need to double-check this.

sdk/dotnet/Iam/Role.cs
```diff
@@ -204,6 +204,8 @@ namespace Pulumi.Aws.Iam
     /// 
     /// ### Example of Exclusive Managed Policies
     /// 
+    /// &gt; The `managed_policy_arns` argument is deprecated. Use the `aws.iam.RolePolicyAttachmentsExclusive` resource instead.
```
t0yv0, line 207: Hmm.

flostadler (approved):
LGTM

sdk/dotnet/Iam/Role.cs
flostadler: I feel like this is actually a good thing. The inline `managedPolicyArns` are a major foot gun.

I feel like by not surfacing that deprecation we might create more pain when the next major comes around and they possibly remove it (not sure that's upstream's plan though)

t0yv0: Discussing internally.

patches/0070-Temporarily-undo-aws.iam.Role-managed_policy_arns-an.patch
```diff
@@ -0,0 +1,32 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Anton Tayanovskyy <anton@pulumi.com>
+Date: Fri, 25 Oct 2024 13:17:15 -0400
+Subject: [PATCH] Temporarily undo aws.iam.Role managed_policy_arns and
```
flostadler, line 4: Do we already have a ticket for tracking how we wanna tackle this long term?

provider/replacements.json
```diff
@@ -1035,6 +1035,10 @@
...
+    },
+    {
+      "old": "* `managed_policy_arns` - (Optional, **Deprecated**) Set of exclusive IAM managed policy ARNs to attach to the IAM role. If this attribute is not configured, Terraform will ignore policy attachments to this resource. When configured, Terraform will align the role's managed policy attachments with this set by attaching or detaching managed policies. Configuring an empty set (i.e., `managed_policy_arns = []`) will cause Terraform to remove _all_ managed policy attachments.\n",
+      "new": "* `managed_policy_arns` - (Optional, **Deprecated**) Set of exclusive IAM managed policy ARNs to attach to the IAM role. If this attribute is not configured, Pulumi will ignore policy attachments to this resource. When configured, Pulumi will align the role's managed policy attachments with this set by attaching or detaching managed policies. Configuring an empty set (i.e., `managed_policy_arns = []`) will cause Pulumi to remove _all_ managed policy attachments.\n"
```
flostadler, line 1041: Shouldn't we remove the deprecation here as well?

patches/0070-Temporarily-undo-aws.iam.Role-managed_policy_arns-an.patch
t0yv0: Yes. https://github.com/pulumi/pulumi-aws/issues/4654 tracks figuring this out as soon as time permits.

provider/replacements.json
t0yv0: I do not mind this one as much since it is not showing in the builds and the lift on scrubbing all the documentation is higher.

patches/0070-Temporarily-undo-aws.iam.Role-managed_policy_arns-an.patch
```diff
@@ -0,0 +1,32 @@
...
+ inline_policy deprecations
```
flostadler, line 5: Is it intentional that we're removing the `inline_policy` deprecation as well? This was introduced a couple of releases ago: https://github.com/pulumi/pulumi-aws/pull/4571

provider/replacements.json
flostadler: But it will show in the docs, right? I feel like that will be a bit confusing to users (kind of marked as deprecated in the docs but not on the type side)

I don't think it's a big lift to scrub this, it's just two descriptions (managed policies and inline policies). Am I missing something?

t0yv0: Do you think this is the only place where it shows in the docs? Let me have a look.

patches/0070-Temporarily-undo-aws.iam.Role-managed_policy_arns-an.patch
t0yv0: You are right. It's currently published. My mistake, my objective here is to minimize changes in this PR before we have a chance to have an informed conversation. I'll let inline policy stay deprecated, one sec.

flostadler: Awesome!

t0yv0:
My hope is that it should go in after https://github.com/pulumi/pulumi-aws/pull/4673 as the failing test started failing independently. This should not release though as subsequent upstream release fixes some immediate bugs.

provider/replacements.json
corymhall: I'm kind of in favor of keeping it in the docs so at least the users know of the possibility. But I also understand if the decision is to completely hide this from the user, in which case we should remove from the docs as well.

PR #4650, opened 2024-10-16
Discard log messages when initializing the provider

### LINKED ISSUE #4645: Initializing Terraform Provider diagnostic printed for every program
### Describe what happened

An init message is printed into stdout when using pulumi-aws:

```
Previewing update (dev)

View in Browser (Ctrl+O): https://app.pulumi.com/anton-pulumi-corp/bucket-sets/dev/previews/edd87a15-9430-4bac-976f-7da66ca493f5

     Type                 Name             Plan     Info
     pulumi:pulumi:Stack  bucket-sets-dev           1 message

Diagnostics:
  pulumi:pulumi:Stack (bucket-sets-dev):
    2024/10/16 11:16:28 Initializing Terraform AWS Provider...

Resources:
    2 unchanged

Updating (dev)

View in Browser (Ctrl+O): https://app.pulumi.com/anton-pulumi-corp/bucket-sets/dev/updates/4

     Type                 Name             Status     Info
     pulumi:pulumi:Stack  bucket-sets-dev             1 message

Diagnostics:
  pulumi:pulumi:Stack (bucket-sets-dev):
    2024/10/16 11:16:32 Initializing Terraform AWS Provider...

```

### Sample program

```typescript
import * as awsx from "@pulumi/awsx";

// Create an AWS resource (S3 Bucket)
const role = new aws.iam.Role("test-role", {
    assumeRolePolicy: aws.iam.assumeRolePolicyForPrincipal({ Service: "lambda.amazonaws.com" }),
    managedPolicyArns: ["arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"]
})

```

### Log output

N/A

### Affected Resource(s)

_No response_

### Output of `pulumi about`

N/A

### Additional context

```
CLI          
Version      3.130.0
Go Version   go1.22.6
Go Compiler  gc

Plugins
KIND      NAME    VERSION
resource  aws     6.56.0
resource  awsx    2.16.1
resource  docker  4.5.6
resource  docker  3.6.1
language  nodejs  unknown

Host     
OS       darwin
Version  14.6.1
Arch     arm64

This project is written in nodejs: executable='/Users/anton/bin/node' version='v18.18.2'

Current Stack: anton-pulumi-corp/bucket-sets/dev

TYPE                  URN
pulumi:pulumi:Stack   urn:pulumi:dev::bucket-sets::pulumi:pulumi:Stack::bucket-sets-dev
pulumi:providers:aws  urn:pulumi:dev::bucket-sets::pulumi:providers:aws::default_6_56_0
aws:iam/role:Role     urn:pulumi:dev::bucket-sets::aws:iam/role:Role::test-role


Backend        
Name           pulumi.com
URL            https://app.pulumi.com/anton-pulumi-corp
User           anton-pulumi-corp
Organizations  anton-pulumi-corp, moolumi, demo, pulumi
Token type     personal

Dependencies:
NAME            VERSION
@pulumi/pulumi  3.136.1
@types/node     18.19.55
typescript      5.6.3
@pulumi/aws     6.56.0
@pulumi/awsx    2.16.1

Pulumi locates its logs in /var/folders/gd/3ncjb1lj5ljgk8xl5ssn_gvc0000gn/T/com.apple.shortcuts.mac-helper// by default

```

### Contributing

t0yv0:
Messages sent early to the "log" package end up being forwarded to stderr, and Pulumi displays stderr by default to users, which is undesirable.

Later in the provider life-cycle, Pulumi Terraform Bridge will reset `log.SetOutput` again to capture and appropriately track diagnostic logs, which should still work with this change.

Fixes #4645

A test is added to ensure we do not leak diagnostics again on a simple program.

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"path/filepath"
@@ -5875,4 +5876,8 @@ func setupComputedIDs(prov *tfbridge.ProviderInfo) {
 		// the Pulumi ID as well.
 		return attrWithSeparator(state, ":", "name", "restoreTestingPlanName"), nil
 	}
+}
+
+func init() {
+	log.SetOutput(io.Discard)
 }
--- a/provider/provider_nodejs_test.go
+++ b/provider/provider_nodejs_test.go
@@ -123,6 +123,23 @@ func TestChangingRegion(t *testing.T) {
 	})
 }
 
+func TestNoExtranousLogOutput(t *testing.T) {
+	skipIfShort(t)
+	dir := filepath.Join("test-programs", "bucket-obj")
+	cwd, err := os.Getwd()
+	require.NoError(t, err)
+	providerName := "aws"
+	options := []opttest.Option{
+		opttest.LocalProviderPath(providerName, filepath.Join(cwd, "..", "bin")),
+	}
+	test := pulumitest.NewPulumiTest(t, dir, options...)
+	result := test.Preview(t)
+	assert.NotContainsf(t, result.StdOut, "Diagnostics:",
+		"No diagnostics should be emitted to stdout for simple programs")
+	assert.NotContainsf(t, result.StdErr, "Diagnostics:",
+		"No diagnostics should be emitted to stderr for simple programs")
+}
+
 func TestRegressAttributeMustBeWholeNumber(t *testing.T) {
 	// pulumi/pulumi-terraform-bridge#1940
 	skipIfShort(t)


provider/resources.go
```diff
@@ -5876,3 +5877,7 @@ func setupComputedIDs(prov *tfbridge.ProviderInfo) {
 		return attrWithSeparator(state, ":", "name", "restoreTestingPlanName"), nil
 	}
 }
+
+func init() {
+	log.SetOutput(io.Discard)
```
flostadler, line 5882: Can we test in any way that the bridge actually resets this again in the end? I'm a bit worried that the bridge behavior will be eventually changed and then this here just send everything into the void for aws

t0yv0: Can possibly try testing something like this statement in ACM certificate read:

		log.Printf("[WARN] ACM Certificate %s not found, removing from state", arn)

t0yv0: I couldn't find an existing test for this so constructed one using aws.s3.Bucket and object.

iwahbe: This is very similar to https://github.com/pulumi/pulumi-databricks/issues/474#issuecomment-2158908967. If this works, then I don't think it needs to be in `init` (I think `init` occurs after all dependencies have been loaded, ordering of Go's life before main is largely UB). If you put this at the top of the `Provider` function, does it not work? That would give you guaranteed scoping with a `defer log.SetOutput(os.Stdout)`.

provider/upstream_provider.go
```diff
@@ -30,6 +33,9 @@ import (
 // runtime startup is somewhat performance sensitive. Therefore any modifications undertaken here should complete
 // quickly.
 func newUpstreamProvider(ctx context.Context) awsShim.UpstreamProvider {
+	log.SetOutput(io.Discard)      // ignore logging from upstream constructors
```
t0yv0, line 36: I have moved here @iwahbe @flostadler that still works.

This is not entirely the right thing to do. We need to revisit bridge Main APIs and clean them up. The bridge should be initializing the provider and setting up logging against the Pulumi CLI host before calling into the provider-specific constructors. Then instead of discarding these logs would get the regular log sink.

flostadler (approved):
LGTM

provider/upstream_provider.go
iwahbe: We can only get so far with that strategy, but we should do it just to clean up the proliferation of Main APIs.

pulumi-databricks hit log statements during variable setup (before main).

PR #4663, opened 2024-10-24
Upgrade the bridge to v3.93.1

t0yv0:
Fixes https://github.com/pulumi/pulumi-aws/issues/4658

This is a routine update but it also accounts for the fact that /pf modules are deprecated in the bridge and standardizes on the single bridge Go module.


t0yv0:
chmod: changing permissions of '/tmp/pulumi-resource-aws': Operation not permitted
              v3.137.0

I think IMDS test is a little brittle as it assumes a lot of things about the execution environment. It might have gotten broken irrespective of these changes. Looking this up.

t0yv0:
It is passing in master.yml and cron.yml though. Investigating.

t0yv0:
My hope is that https://github.com/pulumi/pulumi-aws/pull/4673 will unblock this.

flostadler:
@t0yv0 I merged in the latest changes from the main branch

corymhall (approved):

Issue #4670, opened 2024-10-26
TestIMDSAuth started failing


t0yv0:
### Describe what happened

I see TestIDMSAuth failing on separate changes:

- upgrading the bridge
- upgrading upstream

Surprisingly cron.yml and master.yml seem to indicate it passing. Maybe it will fail though if re-triggered. I suspect something changed out of band with respect to the repo history and needs investigating.

### Sample program

N/A

### Log output

N/A

### Affected Resource(s)

N/A

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
I can repro this outside of CI, taking a look.

PR #4673, opened 2024-10-26
Fix IMDSv2 test to allow /tmp to be immutable

t0yv0:
The IMDSv2 test provisions a RC of the pulumi-aws-provider inside an AWS test account and checks if it can authenticate. Unfortunately something changed in the cloud that no longer permits unix programs to make changes to `/tmp`. The test relied to this for no good reason; the updated version creates a dedicated folder /home/ec2-user/repro to perform the check.

Fixes https://github.com/pulumi/pulumi-aws/issues/4670

### DIFF
--- a/provider/test-programs/imds-auth/imds-v2/Pulumi.yaml
+++ b/provider/test-programs/imds-auth/imds-v2/Pulumi.yaml
@@ -94,7 +94,7 @@ resources:
       bucket: ${upload-bucket.id}
       key: pulumi-resource-aws
       source:
-        fn::FileAsset: ${localProviderBuild}
+        fn::fileAsset: ${localProviderBuild}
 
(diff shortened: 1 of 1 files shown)


provider/test-programs/imds-auth/imds-v2/Pulumi.yaml
```diff
@@ -94,7 +94,7 @@ resources:
       bucket: ${upload-bucket.id}
       key: pulumi-resource-aws
       source:
-        fn::FileAsset: ${localProviderBuild}
+        fn::fileAsset: ${localProviderBuild}
```
t0yv0, line 97: Removes a warning; the older form is obsolete.

flostadler (approved):

PR #4693, opened 2024-10-31
Remove slack notification from workflows

### LINKED ISSUE #4691: Workflow failure: master
## Workflow Failure

[master](https://github.com/pulumi/pulumi-aws/blob/master/.github/workflows/master.yml) has failed. See the list of failures below:

- [2024-10-30T20:54:15.000Z](https://github.com/pulumi/pulumi-aws/actions/runs/11599929839)

corymhall:
Slack notifications have been removed from ci-mgmt, but we still had
some specified in `.ci-mgmt.yaml`.

It looks like the Slack Secret was recently removed causing these jobs
to fail

closes #4691


t0yv0 (approved):
Nice, thank you!

PR #4714, opened 2024-11-04
Upstream v5.74.0

t0yv0:
Upgrade upstream to v5.74.0

New resources:

opensearch/authorizeVpcEndpointAccess.AuthorizeVpcEndpointAccess
ssm/quicksetupConfigurationManager.QuicksetupConfigurationManager

New functions:

lb/getListenerRule.getListenerRule

Fixes https://github.com/pulumi/pulumi-aws/issues/4707


flostadler:
What about the breaking change in the `getImageRecipe` function?

Upstream changed the output to use `TypeNullableBool` for the `deleteOnTermination` and `encrypted` [here](https://github.com/hashicorp/terraform-provider-aws/blob/2d8df6e0c897d90d71584735cd1f09d2a4ac7ffb/internal/service/imagebuilder/image_recipe_data_source.go#L46-L53). That type is a string under the hood, is there something we can do on the pulumi side to properly represent this as an optional bool?

flostadler:
The upstream PR that introduced it makes it seem that this is actually a bug fix and not a breaking change because the value was actually a string under the hood: https://github.com/hashicorp/terraform-provider-aws/pull/39928

Worth checking if that's also the case for pulumi, in that case we should be good to take the change.

t0yv0:
I have filed https://github.com/pulumi/pulumi-aws/issues/4723 to consider - there is a way to project these as booleans but we had some trouble with skipMetadataApiCheck as this complicates the job for the bridge. Currently this projects as a string:

```
    export interface GetImageRecipeBlockDeviceMappingEb {
        /**
         * Whether to delete the volume on termination. Defaults to unset, which is the value inherited from the parent image.
         */
        deleteOnTermination: string;
        /**
         * Whether to encrypt the volume. Defaults to unset, which is the value inherited from the parent image.
         */

```

I would not block the release on this unless we believe this data source is widely used.

flostadler (approved):
I think we're good with the breaking changes in the data source. That change is actually a bug fix

PR #4765, opened 2024-11-13
Add display name

iwahbe:


provider/cmd/pulumi-resource-aws/schema.json
```diff
@@ -1,5 +1,6 @@
 {
     "name": "aws",
+    "displayName": "AWS",
```
flostadler, line 3: I assume aws-native is gonna be called `AWS Cloud Control`?

flostadler (approved):

flostadler:
@iwahbe you have to rebuild the minimal schema as well (`make minimal_schema`)

provider/cmd/pulumi-resource-aws/schema.json
iwahbe: That's what it's called now:

https://github.com/pulumi/registry/blob/eb8278beb938bedbac4f28c911a1cd50644fa2fb/tools/resourcedocsgen/pkg/lookup.go#L131

t0yv0:
@corymhall can we get this one in here, I think @iwahbe move don. Thanks a ton!

Issue #4654, opened 2024-10-17
Support Exclusive Relationship Management Resources


t0yv0:
There is a series of changes where upstream provider is deprecating but not removing popular attributes in favor of more side-by-side resources.

https://hashicorp.github.io/terraform-provider-aws/design-decisions/exclusive-relationship-management-resources
https://github.com/hashicorp/terraform-provider-aws/issues/39376#issuecomment-2371332496

We need to look at how the pulumi-aws provider should handle these. In particular, migrating in Pulumi might need additional guidance specific to Pulumi, and building tooling to support it may require more effort.

To add some more links for inspecting the effect of these changes, https://github.com/hashicorp/terraform-provider-aws/releases/tag/v5.72.0 is one of the first releases where these changes are taking effect, and https://github.com/pulumi/pulumi-aws/pull/4647 is our PR of upgrading the provider to match.

t0yv0:
Marking P1 to prioritize making a final decision here on how to translate this to Pulumi.

PR #4769, opened 2024-11-13
Remove aws.iam.Role inlinePolicies deprecation

t0yv0:
This change removes aws.iam.Role inlinePolicies deprecation until Pulumi can support exclusive management resources better. In addition, managedPolicyArns deprecation removal is moved from a patch to regular code.  

The deprecation removal affects SDK generators only - documentation may still indicate that this is deprecated.

Fixes: https://github.com/pulumi/pulumi-aws/issues/4654
Context: https://github.com/pulumi/pulumi-aws/issues/4654


patches/0055-Create-Logging-Middleware-for-Lambda-service-that-do.patch
```diff
@@ -13,7 +13,7 @@ calls.
 
 diff --git a/internal/service/lambda/request_response_logger.go b/internal/service/lambda/request_response_logger.go
 new file mode 100644
-index 00000000000..737faef4a73
+index 0000000000..737faef4a7
```
t0yv0, line 16: Is this some configuration setting in my Git that is making this to churn?

provider/pkg/iam/iam.go
```diff
@@ -0,0 +1,42 @@
...
+	// Pulumi especially poorly, impacting the upgrade path for Pulumi users attempting to address the deprecation
+	// of inlinePolicy and managedPolicyArns. Until these are sorted out, deprecation notices are removed from
+	// the Pulumi AWS provider.
+	inlinePolicy, ok := iamRole.Schema["inline_policy"]
+	contract.Assertf(ok, "Expected an inline_policy parameter on the aws_iam_role resource")
+	inlinePolicy.Deprecated = ""
+
+	managedPolicyArns, ok := iamRole.Schema["managed_policy_arns"]
+	contract.Assertf(ok, "Expected an managed_policy_arns parameter on the aws_iam_role resource")
+	managedPolicyArns.Deprecated = ""
```
corymhall, line 41: Nice!

Issue #4780, opened 2024-11-15
Test `TestIMDSAuth/IDMSv2` broken


flostadler:
The test `TestIMDSAuth/IDMSv2` keeps on failing with this error:
```
Error: error resolving type of resource my-bucket: internal error loading package "aws": Could not automatically download and install resource plugin 'pulumi-resource-aws' at version v6.59.1, install the plugin using `pulumi plugin install resource aws v6.59.1`: error installing provider aws: untarring file /home/ec2-user/.pulumi/plugins/resource-aws-v6.59.1/pulumi-resource-aws: write /home/ec2-user/.pulumi/plugins/resource-aws-v6.59.1/pulumi-resource-aws: no space left on device
```

It shouldn't download the latest provider, but rather use the provider under test. This makes me think that downloading the provider under test from S3 is also failing: https://github.com/pulumi/pulumi-aws/blob/d13c83273b7f9c23e4c19cf2617da2396b704e64/provider/test-programs/imds-auth/imds-v2/Pulumi.yaml#L131

Issue #4383, opened 2024-08-19
Reduce provider binary on-disk size


t0yv0:
Consider looking at options to reduce provider on-disk size.

Per a customer comment: it grow from 5.16 (~400MB) to 6.49 (~800MB) unpacked on disk.

Benefits of a leaner on-disk provider:

- faster download times
- less pressure on CI environment to have sufficient disk space

Possible culprits here:

- embedded schema.json including more resources and examples for said resources, in more languages such as Java; could it be feasible to strip descriptions or at least examples from the schema distributed in the binary?

- more embedded provider metadata, is there any compression that can be applied?

- more Go dependencies statically linked in, anything that is possible to prune?

t0yv0:
Some information. Most of the size is present in the upstream provider build:

```
du -sh terraform-provider-aws                                                                                                                                                                     ~/code/terraform-provider-aws
752M    terraform-provider-aws
```

t0yv0:
From https://github.com/t0yv0/gobuildsize report on the upstream provider, major contributing packages are:

```
github.com/aws/aws-sdk-go-v2/service/ec2 137603662
github.com/hashicorp/terraform-provider-aws/internal/service/lexv2models 90502114
github.com/hashicorp/terraform-provider-aws/internal/service/batch 65336638
github.com/hashicorp/terraform-provider-aws/internal/service/ec2 60966940
github.com/hashicorp/terraform-provider-aws/internal/service/bedrockagent 59476792
github.com/aws/aws-sdk-go/service/sagemaker 56084368
github.com/aws/aws-sdk-go/service/quicksight 49658874
github.com/aws/aws-sdk-go-v2/service/iot 49391552
github.com/aws/aws-sdk-go-v2/service/glue 48909028
github.com/hashicorp/terraform-provider-aws/internal/service/securitylake 44860032
github.com/hashicorp/terraform-provider-aws/internal/service/cognitoidp 44200606
github.com/hashicorp/terraform-provider-aws/internal/service/securityhub 43308668
github.com/aws/aws-sdk-go-v2/service/rds 42986552
github.com/hashicorp/terraform-provider-aws/internal/service/verifiedpermissions 42005986
github.com/hashicorp/terraform-provider-aws/internal/service/appfabric 40250200
github.com/hashicorp/terraform-provider-aws/internal/service/rekognition 39744438
github.com/hashicorp/terraform-provider-aws/internal/service/ssmcontacts 39644536
[...]

PR #4781, opened 2024-11-15
Fix `TestIMDSAuth` test by increasing root volume size

flostadler:
The `TestIMDSAuth` was broken because the default 8GB root volume was not enough anymore to fit pulumi and the aws provider.
This change fixes that.

Fixes https://github.com/pulumi/pulumi-aws/issues/4780


t0yv0 (approved):
Hm. Did the provider size increase drastically recently?

flostadler:
> Hm. Did the provider size increase drastically recently?

Yeah it did: https://github.com/pulumi/pulumi-aws/issues/4383

PR #4797, opened 2024-11-18
Fix TestIMDS flaky test

corymhall:
The ami filter we were using picked up a new AMI which had a minimum
volume size of 30. This PR modifies the filter to be a little more
specific to filter out these special case AMIs.

e.g.
- `amazon/al2023-ami-2023.6.20241111.0-kernel-6.1-x86_64` (good)
- `amazon/al2023-ami-ecs-neuron-hvm-2023.0.20241115-kernel-6.1-x86_64` (bad)


t0yv0 (approved):
Thanks so much!

t0yv0 (approved):

PR #4828, opened 2024-11-25
Upstream v5.77.0

t0yv0:
Update upstream provider to v5.77.0

Fixes https://github.com/pulumi/pulumi-aws/issues/4819

New resources:

* rds/instanceDesiredState.InstanceDesiredState

Ephemeral resources introduced upstream are not yet supported in Pulumi but this is tracked in https://github.com/pulumi/pulumi-aws/issues/4831


provider/resources.go
```diff
@@ -3130,6 +3130,11 @@ compatibility shim in favor of the new "name" field.`)
 					},
 				},
 			},
+			"aws_rds_instance_state": {
+				// Pulumi cannot call this resource InstanceState because SDKs such as the Node SDK
```
t0yv0, line 3134: Calling this out.

mjeffryes (approved):

PR #4829, opened 2024-11-25
chore: remove minimal schema from source

t0yv0:
Keeping a copy of minimal schema (introduced in https://github.com/pulumi/pulumi-aws/pull/4587) is proving to conflict with Pulumi tooling such as `upgrade-provider` and is not strictly necessary. With this change the minimal schema itself as well as CI checks to make sure it is up to date are removed. Instead it will be computed on-the-fly by the release job as before:
 
- GitHub actions invoke this: https://github.com/pulumi/pulumi-aws/blob/916e6f28039407d7dcb1a295c6f9e43613dd6f87/.github/workflows/build_provider.yml#L60

- Make extension ensures that this runs go generate prior to the actual build so that the minimal schema is re-computed: https://github.com/pulumi/pulumi-aws/blob/916e6f28039407d7dcb1a295c6f9e43613dd6f87/.mk/minimal_schema.mk#L9


provider/cmd/pulumi-resource-aws/main.go
```diff
@@ -30,16 +29,15 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
 
-//go:embed schema-embed.json
-var pulumiSchema []byte
-
-//go:embed schema-minimal-embed.json
-var pulumiMinimalSchema []byte
+//go:embed schema-*-embed.json
```
t0yv0, line 32: The advantage of the `*` is that code still compiles even when minimal schema is not present on disk (such as upon checkout).

corymhall (approved):

t0yv0:
I need one more tweak, but I think this can work.

PR #4830, opened 2024-11-25
chore: remove accidentally introduced files (#4794)

t0yv0:
#4794 introduced files that were not part of the repository prior. I think these might not be required.


flostadler (approved):
Uh good catch. upgrade-provider seemed to have picked those up for me locally :(

Issue #4759, opened 2024-11-12
Created EKS Addon does not get saved to state if it does not become active


flostadler:
### Describe what happened

When creating an EKS Addon, the provider will first send the `CreateAddon` API call to AWS and then wait for the addon to become active.
Some addons, like corends, take a longer time to become active and might hit wait timeouts.
If the resource creation fails while waiting for the Addon to become active, the resource isn't saved to state.

Re-running `pulumi up` is now guaranteed to fail because Pulumi will wants to create the addon, even though it already exists in the cluster.

As a workaround, users either need to delete the addon from the cluster manually or import the Addon into Pulumi state.

### Sample program

```
import * as pulumi from "@pulumi/pulumi";
import * as aws from "@pulumi/aws";
import * as awsx from "@pulumi/awsx";
import * as eks from "@pulumi/eks";

// Grab some values from the Pulumi configuration (or use default values)
const config = new pulumi.Config();
const vpcNetworkCidr = config.get("vpcNetworkCidr") || "10.0.0.0/16";

const env = "aws-addon-bug"

// Create a new VPC
const eksVpc = new awsx.ec2.Vpc("eks-vpc", {
    enableDnsHostnames: true,
    cidrBlock: vpcNetworkCidr,
});

const instanceRole = new aws.iam.Role('testrole', {
    assumeRolePolicy: JSON.stringify({
        Version: '2012-10-17',
        Statement: [
            {
                Effect: 'Allow',
                Principal: {
                    Service: 'ec2.amazonaws.com',
                },
                Action: 'sts:AssumeRole',
            },
        ],
    }),
    managedPolicyArns: [
        'arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy',
        'arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryPullOnly',
        'arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy',
    ],
})

const eksCluster = new eks.Cluster(`${env}-cluster`, {
    vpcId: eksVpc.vpcId,
    authenticationMode: eks.AuthenticationMode.Api,
    corednsAddonOptions: {
        enabled: false,
    },
    createOidcProvider: true,
    enabledClusterLogTypes: ['api', 'audit', 'authenticator'],
    fargate: false,
    instanceRole: instanceRole,
    kubeProxyAddonOptions: {
        enabled: false,
    },
    nodeAssociatePublicIpAddress: false,
[...]

PR #4898, opened 2024-12-10
Upgrade pulumi-terraform-bridge to v3.97.0

flostadler:
This PR was generated via `$ upgrade-provider pulumi/pulumi-aws --kind=bridge`.

---

- Upgrading pulumi-terraform-bridge from v3.96.0 to v3.97.0.

**Manual Changes**:
Please pay special attention to `provider/pkg/batch/compute_environment.go` and `provider/pkg/batch/compute_environment_test.go`. The test was previously using the `logging` package from the `unstable` module that shouldn't have been used outside of the bridge. This package was recently moved to be internal (https://github.com/pulumi/pulumi-terraform-bridge/pull/2687). To get around that we now pass in the logger as a dependency and fake it in tests.

Fixes https://github.com/pulumi/pulumi-aws/issues/4759
Fixes https://github.com/pulumi/pulumi-aws/issues/4894


VenelinMartinov (approved):
lgtm

PR #4909, opened 2024-12-11
Simplify OIDC extraTest

blampe:
Today `TestAccCloudWatch` and `TestAccCloudWatchOidcManual` are each executed twice. Once as part of the `test (nodejs, local)` (examples) job, and again as part of the `test_oidc` job defined in `extraTests`.

The `test_oidc` job executes `TestAccCloudWatchOidcManual` without any ambient AWS credentials, and then `TestAccCloudWatch` after assuming `OIDC_ROLE_ARN`.

This PR removes the extra `test_oidc` job by incorporating these tests into the usual examples job.
* `TestAccCloudWatch` and `TestAccCloudWatchOIDCAmbient` preserve the existing behavior of running these using the ambient credentials from the examples job. (Do we want to keep this?)
* `TestAccCloudWatchOIDCManual` and `TestAccCloudWatchOIDC` preserving the `test_oidc` behavior by running without ambient credentials and after assuming `OIDC_ROLE_ARN`, respectively.


corymhall (approved):
Yep, this is better!

PR #4911, opened 2024-12-11
Fix dependencies between workflows

corymhall:
This fixes dependencies for two of the custom workflows we have defined.

Reran `make ci-mgmt`

### DIFF
--- a/.ci-mgmt.yaml
+++ b/.ci-mgmt.yaml
@@ -129,7 +129,9 @@ extraTests:
 
   test_oidc:
     name: test_oidc
-    needs: build_sdk
+    needs:
+      - build_sdk
+      - build_provider
     permissions:
       contents: read
       id-token: write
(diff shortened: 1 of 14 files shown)


blampe:
See https://github.com/pulumi/pulumi-aws/pull/4909 -- we can simplify things by folding this test into the usual job.

corymhall:
> See https://github.com/pulumi/pulumi-aws/pull/4909 -- we can simplify things by folding this test into the usual job.

Good idea! I update the PR to only update the `provider_test` workflow.

blampe:
> > See #4909 -- we can simplify things by folding this test into the usual job.
> 
> Good idea! I update the PR to only update the `provider_test` workflow.

Similar thought in https://github.com/pulumi/pulumi-aws/pull/4910 :)

Looks like this failed due to missing `az` -- I added that bit recently. Not sure why it's missing here but I'll look into it.

PR #4917, opened 2024-12-12
Test shim as a normal unit test

blampe:
This removes the `go_test_shim` job defined in `extraTests` by making it a normal provider unit test instead of something that needs to be run as part of upstream.

We currently attempt to upload coverage as part of this test, but it [doesn't work](https://app.codecov.io/gh/pulumi/pulumi-aws). https://github.com/pulumi/ci-mgmt/pull/1224 will start uploading unit test coverage by default.

### DIFF
--- a/provider/shim_test.go
+++ b/provider/shim_test.go
@@ -1,0 +1,31 @@
+// Copyright 2024, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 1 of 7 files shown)


t0yv0 (approved):
I'm not 100% sure but I  think TestProviderShim given all it does is NewUpstreamProvider is adequately tested by the rest of the suite. There's no danger to coverage here - the only downside is not failing faster on some obvious mistakes. But that's fine.

provider/shim_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"context"
+	"testing"
+
+	shim "github.com/hashicorp/terraform-provider-aws/shim"
+)
+
+// This checks that any runtime checks in the underlying provider (with patches) are passed.
+func TestProviderShim(t *testing.T) {
```
t0yv0, line 25: Ah here it is. This is good!

blampe:
Upstream failure is unrelated
```
panic.go:629: Error running post-test destroy, there may be dangling resources: exit status 1
        
        Error: updating WAF ByteMatchSet (b47b2c2f-e5d8-437a-b3b7-da1027bf3781): operation error WAF: UpdateByteMatchSet, https response error StatusCode: 400, RequestID: abd63009-1807-4e89-8e98-a8bf9a595788, WAFInvalidOperationException: Operation is invalid for this entity.
```

PR #4918, opened 2024-12-12
Lint upstream as a normal unit test

blampe:
This removes the last job in `extraTests` by making it a standard unit test.

Currently the job runs without caching and takes upwards of 30-40 minutes because it re-compiles everything from scratch. This makes it the long poll when sharding is [turned on](https://github.com/pulumi/pulumi-aws/actions/runs/12321334449).

By running with the rest of our tests it can benefit from caching. Locally the test takes only ~20s.

This is run during unit tests, which still take about ~5 minutes before and after this change. They're actually slightly faster because I parallelized some of them while I was in here.


provider/resources.go
```diff
@@ -493,9 +494,13 @@ var namespaceMap = map[string]string{
 	"aws": "Aws",
 }
 
+var _nslock = sync.Mutex{}
```
t0yv0, line 497: Curious what happened here, we ran awsMember call from multiple goroutines?

t0yv0: That's probably because of all the t.Parallel introduced into the tests? Interesting.

t0yv0 (approved):

blampe:
Assuming the upstream waf tests are unrelated.

Issue #4760, opened 2024-11-12
SQS RedriveAllowPolicy shows spurious JSON diff


csssuf:
### Describe what happened

Similar to #2307, a JSON diff always shows up for the `sqs.RedriveAllowPolicy` resource, even when no actual changes will be made.

### Sample program

```python
import pulumi
from pulumi_aws import sqs

dlq = sqs.Queue("demo-dlq")
queue = sqs.Queue(
    "demo-queue",
    redrive_policy=pulumi.Output.json_dumps(
        {
            "deadLetterTargetArn": dlq.arn,
            "maxReceiveCount": 2,
        }
    )
)

redrive_allow_policy = sqs.RedriveAllowPolicy(
    "demo",
    queue_url=dlq.id,
    redrive_allow_policy=pulumi.Output.json_dumps(
        {
            "redrivePermission": "byQueue",
            "sourceQueueArns": [queue.arn],
        }
    )
)
```

### Log output

```
Enter your passphrase to unlock config/secrets
Previewing update (dev):
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev::pulumi-sqs-redrive-allow-policy::pulumi:pulumi:Stack::pulumi-sqs-redrive-allow-policy-dev]
    ~ aws:sqs/redriveAllowPolicy:RedriveAllowPolicy: (update)
        [id=https://sqs.us-east-1.amazonaws.com/<snip>/demo-dlq-478d13f]
        [urn=urn:pulumi:dev::pulumi-sqs-redrive-allow-policy::aws:sqs/redriveAllowPolicy:RedriveAllowPolicy::demo]
        [provider=urn:pulumi:dev::pulumi-sqs-redrive-allow-policy::pulumi:providers:aws::default_6_59_0::a91c1d2b-d05e-49d0-9084-4ff4e6e2003d]
      ~ redriveAllowPolicy: "\"{\\\"redrivePermission\\\": \\\"byQueue\\\", \\\"sourceQueueArns\\\": [\\\"arn:aws:sqs:us-east-1:<snip>:demo-queue-b866e9f\\\"]}\"" => "\"{\\\"redrivePermission\\\": \\\"byQueue\\\", \\\"sourceQueueArns\\\": [\\\"arn:aws:sqs:us-east-1:<snip>:demo-queue-b866e9f\\\"]}\""
Resources:
    ~ 1 to update
    3 unchanged
```

### Affected Resource(s)

[...]

PR #4957, opened 2024-12-20
Upgrade terraform-provider-aws to v5.82.1

### LINKED ISSUE #4953: Upgrade terraform-provider-aws to v5.82.1
Release details: https://github.com/hashicorp/terraform-provider-aws/releases/tag/v5.82.1

> [!NOTE]
> This issue was created automatically by the upgrade-provider tool and should be automatically closed by a subsequent upgrade pull request.

flostadler:
This PR was generated via `$ upgrade-provider pulumi/pulumi-aws --kind=provider --target-version=5.82.1`.

New resources:

apigateway/domainNameAccessAssociation.DomainNameAccessAssociation
cloudfront/vpcOrigin.VpcOrigin
memorydb/multiRegionCluster.MultiRegionCluster
networkmanager/dxGatewayAttachment.DxGatewayAttachment
rds/clusterSnapshotCopy.ClusterSnapshotCopy

New functions:

servicecatalog/getAppregistryAttributeGroupAssociations.getAppregistryAttributeGroupAssociations

---

- Upgrading terraform-provider-aws from 5.81.0  to 5.82.1.
	Fixes #4953

Fixes https://github.com/pulumi/pulumi-aws/issues/4760


flostadler:
"aws:elasticache/replicationGroup:ReplicationGroup":
🟡 inputs: "atRestEncryptionEnabled" type changed from "boolean" to "string"
🟡 properties: "atRestEncryptionEnabled" type changed from "boolean" to "string"

That seems bad at a first glance. Checking why that happened

flostadler:
https://github.com/hashicorp/terraform-provider-aws/pull/40514 changed `atRestEncryptionEnabled` of `aws:elasticache/replicationGroup:ReplicationGroup` to a nullable bool which we model as a string in the bridge (https://github.com/pulumi/pulumi-terraform-bridge/issues/1514).

The `ReplicationGroup` is a rather widely used resource and `atRestEncryptionEnabled` is a setting almost everybody running Redis/Valkey in production would set. This is a rather severe breaking change that we cannot ship in a minor IMO

flostadler:
`regress-4457` is failing because a new property was added to `autoscaling.Group`. That test is executing `pulumi import` and then editing the code to add the necessary changes. The problem is that our python tests use the latest release version of the SDK and that does not include newly added properties, but `pulumi import` will include them in the generated code. Chicken/Egg problem...

We should rewrite that test to use nodejs (or yaml?) because they are not susceptible to this issue.

flostadler:
Added override for ReplicationGroup `atRestEncryptionEnabled` to be a boolean and added a test for it

examples/provider_test.go
```diff
@@ -317,3 +324,32 @@ func planEqual(t *testing.T, firstPlan, secondPlan string) bool {
 
 	return assert.Equal(t, firstPlanData, secondPlanData)
 }
+
+func inPlacePulumiTest(t *testing.T, dir string, opts ...opttest.Option) *pulumitest.PulumiTest {
```
flostadler, line 328: This allows us now to use an in-process provider for the tests. Which will also give us code coverage :)

examples/examples_py_test.go
```diff
@@ -179,6 +179,7 @@ func TestRegress2534(t *testing.T) {
 }
 
 func TestRegress4457(t *testing.T) {
+	t.Skipf("TODO[pulumi/pulumi-aws#4957]")
```
t0yv0, line 182: What's happening here?

flostadler: Ah wait, just noticed I linked the PR and not the issue

flostadler: corrected the link, thanks for noticing!

examples/provider_test.go
t0yv0: This is good but might want to be careful to not use this for production release verification in case there's accidentally some behavioral difference between the prod binary and this form of running the provider.

examples/test-programs/elasticache-replication-group/Pulumi.yaml
```diff
@@ -0,0 +1,17 @@
...
+runtime: yaml
+config:
+  pulumi:tags:
+    value:
+      pulumi:template: aws-yaml
+outputs:
+  replicationGroupArn: ${replicationGroup.arn}
+resources:
+  replicationGroup:
+    type: aws:elasticache:ReplicationGroup
```
t0yv0, line 11: Nice.

t0yv0 (approved):
Thanks!

examples/provider_test.go
flostadler: Yeah, before adopting this for all tests we should add some sort of flag/env variable to fall back to attaching the binary for releases

t0yv0:
This was in https://github.com/hashicorp/terraform-provider-aws/releases/tag/v5.82.0

PR #4971, opened 2024-12-20
Remove unused npm dependencies

justinvp:
It looks like the `resolve` and `builtin-modules` dependencies aren't actually used anywhere, so stop depending on them.

The motivation for this change is that the dependency on `resolve` causes errors to be logged in logs during plugin discovery, because the package includes tests that have malformed `package.json` files. If we don't actually need the dependency it'd be nice to remove it so those errors are no longer logged (so users don't [ask us about them](https://github.com/pulumi/pulumi/issues/17578)).

Also remove the `builtin-modules` dependency while cleaning this up, since it doesn't appear to be used either AFAICT.

Similar to https://github.com/pulumi/pulumi-aws/pull/3238
Reference: https://github.com/pulumi/pulumi/issues/17578

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -5275,9 +5275,7 @@ compatibility shim in favor of the new "name" field.`)
 		},
 		JavaScript: &tfbridge.JavaScriptInfo{
 			Dependencies: map[string]string{
-				"mime":            "^2.0.0",
-				"builtin-modules": "3.0.0",
-				"resolve":         "^1.7.1",
+				"mime": "^2.0.0",
 			},
 			DevDependencies: map[string]string{
 				"@types/node": "^10.0.0", // so we can access strongly typed node definitions.
--- a/sdk/nodejs/package.json
+++ b/sdk/nodejs/package.json
@@ -14,9 +14,7 @@
     },
     "dependencies": {
         "@pulumi/pulumi": "^3.142.0",
-        "builtin-modules": "3.0.0",
-        "mime": "^2.0.0",
-        "resolve": "^1.7.1"
+        "mime": "^2.0.0"
     },
     "devDependencies": {
         "@types/mime": "^2.0.0",


t0yv0 (approved):
These dependencies are from 2018 it seems reasonable that we stopped actually depending on these.

justinvp:
What's going on with the failing waf test? Something we need to clean up in a test account?

```
=== NAME  TestAccWAFRuleGroup_basic
    rule_group_test.go:34: Step 1/2 error: Error running apply: exit status 1
        
        Error: creating WAF Rule Group (tfaccze0ep): operation error WAF: CreateRuleGroup, https response error StatusCode: 400, RequestID: ac6aaf1f-be43-45b9-8044-589560f4e5b7, WAFLimitsExceededException: Operation would result in exceeding resource limits.
        
          with aws_waf_rule_group.test,
          on terraform_plugin_test.tf line 17, in resource "aws_waf_rule_group" "test":
          17: resource "aws_waf_rule_group" "test" {
        
--- FAIL: TestAccWAFRuleGroup_basic (15.67s)
```

t0yv0:
Yeah I think so. Having a quick look.

t0yv0:
```
=== NAME  TestAccWAFRuleGroup_changeNameForceNew
    rule_group_test.go:75: Step 1/3 error: Error running apply: exit status 1

        Error: creating WAF Rule Group (tfaccyzotq): operation error WAF: CreateRuleGroup, https response error StatusCode: 400, RequestID: fcb7db05-54ed-4ab1-8a3b-cecdb02574fb, WAFLimitsExceededException: Operation would result in exceeding resource limits.

          with aws_waf_rule_group.test,
          on terraform_plugin_test.tf line 17, in resource "aws_waf_rule_group" "test":
          17: resource "aws_waf_rule_group" "test" {

```

ForceNew may be trying to create two of these. Scrolling back a little I think our intent with this test was to smoke-test upstream post-patching with their own tests, somehow we selected `waf` instead of the newer `wafv2` module for this purpose. Rather than raising our account `waf` quotas we can disable this test or switch to wafv2.

t0yv0:
https://github.com/pulumi/pulumi-aws/pull/5001 I'll land that first and rebase this to get it in.

t0yv0:
Thank you for getting this in!

PR #5013, opened 2024-12-27
chore: configure Renovate to ignore bridge and pkg/sdk dependencies

t0yv0:
Adds a config file for the Renovate bot.

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -1,0 +1,22 @@
+{
+  $schema: "https://docs.renovatebot.com/renovate-schema.json",
+  extends: ["github>pulumi/renovate-config//default.json5"],
+  packageRules: [
+    {
+      // These dependencies are special and are pulled in when the bridge dependency updates:
+      //
+      //     github.com/pulumi/pulumi-terraform-bridge/v3
+      //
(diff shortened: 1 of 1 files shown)


renovate.json5
```diff
@@ -0,0 +1,22 @@
...
+      matchManagers: ["gomod"],
+      matchPackageNames: ["github.com/pulumi/pulumi/pkg/v3", "github.com/pulumi/pulumi/sdk/v3"],
+      enabled: false
+    },
+    {
+      // Currently the bridge dependency is managed by pulumi/upgrade-provider
+      matchManagers: ["gomod"],
+      matchPackageNames: ["github.com/pulumi/pulumi-terraform-bridge/v3"],
+      enabled: false
+    }
```
blampe, line 20: @t0yv0 something interesting you could do is leave this enabled but set `automerge: false`, `ignoreUnstable: false`, and [`respectLatest: false`](https://docs.renovatebot.com/configuration-options/#respectlatest).

I think that would give you a perpetual canary PR that Renovate would periodically update to the latest bridge. (You could put it on a daily schedule if you don't want every commit to trigger an update.)

blampe (approved):

PR #5053, opened 2025-01-08
chore: enable release verification take 2

corymhall:
This enables release verification for AWS.

Tested by manually executing the `Verify-Release` workflow using this branch. Run can be seen here https://github.com/pulumi/pulumi-aws/actions/runs/12691069824

closes https://github.com/pulumi/ci-mgmt/issues/1265

### DIFF
--- a/examples/release-verification/index.ts
+++ b/examples/release-verification/index.ts
@@ -1,0 +1,71 @@
+// Copyright 2016-2018, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


examples/webserver-py/Pulumi.yaml
```diff
@@ -1,3 +1,6 @@
 name: webserver-py
-runtime: python
+runtime:
+  name: python
+  options:
```
corymhall, line 4: This is required based on https://github.com/pulumi/verify-provider-release?tab=readme-ov-file#python

examples/release-verification/index.ts
corymhall, line 1: All the existing nodejs example tests required a config value to be set. Copied over the `bucket` example and removed that config.

upstream.sh
```diff
@@ -23,6 +23,7 @@ COMMANDS
   check_in              Write checkedout commits back to patches, add upstream
                         and patches changes to the git staging area and exit
                         checkout mode.
+  file_target           Print a file path to depend on in make.
```
t0yv0, line 26: This came from ci-mgmt yes?

t0yv0 (approved):
Let's try this.

PR #5088, opened 2025-01-11
Added thailand region

tawanorg:
Added the new AWS Thailand (Bangkok) region `ap-southeast-7` to the Region enum/interfaces.

This region was recently announced by AWS and is now generally available.

Reference: https://aws.amazon.com/local/thailand/


corymhall:
@tawanorg thanks for the PR. We're unable to merge contributions from outside contributors directly due to the way our GitHub workflows work so I merged your commit into my branch. Your fix just merged in #5104 🎉

PR #5110, opened 2025-01-15
chore: remove make extenders

t0yv0:
Move minimal schema computation from a .mk/ Make extension to a script. Update ci-mgmt workflows.


.ci-mgmt.yaml
```diff
@@ -5,6 +5,9 @@ parallel: 1
 timeout: 150
 generate-nightly-test-workflow: true
 providerVersion: github.com/hashicorp/terraform-provider-aws/version.ProviderVersion
+
+buildProviderPre: "VERSION=${VERSION_GENERIC} ./scripts/minimal_schema.sh"
```
flostadler, line 9: So we basically replaced makefile extensions with a generic shell based pointcut?

flostadler (approved):

.ci-mgmt.yaml
t0yv0: Yes. It has a slight advantage that it does not interact with `make --touch` wildly, and it's documented in the cimgmt config.

PR #5112, opened 2025-01-17
Fix NOTEs on BucketReplicationConfigRules

### LINKED ISSUE #4921: Missing important upstream documentation related to `BucketReplicationConfig`
Upstream there is some [important documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_replication_configuration#rule) that isn't present in [our documentation](https://www.pulumi.com/registry/packages/aws/api-docs/s3/bucketreplicationconfig/#bucketreplicationconfigrule)
> [!NOTE]  
> The existing_object_replication parameter is not supported by Amazon S3 at this time and should not be included in your rule configurations. Specifying this parameter will result in MalformedXML errors. To replicate existing objects, please refer to the [Replicating existing objects with S3 Batch Replication](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-batch-replication-batch.html) documentation in the Amazon S3 User Guide.

This is important to doing replication and can lead users to incorrectly configure the bucket replication and hit an error without a clear direction forward.

guineveresaenger:
While this fix is not the complete underlying docsgen fix that would fix related issues, it will resolve the specific instance of difficulties experienced due to missing NOTEs on Rule.:wq

Fixes #4921.

- **Fix up Bucket Replication Config with missing note**
- **build schema and SDKs**


provider/cmd/pulumi-resource-aws/schema.json
```diff
@@ -351282,7 +351282,7 @@
...
                     },
-                    "description": "List of configuration blocks describing the rules managing the replication. See below.\n"
+                    "description": "List of configuration blocks describing the rules managing the replication. See below.\n\u003e **NOTE:** Replication to multiple destination buckets requires that `priority` is specified in the `rule` object. If the corresponding rule requires no filter, an empty configuration block `filter {}` must be specified.\n\n\u003e **NOTE:** Amazon S3's latest version of the replication configuration is V2, which includes the `filter` attribute for replication rules.\n\n\u003e **NOTE:** The `existing_object_replication` parameter is not supported by Amazon S3 at this time and should not be included in your `rule` configurations. Specifying this parameter will result in `MalformedXML` errors.\nTo replicate existing objects, please refer to the [Replicating existing objects with S3 Batch Replication](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-batch-replication-batch.html) documentation in the Amazon S3 User Guide.\n"
```
iwahbe, line 351285: We should probably use TypeScript notation here, not HCL.

```suggestion
                    "description": "List of configuration blocks describing the rules managing the replication. See below.\n\u003e **NOTE:** Replication to multiple destination buckets requires that `priority` is specified in the `rule` object. If the corresponding rule requires no filter, an empty configuration block `filter: {}` must be specified.\n\n\u003e **NOTE:** Amazon S3's latest version of the replication configuration is V2, which includes the `filter` attribute for replication rules.\n\n\u003e **NOTE:** The `existingObjectReplication` parameter is not supported by Amazon S3 at this time and should not be included in your `rule` configurations. Specifying this parameter will result in `MalformedXML` errors.\nTo replicate existing objects, please refer to the [Replicating existing objects with S3 Batch Replication](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-batch-replication-batch.html) documentation in the Amazon S3 User Guide.\n"
```

iwahbe (approved):
One nit, otherwise this looks good.

provider/cmd/pulumi-resource-aws/schema.json
guineveresaenger: nice catch! ty!

provider/doc_edits.go
```diff
@@ -134,6 +135,24 @@ var fixUpEcsServiceNameForceNewDeployment = targetedSimpleReplace(
 		"(e.g., `myimage:latest`), roll Fargate tasks onto a newer platform version, or immediately deploy "+
 		"`ordered_placement_strategy` and `placement_constraints` updates.\n"+
 		"When using the forceNewDeployment property you also need to configure the triggers property.\n")
+var fixUpBucketReplicationConfig = targetedSimpleReplace(
+	"s3_bucket_replication_configuration.html.markdown",
```
t0yv0, line 139: Could we put a check in (a test) that ensures this content is matching upstream, if not take from upstream content directly? This way if upstream edits it away we'll remember to update the copy.

guineveresaenger: We have hard failure patterns during `make schema`; I'm happy to add those. we should add them for the other fixes as well. It may be noisy during upgrades; that's the tradeoff. I think you're right to request this though.

t0yv0: Not sure what the "hard failure patterns" are but appreciated - the change looks like we add content we own where in fact this is a workaround to propagate upstream content, this postpones the issue of possible discrepancies to the future, and I must assume this page is high visibility since we are attempting to patch it. Thanks!

guineveresaenger: I've added a new function that follows the [suggested pattern](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/docs/guides/docs-edits.md#edit-rules).

I'm going to follow up with a PR that addresses this for other large replaces.

provider/doc_edits.go
```diff
@@ -109,6 +109,26 @@ func targetedSimpleReplace(filePath, from, to string) tfbridge.DocsEdit {
 	}
 }
 
+func targetedReplace(filePath, from, to string) tfbridge.DocsEdit {
+	fromBytes, toBytes := []byte(from), []byte(to)
+	return tfbridge.DocsEdit{
+		Path: filePath,
+		Edit: func(_ string, content []byte) ([]byte, error) {
+			if bytes.Contains(content, fromBytes) {
```
t0yv0, line 117: Unfortunately this does not ensure that toBytes does not digress, and yet we copied it from the source material. IN this case the toBytes is the bit that contains the notes we're copying.

guineveresaenger: You're totally right. The new check is more specific to the situation and checks that both the added text and the fromBytes have not changed, else it errors.

provider/doc_edits.go
```diff
@@ -134,6 +155,74 @@ var fixUpEcsServiceNameForceNewDeployment = targetedSimpleReplace(
...
+		} else {
+			// Hard error to ensure we keep this content up to date
+			return nil, fmt.Errorf("could not find NOTE snippets in upstream %s, "+
+				"please verify replace content in doc_edits.go: %s", path, string(noteBytes))
+		}
+		return content, nil
+	},
+}
+
+//				targetedReplace(
```
t0yv0, line 208: Not needed?

t0yv0 (approved):

PR #5133, opened 2025-01-25
Add hard errors to a few more involved docs replaces

guineveresaenger:
While working on https://github.com/pulumi/pulumi-aws/pull/5112, I noticed a few more docs replaces could use a hard error in the case where the upstream docs source changes. `make schema` will fail in this case, and we will have an opportunity to update the doc correctly.

Drive-by PR; no associated issue. 😅


t0yv0 (approved):
Thanks!

PR #5161, opened 2025-02-04
Upstream 5.85.0

corymhall:
Upgrade [terraform-provider-aws to v5.85.0](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v5.85.0)


corymhall:

- #5161 ◀

patches/0071-Restore-conns-factory-for-SDKv1-Go-clients-used-by-p.patch
```diff
@@ -48,8 +48,8 @@ index 57bf4c7c94..09b4279db3 100644
 +		}
 +	}
 +
-+	sp, ok := c.ServicePackages[servicePackageName]
-+	if !ok {
++	sp := c.ServicePackage(ctx, servicePackageName)
```
corymhall, line 51: `c.ServicePackages` was changed to a private type with accessor methods.

provider/resources.go
```diff
@@ -6023,12 +6027,20 @@ func setupComputedIDs(prov *tfbridge.ProviderInfo) {
 		return attr(state, "multiRegionClusterName"), nil
 	}
 
+	prov.Resources["aws_guardduty_member_detector_feature"].ComputeID = func(
+		ctx context.Context, state resource.PropertyMap,
+	) (resource.ID, error) {
+		return attrWithSeparator(state, ",", "detectorId", "accountId", "name"), nil
```
corymhall, line 6033: This resource currently doesn't support import, so I'm not 100% sure this is correct. Based on the `Read` method it looks like these three are required

https://github.com/hashicorp/terraform-provider-aws/blob/658a64579f91e53276d2a71da00aa43b854d3de7/internal/service/guardduty/member_detector_feature.go?plain=1#L143

provider/resources.go
```diff
@@ -4666,6 +4668,8 @@ compatibility shim in favor of the new "name" field.`)
 			"aws_ec2_managed_prefix_list":          {Tok: awsDataSource(ec2Mod, "getManagedPrefixList")},
 			"aws_ec2_transit_gateway_route_tables": {Tok: awsDataSource(ec2Mod, "getTransitGatewayRouteTables")},
 			"aws_ec2_instance_types":               {Tok: awsDataSource(ec2Mod, "getInstanceTypes")},
+			"aws_vpc_ipam":                         {Tok: awsDataSource(ec2Mod, "getVpcIpam")},
```
corymhall, line 4671: This is `aws_vpc_*`, but everything else we've been putting in `aws_ec2` so just following that.

provider/resources.go
t0yv0: Ah yes. The ID allocation is best-effort. Technically new GUID is a valid ID. We've been trying to match ID  to import IDs where possible.

provider/resources.go
t0yv0: Which module does it show up in the TF documentation site? Alongside which resources? Can we match that for minimum surprise?

corymhall: Well the problem is that we haven't been matching TF so far. The TF docs site has a separate section for `Vpc` (which includes a lot more resources that our Vpc docs section) and a separate `Vpc IPAM` section.

It seems like we've already diverged pretty far from what the TF docs do, but that also might be because they are able to re-organize the resources on the docs site when they want to.

Screenshot 2025-02-05 at 10 15 05 AM


Also, it might be a moot point because I didn't realize we already had a `aws_vpc_ipam` `resource` that we put under the `ec2` module so I think we need to also put the datasource there as well.

patches/0035-Disable-retry-for-KMS-access-denied-in-lambda.patch
```diff
@@ -45,14 +45,14 @@ index 0000000000..79867519db
...
-index df09e00772..432b494788 100644
+index d05c9a6185..7a25d187fe 100644
 --- a/internal/service/lambda/service_package_gen.go
 +++ b/internal/service/lambda/service_package_gen.go
 @@ -164,6 +164,7 @@ func (p *servicePackage) NewClient(ctx context.Context, config map[string]any) (
- 	return lambda.NewFromConfig(cfg,
  		lambda.WithEndpointResolverV2(newEndpointResolverV2()),
  		withBaseEndpoint(config[names.AttrEndpoint].(string)),
+ 		withExtraOptions(ctx, p, config),
 +		p.pulumiCustomizeLambdaRetries(cfg),
```
t0yv0, line 55: Anything special going on here?

provider/resources.go
corymhall: That's a good idea, we should think about how to reorganize these resources for V7. I'll create an issue to track.

t0yv0 (approved):
PATCH changes LGTM. Let me know if need a hand fighting CI.

patches/0035-Disable-retry-for-KMS-access-denied-in-lambda.patch
corymhall: Looks like they added the `withExtraOptions` thing in there so I just fixed the rebase to add our customization after.

provider/resources.go
corymhall: https://github.com/pulumi/pulumi-aws/issues/5168

Issue #5217, opened 2025-02-10
Protocol missing in Resource Configuration example


Lucapaulo:
[themes/default/content//registry/packages/aws/api-docs/vpclattice/resourceconfiguration](https://www.pulumi.com/registry/packages/aws/api-docs/vpclattice/resourceconfiguration/)

### IP Address Example

See in line comment

```
import pulumi
import pulumi_aws as aws

example = aws.vpclattice.ResourceConfiguration("example",
    name="Example",
    resource_gateway_identifier=example_aws_vpclattice_resource_gateway["id"],
    port_ranges=["80"],
    protocol="TCP", # this is missing in the example, it will throw an error otherwise
    resource_configuration_definition={
        "ip_resource": {
            "ip_address": "10.0.0.1",
        },
    },
    tags={
        "Environment": "Example",
    })
```

guineveresaenger:
Thank you for letting us know. We'll address this as soon as we can.

guineveresaenger:
Fixed via #5221

PR #5221, opened 2025-02-20
Fix incorrect example for Vpclattice ResourceConfiguration

guineveresaenger:
Fixes #https://github.com/pulumi/pulumi-aws/issues/5217

 **Add doc edits rule**
- **make schema and build SDKs**


provider/doc_edits.go
```diff
@@ -232,6 +233,47 @@ var fixUpBucketReplicationConfig = tfbridge.DocsEdit{
...
+			return nil, fmt.Errorf("could not find text in upstream %s, "+
+				"please verify replace content in doc_edits.go: %s", path, string(fromBytes),
+			)
+		}
+		return content, nil
+	},
+}
+
+//
+//				 targetedReplace(
```
corymhall, line 270: Looks like some missed cleanup

corymhall (approved):

PR #5247, opened 2025-02-27
chore: unpin .NET 6

t0yv0:
This repository does not need to pin .NET 6 version with respect to ci-mgmt defaults.


corymhall (approved):
Looks like we need to update any `TargetFramework`s

https://github.com/pulumi/pulumi-aws/blob/0013ca4a6b9f022242f4aac2127994f1671dd7fd/examples/sqs-fifo-queue/csharp/testing-aws-fifo-dotnet.csproj#L5-L5

t0yv0:
Yeah Eron wrote a migration, but a previous migration is broken in AWS and won't run eh. I'll fix this.

Issue #5250, opened 2025-02-27
Remove dangling references from schema


t0yv0:
### Describe what happened

In https://github.com/pulumi/pulumi-aws/issues/2565 we looked at making the schema conforming to checks by removing dangling references. Unfortunately due to legacy overlays this change is breaking. And as a workaround a "minimal schema" is made available that does not model these overlays but is otherwise conformant. This ticket is a reminder that at the next major version of the provider we should clean this up and take the breaking changes so there is only one schema and it is the good done.

### Sample program

N/A

### Log output

N/A

### Affected Resource(s)

N/A

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

guineveresaenger:
https://github.com/pulumi/pulumi-aws/pull/5508 resolves the dangling Region reference

PR #5292, opened 2025-03-13
Add PolicyDocument types to KMS inputs

kmosher:
KMS key policies have slightly stricter requirements around the `Principal` field then normal policy documents, but are otherwise shaped the same.

This mirrors our other input types for resources that support policy documents. While the Typescript gets nicer. Looking at the generated golang, it seems like it makes things slightly worse though. Since our golang sdk doesn't seem to support an `aws.iam.PolicyDocument` type and recommends `GetPolicyDocument` as universal practice, should SDK gen just ignore that alt type and stick to strings?

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -2869,12 +2869,30 @@ compatibility shim in favor of the new "name" field.`)
 					}),
 				},
 			},
-			"aws_kms_ciphertext":           {Tok: awsResource(kmsMod, "Ciphertext")},
-			"aws_kms_custom_key_store":     {Tok: awsResource(kmsMod, "CustomKeyStore")},
-			"aws_kms_external_key":         {Tok: awsResource(kmsMod, "ExternalKey")},
-			"aws_kms_grant":                {Tok: awsResource(kmsMod, "Grant")},
-			"aws_kms_key":                  {Tok: awsResource(kmsMod, "Key")},
-			"aws_kms_key_policy":           {Tok: awsResource(kmsMod, "KeyPolicy")},
+			"aws_kms_ciphertext":       {Tok: awsResource(kmsMod, "Ciphertext")},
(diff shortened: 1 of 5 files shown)


t0yv0:
This is an interesting change, I'm on the fence about this, thanks @kmosher for submitting. The TypeScript gets nicer indeed. Go change is non-breaking but loses precision - that is probably acceptable as this is how the unions are encoded in Go. What was more worrisome to me is that Python and .NET are not changing. I think this is because the token used in the union declaration is dangling and not defined in the schema. 

I think this relies on the same undocumented workaround, as used elsewhere in this provider, to permit a better experience for TypeScript only at the expense of other languages. 

We have committed to clean up this by the v-next major at the request of our partners (https://github.com/pulumi/pulumi-aws/issues/5250) 

If we take the PR, it will add more to the scope of that cleanup, and will break all programs that rely on the new functionality one more time.

In light of this I think we should decline for now.

However, if you are committed to the change, perhaps we could work with you to express it in terms of correctly schematized types, that is instead of "#/types/aws:iam/documents:PolicyDocument" actually schematize a type as needed. This could be interesting. For TypeScript replacing something with a union would be non-breaking, but I wonder if all the other languages bear the same guarantee. I'm not sure off the top of my head. If this change is indeed non-breaking we could ship it in the v6 series.

kmosher:
Hmm, yeah, I agree it's pointless to do this is we're going to have to turn around and rip it out for v7. I love the extra type information in Typescript, but definitely feels like we need to step back on the implementation to make it work better across all our SDKs.

kmosher:
@t0yv0 wrote this all up in https://github.com/pulumi/pulumi-aws/issues/5338. This is where I'm not a providers dev, so I've no idea where the different approaches fall on the spectrum of "sensible extension" to "will need a total rewrite to avoid horrible hacks to support"

PR #5339, opened 2025-03-18
Remove remove-ephemeral-resources.patch

### LINKED ISSUE #5231: PATCH 0085-Remove-ephemeral-resources.patch
This patch removes changes to resources that now have ephemeral (write-only) resource fields

It can be removed when either https://github.com/pulumi/pulumi-aws/issues/5215 is implemented, or when https://github.com/pulumi/pulumi/issues/16600 is fully finished, whichever comes first.

guineveresaenger:
This pull request combines new bridge functionality and an attribute traversal to automate omitting any of the kind of write-only and write-only-adjacent resources that a patch addressed previously.

While there's a bit of metadata tracking cruft due to the attribute traversal wanting build-time information, the schema itself has no changes.

Fixes #5231.
- **Remove Remove-ephemeral-resources.patch and reformat patches**
- **Traverse properties and filter write-only supporting ones, then Omit them**
- **Using bridge-supported traversal implies some metadata tracking**


provider/resources.go
```diff
@@ -5718,6 +5719,24 @@ compatibility shim in favor of the new "name" field.`)
...
 
+	// TODO[pulumi/pulumi-terraform-bridge#2938]
+	// Currently, the bridge omits write-only attributes from being written to a provider schema.
+	// In this provider, we additionally have some attributes  called things like "has_value_wo" or "version_wo".
+	// To avoid user confusion, we omit these here as well.
+	tfbridge.MustTraverseProperties(&prov, "write-only-supporting-properties",
+		func(propertyVisitInfo tfbridge.PropertyVisitInfo) (tfbridge.PropertyVisitResult, error) {
+			schemaPath := propertyVisitInfo.SchemaPath()
+			if key, ok := schemaPath[len(schemaPath)-1].(walk.GetAttrStep); ok {
+				if strings.Contains(key.Name, "_wo_") || strings.HasSuffix(key.Name, "_wo") {
```
t0yv0, line 5730: I like this, this seems reasonable. If we are worried that this will over-eagerly skip a property that does not need skipping,  we can add a file with a checklist of removed properties, say (resource, SchemaPath) pairs and force PRs to change that, perhaps in a test.

patches/0085-Remove-ephemeral-resources.patch
```diff
@@ -1,1002 +0,0 @@
-From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
-From: guineveresaenger <guinevere@pulumi.com>
```
t0yv0, line 2: Nice.

t0yv0 (approved):
This LGTM.

Issue #5119, opened 2025-01-21
Nodejs AWS ECR lifecycle policy doesn't support tagPatternList parameter


joneteus:
### Describe what happened

Missing feature.

Nodejs SDK type `Selection` for ECR lifecycle policies is missing tagPatternList option. Tag pattern list allows to use wildcards in policy rules.

https://aws.amazon.com/about-aws/whats-new/2023/12/amazon-elastic-container-registry-wildcards-lifecycle-policies/

https://docs.aws.amazon.com/AmazonECR/latest/userguide/lifecycle_policy_parameters.html#lp_tag_pattern_list

https://github.com/pulumi/pulumi-aws/blob/2b29733b78a352d7b6cccf00b83f22fb434347fe/sdk/nodejs/ecr/lifecyclePolicyDocument.ts#L53C18-L53C27

Looks like it is implemented in Pulumi AWS dotnet and Java SDKs but not Nodejs.

### Sample program

Missing feature. Nothing to demonstrate.

### Output of `pulumi about`

Missing feature. Nothing to demonstrate.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

corymhall:
@joneteus thanks for raising this with us. It looks like the `LIfecyclePolicyDocument` type is a nodejs overlay type which means that it is manually maintained. If you look at the input type for other languages you will notice that the type is just `string` https://www.pulumi.com/registry/packages/aws/api-docs/ecr/lifecyclepolicy/#inputs

Alternatively you can use the [getLifecyclePolicyDocument](https://www.pulumi.com/registry/packages/aws/api-docs/ecr/getlifecyclepolicydocument/) function which does have the `tagPatternList` parameter.

We would happily take a contribution to add the `tagPatternList` parameter to the nodejs overlay type [here](https://github.com/pulumi/pulumi-aws/blob/2b29733b78a352d7b6cccf00b83f22fb434347fe/sdk/nodejs/ecr/lifecyclePolicyDocument.ts#L53C18-L53C27)

gnana997:
Hi @corymhall , Can I pick this issue?

PR #5365, opened 2025-03-21
Added tagPatternLists type to getLifecyclePolicyDocument in ecr

gnana997:
Added tagPatternLists parameter in getLifecyclePolicyDocument function.

Fixes: #5119


t0yv0:
Thank you! This looks right.

PR #5380, opened 2025-03-27
Fix route53.ResolverEndpoint hangs with 6-10 IPs

### LINKED ISSUE #5219: Pulumi hangs when creating route 53 resolver endpoint with more than 6 ip/subnets
### Describe what happened

When you create an inbound or outbound route 53 resolver endpoint with more than 6 ip/subnets - pulumi will hang, sometimes for up to an hour, and not provide any error message. Sometimes pulumi will even report the resource creation as successful even when it hasn't created anything.

AWS has service quotas for IP addresses for these resolver endpoints. See [docs](https://docs.aws.amazon.com/general/latest/gr/r53.html#limits_route53). These limits are:

Target IP addresses per resolver rule: 6 per region (hard limit, not adjustable)
IP addresses per resolver endpoint: 6 per region (soft limit, adjustable)

The Terraform SDK does not have any validation implemented for target IP addresses:
https://github.com/hashicorp/terraform-provider-aws/blob/982985456da51701cc7dc748c25fd78a82d3dec5/internal/service/route53resolver/rule.go#L88

The Terraform SDK implements a hard limit of 10 IP addresses per endpoint:
https://github.com/hashicorp/terraform-provider-aws/blob/982985456da51701cc7dc748c25fd78a82d3dec5/internal/service/route53resolver/endpoint.go#L66

The Terraform SDK is exhibiting similar behavior:
https://github.com/hashicorp/terraform-provider-aws/issues/40480

The limit exceeded error does not appear to be handled correctly.

Pulumi is inheriting this issue.

### Sample program

N/A

### Log output

_No response_

### Affected Resource(s)

_No response_

### Output of `pulumi about`

pulumi cli v3.124.0
python package pulumi-aws 6.67.0

### Additional context

_No response_

### Contributing

t0yv0:
Fixes #5219 

LimitExceededException is no longer retried but fails fast.

Added a patch tracking issue:

https://github.com/pulumi/pulumi-aws/issues/5385

Sent an upstream PR.

### DIFF
--- a/examples/examples_nodejs_test.go
+++ b/examples/examples_nodejs_test.go
@@ -1080,6 +1080,33 @@ func TestRegress4568(t *testing.T) {
 	})
 }
 
+func TestRegress5219(t *testing.T) {
+	skipIfShort(t)
+	dir := filepath.Join("test-programs", "regress-5219")
+	cwd, err := os.Getwd()
+	require.NoError(t, err)
+	providerName := "aws"
+	options := []opttest.Option{
+		opttest.LocalProviderPath(providerName, filepath.Join(cwd, "..", "bin")),
+		opttest.YarnLink("@pulumi/aws"),
+	}
+	test := pulumitest.NewPulumiTest(t, dir, options...)
+	ef := &expectFailure{T: t}
+	upResult := test.Up(ef)
+	t.Logf("%#v", upResult.Summary.ResourceChanges)
+	assert.Truef(t, ef.failed, "Expected pulumi up to fail")
+}
+
+type expectFailure struct {
+	*testing.T
+	failed bool
+}
+
+func (ef *expectFailure) FailNow() {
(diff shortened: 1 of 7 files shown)


examples/examples_nodejs_test.go
```diff
@@ -1080,6 +1080,33 @@ func TestRegress4568(t *testing.T) {
 	})
 }
 
+func TestRegress5219(t *testing.T) {
```
corymhall, line 1083: Is there a way to make sure it is failing fast? It might eventually error due to a timeout.

corymhall (approved):
We should probably add the service to the upstream tests just to be safe (you could also include your upstream test in the patch)

https://github.com/pulumi/pulumi-aws/blob/0c4b423ad93499dee8a4d8931aeede1c7eaee684/.github/workflows/aws-upstream-tests.yml#L46

PR #5430, opened 2025-04-14
Explore removing custom tagging logic for PF based resources

t0yv0:


t0yv0:
This looks promising, fighting the tests a bit. Noticing that we over-index the PF tags testing on AppConfig.Environment which is a little weird - if you send it an empty tag it treats it as null. Also Go AWS SDK APIs return this tag as null, e.g. tags are `map[string]*string`. We coerce nulls back to empty strings. This was an interesting rabbit hole but probably not exactly pertinent to test failures.

```
# aws_appconfig_environment.env:
resource "aws_appconfig_environment" "env" {
    application_id = "mk9szuf"
    arn            = "arn:aws:appconfig:us-west-2:616138583583:application/mk9szuf/environment/3gqa1i8"
    description    = null
    environment_id = "3gqa1i8"
    id             = "3gqa1i8:mk9szuf"
    name           = "anton-test-env"
    state          = "ReadyForDeployment"
    tags           = {
        "x" = null
    }
    tags_all       = {
        "x" = null
        "y" = "s"
    }
}
```

t0yv0:
The other bit I noticed is that I cannot pass appconfig.GetEnvironment asserts anymore. I would like to really understand what is the problem there. TF `refresh` and TF `import` work fine on a basic AWS TF provider. I wonder if there are remaining patches that impede Read() functionality for our patched AWS TF provider, or there is some eventual consistency, or something else is off.

t0yv0:
Hmm even on our patched provider, `terraform import` succeeds in pulling the null tag. Suspect this is something in our Go plumbing. To be continued.

examples/examples_go_test.go
```diff
@@ -282,27 +289,26 @@ func (st tagsState) validateStateResult(phase int) func(
...
+				// st.assertTagsEqualWithRetry(t,
+				// 	func() (map[string]string, error) {
+				// 		s := stack.Outputs["get-appconfig-env"].(string)
+				// 		var tags map[string]string
+				// 		if err := json.Unmarshal([]byte(s), &tags); err != nil {
+				// 			return nil, err
+				// 		}
+				// 		return tags, nil
+				// 	},
+				// 	"appconfig.GetEnvironment did not return the expected tags")
```
t0yv0, line 306: This is the problematic bit that possibly regressed, around R.get. We need to find out what's happening. The patched provider responds correctly to ReadResource under Terraform.

PR #5444, opened 2025-04-18
Simplify our patches by combining similar patches

corymhall:
This PR combines two groups of patches

1. The `tags_all` patches. Our current process adds an additional patch
   for each new resource that needs to add this standard patch. This PR
   combines all those individual patches into a single patch.
2. The `shim.go` patches. There were several patches that edit
   `shim.go`. I've combined those patches into a single patch that
   applies all the optimizations.

I validated that the patches do not diff by running `make upstream` in
this branch and the `master` branch and then runninng `git diff
./upstream ../master/upstream`

```
❯ git diff ./upstream ../master/upstream
:...skipping...
diff --git 1/./upstream/.git 2/../master/upstream/.git
index 20c3a3c86d0..84e75e65647 100644
--- 1/./upstream/.git
+++ 2/../master/upstream/.git
@@ -1 +1 @@
-gitdir: ../../worktrees/simplify-shim-patches/modules/upstream
+gitdir: ../../worktrees/master/modules/upstream
```


scripts/patch_computed_only.sh
```diff
@@ -32,3 +31,10 @@ for f in $FILES; do
     replace="s/names.AttrTagsAll:\s+tftags.TagsAttributeComputedOnly/names.AttrTagsAll: tftags.TagsAttribute/g"
     perl -i -p -e "$replace" "upstream/$f"
 done
+
+# Add the tags fixes to the existing commit that manages the tags_all attribute.
+cd upstream
+git add .
+tags_all_commit=$(git log --pretty=format:"%h" --grep="Apply tags patches" -n 1)
+git commit --fixup="$tags_all_commit"
```
corymhall, line 39: This allows us to add the changes to the existing patch instead of creating new patches for each individual resource.

t0yv0 (approved):
Nice, super helpful.

Issue #5168, opened 2025-02-05
Ensure resources are in the correct module


corymhall:
There are some resources where we have diverged from Terraform in how we organize the resources into modules. For example, Terraform has the EC2, Vpc, & Vpc IPAM modules and we have the EC2 & Vpc modules. Even then our resources are not organized into the EC2 and Vpc modules the same way that Terraform has organized them.

To make this easier on users we should make sure our organization matches how Terraform is organized. This will require a breaking change so should wait until the next major release.

t0yv0:
We might have an option to effect a non-breaking change via "source aliases" whereby old classes are retrained in SDK as-is but new classes are added at the correct location, and both classes point to the same underlying resource token (that's changing). It would increase SDK sizes. I think absent concrete interest in this I would wait until 7.0 indeed as you suggest.

PR #5475, opened 2025-04-30
Upgrade terraform-provider-aws to v5.96.0

guineveresaenger:
This PR was run manually via process described in CONTRIBUTING.md

Upgrades terraform-provider-aws to v5.96.0.

Fixes https://github.com/pulumi/pulumi-aws/issues/5455.

- **Update patches for v5.96.0**
- **Move upstream to v5.96.0**
- **make tfgen**
- **build sdks**


provider/resources.go
```diff
@@ -490,6 +490,7 @@ var moduleMap = map[string]string{
 	"transfer":                        transferMod,
 	"verifiedaccess":                  verifiedaccessMod,
 	"verifiedpermissions":             verifiedpermissionsMod,
+	"vpc_endpoint_associations":       "VpcEndpointAssociations",
```
corymhall, line 493: I think "vpc_endpoint_associations" should be part of the "vpc" mod. We'll probably have to add an entry like this

https://github.com/pulumi/pulumi-aws/blob/a31fb4d445d76d0c8ceab67e259ac87ef2cb4e46/provider/resources.go#L4476-L4478

We've decided we want to organize the resource based on how terraform organizes them in their docs
https://github.com/pulumi/pulumi-aws/issues/5168

corymhall (approved):
Just one update and then good to go!

t0yv0:
I'm picking up ops so let me address this, thank you!

provider/resources.go
t0yv0: It is part of the ec2 module upstream https://github.com/hashicorp/terraform-provider-aws/pull/41918/files probably most naturally belongs there. We have a little too much functionality in that module but still probably err on placing it there.

corymhall: I thought we had decided in #5168 that we should match how the Terraform docs organize things? For #5168 and v7 we need to decide what we base our organization on. For example, upstream doesn't even have a `vpc` folder.

corymhall: https://github.com/pulumi/pulumi-aws/pull/5161#discussion_r1942063652

t0yv0: Hmm, sorry I have missed that. https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc_endpoint.html is organized under VPC indeed. That sounds like a lot more intentional than where the source is located. Let me fix up.

corymhall (approved):

Issue #5483, opened 2025-05-02
Type 'DefaultSecurityGroup' is not assignable to type 'SecurityGroup'


oliparcol:
### Describe what happened

Hell,

I believe that #5442 breaks the compatibility between`DefaultSecurityGroup` and `SecurityGroup`:

```
Types of property 'egress' are incompatible.
        Type 'Output<DefaultSecurityGroupEgress[]>' is not assignable to type 'Output<SecurityGroupEgress[]>'.
          Type 'Output<DefaultSecurityGroupEgress[]>' is not assignable to type 'OutputInstance<SecurityGroupEgress[]>'.
            Types of property 'apply' are incompatible.
              Type '{ <U>(func: (t: DefaultSecurityGroupEgress[]) => Promise<U>): Output<U>; <U>(func: (t: DefaultSecurityGroupEgress[]) => OutputInstance<...>): Output<...>; <U>(func: (t: DefaultSecurityGroupEgress[]) => U): Output<...>; }' is not assignable to type '{ <U>(func: (t: SecurityGroupEgress[]) => Promise<U>): Output<U>; <U>(func: (t: SecurityGroupEgress[]) => OutputInstance<U>): Output<...>; <U>(func: (t: SecurityGroupEgress[]) => U): Output<...>; }'.
                Types of parameters 'func' and 'func' are incompatible.
                  Types of parameters 't' and 't' are incompatible.
                    Type 'DefaultSecurityGroupEgress[]' is not assignable to type 'SecurityGroupEgress[]'.
                      Type 'DefaultSecurityGroupEgress' is not assignable to type 'SecurityGroupEgress'.
                        Types of property 'description' are incompatible.
                          Type 'string | undefined' is not assignable to type 'string'.
                            Type 'undefined' is not assignable to type 'string'.
```

After the change mentioned above, `description` attribute of output rules is always set and cannot be `null`. I believe that `description` of ingress and egress rules of default security groups outputs should also be set to the empty string to fix the issue.

As a temporary fix, casting the default security group to `aws.ec2.SecurityGroup` solves the error: `defaultSecurityGroup as aws.ec2.SecurityGroup`.

Thanks

### Sample program

```
const defaultSecurityGroup = new aws.ec2.DefaultSecurityGroup(
  `defaultSecurityGroup`,
  {
    vpcId: vpc.id,
    ingress: [],
    egress: [],
  },
)

function testFunction(group: aws.ec2.SecurityGroup) {
    ...
}

[...]

PR #5484, opened 2025-05-02
Revert #5442

t0yv0:
Reverts #5442 

The root cause of https://github.com/pulumi/pulumi-aws/issues/5438 is now fixed in the bridge https://github.com/pulumi/pulumi-terraform-bridge/releases/tag/v3.107.0 and we no longer need the workaround.

This also will alleviate https://github.com/pulumi/pulumi-aws/issues/5483

We need to remember to update pulumi-eks with the version that ships this change to avoid tripping up this error again:

https://github.com/pulumi/pulumi-eks/issues/1752

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -2026,19 +2026,9 @@ compatibility shim in favor of the new "name" field.`)
 					// Use "ingress" instead of "ingresses" to match AWS APIs
 					"ingress": {
 						Name: "ingress",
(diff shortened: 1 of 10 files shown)


corymhall (approved):
Do we need a test?

t0yv0:
There is a test from https://github.com/pulumi/pulumi-aws/pull/5442 that is still passing.

PR #5505, opened 2025-05-08
[7.0.0-alpha]: remove tags related patches

corymhall:
There are two main breaking changes in this update.

1. If resource tags are removed _and_ the keys that are removed also
   exist in `DefaultTags` _with a value of ""_, then they will not be updated
   related to [#37297](https://github.com/hashicorp/terraform-provider-aws/issues/37297)
 - Created a follow up issue https://github.com/pulumi/pulumi-aws/issues/5517
2. `tags` and `tagsAll` are now separate things. Previously we set `tags`=`tagsAll` so `tags` always included _all_ the tags. Now only `tagsAll` will contain all tags.

I kept some of our customization on the Pulumi side to deal with a Pulumi only issue. For sdkv2 resources, if you are removing `defaultTags` then the `tagsAll` attribute will only be updated when:
   - `pulumi refresh --run-program` with the changes to defaultTags
   - `pulumi up` to propagate the changes


corymhall:

- #5479
    - #5505 ◀

corymhall:
> Have we considered a workaround for the issue with removing default tags for sdkv2 resources? Do we think that issue is not such a big deal?

My plan is to create a follow up issue for it. It's a pretty niche scenario that also affects Terraform so my opinion is that we could release with that issue. If we decide that we can't then I think it's also something we could fix, but I think we can fix it in a follow up PR.

provider/resources.go
```diff
@@ -5507,73 +5507,57 @@ compatibility shim in favor of the new "name" field.`)
...
-	}
-
-	prov.MustComputeTokens(tks.MappedModules("aws_", "", moduleMap,
-		func(mod, name string) (string, error) {
-			return awsResource(mod, name).String(), nil
-		}))
-
 	prov.P.ResourcesMap().Range(func(key string, res shim.Resource) bool {
+		// only process sdkv2 resources
+		if _, ok := up.ResourcesMap[key]; !ok {
```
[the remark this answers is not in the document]
corymhall: It's the upstream SDKV2Provider

https://github.com/pulumi/pulumi-aws/blob/fd7248450c1a7b7063e91eb697145c30b45c90bd/provider/resources.go#L858-L858

provider/resources.go
```diff
@@ -5507,73 +5507,57 @@ compatibility shim in favor of the new "name" field.`)
...
-		prov.Resources[k] = v
-	}
-
-	prov.MustComputeTokens(tks.MappedModules("aws_", "", moduleMap,
-		func(mod, name string) (string, error) {
-			return awsResource(mod, name).String(), nil
-		}))
-
 	prov.P.ResourcesMap().Range(func(key string, res shim.Resource) bool {
+		// only process sdkv2 resources
```
guineveresaenger, line 5511: It would be worth it, IMO, to add a bit more of the context from the PR description right here as a comment.I'm happy to have it be a followup to avoid wrestling with these massive PRs for small non-code changes.

guineveresaenger (approved):

provider/resources.go
corymhall: Good call, I'll add the comment.

PR #5508, opened 2025-05-09
Remove Region overlay to support removing dangling ref for Region

guineveresaenger:
This change aligns the Node SDK with the other providers.

Prior to this change, the schema has `"$ref": "#/types/aws:index/region:Region"` in the definition for `config.Region`. This allows the Node SDK to use an overlay that defines a Region type.

However, this ref does not match the Region enum type defined in `types.go`. While all SDKs have explicit Regions defined in `enums` as a result of `ExtraTypes`, the provider config for Go, Dotnet, etc, is currently only implemented as a basic string (probably because the type reference is invalid here and thus not used in codegen). 

By removing the (invalid for non-Node SDKs) type token ref , and removing the overlay file, this will result in config.Region being a string in the Node SDK as well.

It appears as though the Enum capability [does not currently work on provider config vars](https://github.com/pulumi/pulumi/issues/13435), but I could be misreading the situation. This would mean that the enum types are not being used at all by the provider type.

Alternatives attempted:

- Changing the type ref to "$ref": "#/types/aws:index/Region:Region" - this results in the loss of the Output for Region in the Node SDK
- Changing the type in types.go to be "aws:index/region:Region" - this has the same effect as uppercasing the type ref - the tokens now match and codegen doesn't handle the config type ref well:
```
/pulumi/pulumi-aws/sdk/dotnet/Provider.cs(397,22): error CS0029: Cannot implicitly convert type 'string' to 'Pulumi.Input<Pulumi.Aws.Region>'
```

I'm open to more suggestions in case I'm missing something here. It's obviously not ideal to move the Node SDK from a stronger-typed Region enum to a string, but it appears as though enums aren't supported completely in the provider config, and the only way to get strong types is to have an overlay; but having that overlay results in a dangling schema type.


corymhall (approved):
Looks good! Can you add a follow up item to remind us that this needs to be added to the upgrade guide? Or you can add it here https://github.com/pulumi/pulumi-aws/blob/7.0.0-alpha/docs/v7-upgrade.md

guineveresaenger:
It turns out that Node implicitly handles the type conversion for us here - it will pass the region enum as the underlying string , so:
```
const provider= new aws.Provider("prov", { region: aws.Region.USWest2 })
``` 
will still work as before; the difference is only in the type that `region` is.

Issue #5226, opened 2025-02-21
Support more than one AWS region with a single Provider


SamuelDudley:
Hi,
You didn't have a template for feature requests so I have used a blank issue.

## Background
I have been using pulumi for about 4 years now, managing quite large AWS deployments (50+ AWS accounts). I really appreciate the tool and the effort that has been spent sustaining and improving it.

I use pulumi to manage whole AWS organizations, the accounts in these organizations and the resources in these accounts. This means providers with multiple AWS accounts and regions all in the same pulumi stack managing thousands of resources.

As the number of AWS accounts and regions I am targeting has grown, the number of AWS and AWS native providers I instantiate has increased too. This is because the concept of a pulumi provider is designed to work in a single AWS account and in a single AWS region. These are defined when the provider is instantiated and can't be changed at runtime (as far as I am aware).

## The issue
Recently I began working on deploying AWS Config recorders, aggregators and delivery channels to all AWS regions (~16 regions are active by default and can't be opted out of) in my AWS accounts.

When a provider is instantiated by the pulumi runtime it creates a new process and this process consumes ~170MB of memory. This causes a situation where you can quickly run out of memory as the number of AWS accounts you are targeting with the stack increase: 

![Image](https://github.com/user-attachments/assets/bd287979-f61a-4058-b0f1-483bbb26f686)
![Image](https://github.com/user-attachments/assets/5e6abfbb-cd00-4223-857b-af84d6e6212d)

## Work around
Currently I am creating a dynamic resource which uses boto3 to manage the resources in other regions. While this works, it adds a lot of code and is throwing away all the benefits from using the underlying terraform provider to check and manage the resource state.

The boto3 approach works because it is far more memory efficient, allowing the creation of a `Session()` and then many clients in many regions from this single `Session`.  In pulumi, it appears the concept of a boto3 `Session` and `Client` (both with the same region) are represented by a pulumi Provider.

## Proposed feature
Create an interface similar to the `boto3` api, where a single authorised `Session()` allows the definition of a `Client()` that takes a region name as an argument. The client inherits the auth model from the session without the significant memory overhead of creating a new provider resource. 

Alternatively, continue to keep the 1 to 1 session to client pattern and greatly reduce the memory overhead of a Provider.

----
Would this be something the pulumi team would consider?
[...]

PR #5511, opened 2025-05-09
[7.0.0-alpha]: Upgrade upstream to v6.0.0-beta1

corymhall:
Upgrade upstream to v6.0.0-beta1

This PR upgrades upstream to the new [v6.0.0-beta1 tag](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v6.0.0-beta1)

The major update to the SDKs in this update is the addition of the new `region` resource level property

### How to review

1. commit [Update patches for](https://github.com/pulumi/pulumi-aws/pull/5511/commits/70da3eb527bb74ca95168bb09b2e5407aa863265)
Contains the changes to patches. Pretty minor changes
2. commit [this commit does a couple of things](https://github.com/pulumi/pulumi-aws/pull/5511/commits/32642a171285e45fdc6ef526cbb096e815314c79)
This is the bulk of the changes I made in this commit
3. commit [fix tests](https://github.com/pulumi/pulumi-aws/pull/5511/commits/e34bb4ee9f32cafc50c72e45a37b06bdf4cbc21d)
Contains all the updates I had to make to the tests. I also regenerated all of our upgrade tests so they are testing the upgrade from v6 => v7.

The SDK changes are all in a single commit. Pretty much every resource is updated, but almost all the changes are related to adding the region parameter. 

re #5226


corymhall:

- #5479
    - #5505
        - #5511 ◀

provider/diff_test.go
corymhall, line 1: Created follow up to re-enable.
https://github.com/pulumi/pulumi-aws/issues/5518

examples/examples_yaml_test.go
```diff
@@ -67,10 +67,6 @@ func TestAccPluginFramework(t *testing.T) {
 	integration.ProgramTest(t, &test)
 }
 
-func TestBucketUpgrade(t *testing.T) {
```
corymhall, line 70: We have much more robust bucket upgrade tests now. Removing this simple one.

examples/examples_yaml_test.go
```diff
@@ -394,85 +372,6 @@ func TestNonIdempotentSnsTopic(t *testing.T) {
 	require.ErrorContains(t, err, "already exists")
 }
 
-func TestOpenZfsFileSystemUpgrade(t *testing.T) {
```
corymhall, line 397: This test was specific to testing the removal of MaxItemsOne during a minor version update and testing we could go from a string to a list. This isn't relevant anymore now that we are going from v6 to v7.

examples/examples_yaml_test.go
```diff
@@ -1384,299 +1280,6 @@ outputs:
...
-// pulumi-aws v6.0.0 and state from pulumi-aws 5.42.0.
-//
-// The first test ensures we don't regress on https://github.com/pulumi/pulumi-aws/issues/2682
-//
-// The second test is when upgrading from pulumi-aws version <5.0.0 to v6.x.x, and
-// prevents regressions on https://github.com/pulumi/pulumi-aws/issues/2823
-//
-// Updated in https://github.com/pulumi/pulumi-aws/pull/3881
-// replacements.
-func TestMigrateRdsInstance(t *testing.T) {
```
corymhall, line 1397: Removing this since it tested an issue with going from v5 to v6. Not relevant anymore for v6 to v7.

patches/0029-Handle-cases-where-region-exists-and-is-nil.patch
```diff
@@ -0,0 +1,21 @@
...
+diff --git a/internal/provider/provider.go b/internal/provider/provider.go
+index 8512e5dfde..f4359bad72 100644
+--- a/internal/provider/provider.go
++++ b/internal/provider/provider.go
+@@ -533,7 +533,9 @@ func initialize(ctx context.Context, provider *schema.Provider) (map[string]conn
+ 
+ 					if isRegionOverrideEnabled && getAttribute != nil {
+ 						if region, ok := getAttribute(names.AttrRegion); ok {
+-							overrideRegion = region.(string)
++							if val, ok := region.(string); ok {
```
corymhall, line 16: TODO: once merged create tracking issue. Maybe a bridge issue.

patches/0016-Fail-fast-when-PF-resources-are-dropped.patch
```diff
@@ -1,22 +0,0 @@
...
-+++ b/internal/provider/fwprovider/provider.go
-@@ -461,9 +461,8 @@ func (p *fwprovider) Resources(ctx context.Context) []func() resource.Resource {
- 	}
- 
- 	if err := errors.Join(errs...); err != nil {
--		tflog.Warn(ctx, "registering resources", map[string]any{
--			"error": err.Error(),
--		})
-+		//lintignore:R009
-+		panic(err)
```
corymhall, line 19: This panic was moved into the enhancements-to-shim patch

patches/0029-Handle-cases-where-region-exists-and-is-nil.patch
guineveresaenger: What's `overrideRegion` prior to this line? The remaining code is in another patch, right?

examples/examples_nodejs_test.go
```diff
@@ -595,6 +595,9 @@ func TestRegress2868(t *testing.T) {
 	test := getJSBaseOptions(t).
 		With(integration.ProgramTestOptions{
 			Dir: filepath.Join(getCwd(t), "regress-2868"),
+			// TODO[pulumi/pulumi-aws#5521] `region` causes permanent diff without refresh
+			PreviewCommandlineFlags: []string{"--refresh"},
+			UpdateCommandlineFlags:  []string{"--refresh"},
```
guineveresaenger, line 600: Same question as above: does the PreCheckCallback not apply here?

guineveresaenger (approved):

patches/0029-Handle-cases-where-region-exists-and-is-nil.patch
corymhall: The rest of the code is just upstream code. The only patch we have in this code is what i've added here. 

The line above which isn't included in the patch just initializes overrideRegion. 

```go
var overrideRegion string
```

`getAttribute` is a function looks up the attribute in the rawState, essentially something like this:

```go
getAttribute := func(key: string): (any, bool) {
  v, ok := rawState[key]; return v, ok
}
```

In Terraform it looks like at the point this code is running `rawState` does not have a `region` attribute at all so `getAttribute` returns `false`. In Pulumi for some reason `rawState` does have a `region` attribute with a `nil` value. I think it's a bridge issue so I'll create a follow up issue.

PR #5519, opened 2025-05-12
Compare schema from v7 branch

guineveresaenger:
This should run schema check against the `7.0.0-alpha` branch for pull requests made against it. 

I don't think that we need to worry about central CI management clobbering this either.

### DIFF
--- a/.github/workflows/prerequisites.yml
+++ b/.github/workflows/prerequisites.yml
@@ -92,7 +92,7 @@ jobs:
         EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
         {
           echo "SCHEMA_CHANGES<<$EOF";
-          schema-tools compare -r github://api.github.com/pulumi -p aws -o "${{ inputs.default_branch }}" -n --local-path=provider/cmd/pulumi-resource-aws/schema.json;
+          schema-tools compare -r github://api.github.com/pulumi -p aws -o 7.0.0-alpha -n --local-path=provider/cmd/pulumi-resource-aws/schema.json;
           echo "$EOF";
         } >> "$GITHUB_ENV"
     - if: inputs.is_pr && inputs.is_automated == false && github.actor != 'dependabot[bot]'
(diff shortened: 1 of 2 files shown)


.github/workflows/prerequisites.yml
```diff
@@ -92,7 +92,7 @@ jobs:
         EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
         {
           echo "SCHEMA_CHANGES<<$EOF";
-          schema-tools compare -r github://api.github.com/pulumi -p aws -o "${{ inputs.default_branch }}" -n --local-path=provider/cmd/pulumi-resource-aws/schema.json;
+          schema-tools compare -r github://api.github.com/pulumi -p aws -o 7.0.0-alpha -n --local-path=provider/cmd/pulumi-resource-aws/schema.json;
```
corymhall, line 95: We might be able to have it always reference the base branch of the PR. This would also be useful in cases of stacked PRs where we would only see the changes relevant to the specific stacked PR.

Instead of passing in the `default_branch` we could use `${{ github.event.base.ref }}`
https://github.com/pulumi/pulumi-aws/blob/a3ae9633b0203acc9e2e159b0da876f447875bfb/.github/workflows/run-acceptance-tests.yml#L50

pull_request event deliveries for reference https://github.com/pulumi/pulumi-aws/settings/hooks/430483297?tab=deliveries

If it works this would be something we could probably upstream to ci-mgmt

corymhall: > But the initial schema change comment on this PR is promising!

🎊

corymhall: Oh maybe it is `${{ github.event.pull_request.base.ref }}`

guineveresaenger: Jinx! looks like that's it! I think we can merge this since ci-mgmt won't clobber the feature branch.

Will do a follow-up there for your more general suggestion.

corymhall (approved):

PR #5575, opened 2025-06-02
upstream v5.99.0

t0yv0:
Upgrades upstream to v5.99.0

Fixes https://github.com/pulumi/pulumi-aws/issues/5569


t0yv0:
Shipped included with https://github.com/pulumi/pulumi-aws/pull/5576

PR #5615, opened 2025-06-18
[v7]: Add back resource references

corymhall:
This adds back the resource references and the minimal schema stuff. We
have decided to keep dangling references for now since we have a path
towards fixing this soon.

re https://github.com/pulumi/pulumi/issues/19860
re https://github.com/pulumi/pulumi-terraform-bridge/issues/3069


mjeffryes (approved):
Thanks for doing this (and raising the breaking change concern to make sure we shipped the right thing for users!) Hopefully we'll get to rip out the minimal schema for good in the near future!

PR #5629, opened 2025-06-23
[v7]: Re-export iam PolicyDocument types

corymhall:
Re-exporting the `PolicyDocument` and `PolicyStatement` types. This
avoids an unnecessary breaking change for users.

### DIFF
--- a/examples/iam-policy-document/ts/index.ts
+++ b/examples/iam-policy-document/ts/index.ts
@@ -20,24 +20,38 @@ import * as aws from '@pulumi/aws'
 const account = aws.getCallerIdentityOutput().accountId;
 const bucket = new aws.s3.Bucket('test-bucket');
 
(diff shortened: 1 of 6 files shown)


examples/iam-policy-document/ts/index.ts
```diff
@@ -20,43 +20,47 @@ import * as aws from '@pulumi/aws'
 const account = aws.getCallerIdentityOutput().accountId;
 const bucket = new aws.s3.Bucket('test-bucket');
 
+const assumeRolePolicyDoc: aws.iam.PolicyDocument = {
+    Version: '2012-10-17',
```
mjeffryes, line 24: Why the changes to the examples?

mjeffryes (approved):

examples/iam-policy-document/ts/index.ts
corymhall: Just to validate that it worked.

PR #5641, opened 2025-06-26
[v7]: Remove ssm parameter tier patch

### LINKED ISSUE #2731: PATCH Workaround-SSM-Parameter-tier-bug.patch
Tracking context on upstream patch 0007-Workaround-SSM-Parameter-tier-bug.patch here.

### LINKED ISSUE #5595: SSM-Parameter Tier Intelligent-Tiering is not working, Standard tier is forced by default
### Describe what happened

Hi,

we just encountered two issues with [the patch](https://github.com/pulumi/pulumi-aws/blob/master/patches/0006-Workaround-SSM-Parameter-tier-bug.patch) mentioned in #2731 with an AWS account containing 10_000 Standard parameters.

There are two problems:

### Default is forced to Standard

https://github.com/pulumi/pulumi-aws/blob/92c35019e316f0007ceae85e68475adb999d83d0/patches/0006-Workaround-SSM-Parameter-tier-bug.patch#L23

This is a problem as you're able to configure Intelligent-Tiering as default for all parameters on AWS account level, which can be configured as documented [here](https://docs.aws.amazon.com/systems-manager/latest/userguide/parameter-store-advanced-parameters.html#parameter-store-advanced-parameters-enabling).

The account level default is not used now blocking us from creating new parameters, we have to specify Tier: Advanced everywhere explicitly.

### Intelligent-Tiering is not working

The other issue is that Intelligent-Tiering as tier is not working at all for us, we have to set Advanced explicitely so that the parameters are not created in Standard tier (which fails due to the 10_000 parameter limit).
I assume as this is due to the change to Intelligent Tiering being suppressed here, even for new resources:

https://github.com/pulumi/pulumi-aws/blob/92c35019e316f0007ceae85e68475adb999d83d0/patches/0006-Workaround-SSM-Parameter-tier-bug.patch#L30

I setup a new terraform project and did not see any of those issues, parameters were created as expected.
As this page is almost two years old ([see initial commit](https://github.com/pulumi/pulumi-aws/commit/e466d931de1658cb9666e2679ecd3f0005cccd17#diff-1cf5587d95e08b0d788c5110e0551ab206ae9d8a9ba7e1d35e2c9e455403c7d8)), I would guess that the remark in this patch about an issue with the terraform provider is not really up to date anymore.

Let me know how I can contribute.

Thanks, Vincent

### Sample program

1. Have an AWS account with 10_000 Standard SSM parameters which is the limit
2. Optional: Configure your AWS SSM account level setting to use Intelligent-Tiering by default, see [here](https://docs.aws.amazon.com/systems-manager/latest/userguide/parameter-store-advanced-parameters.html#parameter-store-advanced-parameters-enabling)
3. Create a plain ssm.Parameter without a tier or with tier `Intelligent-Tiering` - both are not working

### Log output

```console
        * creating SSM Parameter (/vincent/ssm/test): operation error SSM: PutParameter, https response error StatusCode: 400, RequestID: afd29b27-67ae-4d74-bde0-6c261f056cca, ParameterLimi
tExceeded: You have reached the maximum number of standard parameters for this AWS account and Region (10000). To add more parameters, either delete existing parameters or upgrade to the ad
vanced-parameter tier (charges apply).
```

### Affected Resource(s)

_No response_

### Output of `pulumi about`

```console
❯ pulumi about

CLI          
Version      3.175.0
Go Version   go1.24.3
Go Compiler  gc

Plugins
KIND      NAME    VERSION
language  nodejs  3.175.0

Host     
OS       redhat
Version  8.10
Arch     x86_64

This project is written in nodejs: executable='/home/linuxbrew/.linuxbrew/bin/node' version='v24.2.0'

Current Stack: organization/ssm-debug/tst

TYPE                  URN
pulumi:pulumi:Stack   urn:pulumi:tst::ssm-debug::pulumi:pulumi:Stack::ssm-debug-tst
pulumi:providers:aws  urn:pulumi:tst::ssm-debug::pulumi:providers:aws::default_6_82_1
pulumi:providers:aws  urn:pulumi:tst::ssm-debug::pulumi:providers:aws::default_6_74_0


Found no pending operations associated with tst
...
```

### Additional context

_No response_

### Contributing

corymhall:
fixes #5595
fixes #2731


corymhall:
> Can we add a description to why it is now safe to remove this patch?
It's not clear from the descriptions in the linked issue what the underlying issue was that prompted the patch in the first place.

I believe this comment from the linked ticket describes the issue https://github.com/pulumi/pulumi-aws/issues/2731#issuecomment-2121192951. For any of the patches the only way we know for sure whether it is safe to remove is to add tests that validate it works without the patch. I've added those tests in this PR.

> Additionally - does this constitute a breaking change for existing users? if not, then perhaps this change should go into the default branch, not the release branch?

I don't know for sure. I don't _think_ it would be a breaking change, but there could be some cases where it is. For example, in the linked issue it mentions cases where you set a default value at the AWS account level. We can't test this case ourselves though. 

Also at this point our next release from the main branch will probably be v7 so I thought it would just be easier and safer to do it here.

PR #5661, opened 2025-07-17
[v7]: cluster certificateAuthority only returns a single item

corymhall:
Follow up to #5655. The upstream API only returns a single item so we
should remove the list version not the single item version


guineveresaenger (changes requested):
We'll have to edit the migration guide too, but otherwise this LGTM.

PR #5681, opened 2025-07-23
Update upgrade guide

corymhall:
adding additional instructions to the guide.


guineveresaenger (approved):
🎉 tese are so beautiful, thank you Cory!

Issue #5715, opened 2025-08-01
v6 -> v7  Migration leads to `defaultAddonsToRemove` no longer existing


rshade:
### Describe what happened

When migration from v6->v7 the eks input `defaultAddonsToRemove` has been removed. We currently use this heavily to manage EKS clusters and its blocking migration. The field `bootstrapSelfManagedAddons`, won't work because it is an all or nothing, and will lead to collisions with end user component usage. 

### Sample program

Not Needed

### Output of `pulumi about`

pulumi-aws-7

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

Issue #4755, opened 2024-11-12
Deprecate EKS cluster property `defaultAddonsToRemoves` in favor of `bootstrapSelfManagedAddons`


flostadler:
The `defaultAddonsToRemove` property was introduced to EKS cluster configuration to allow bootstrapping clusters without AWS's default components (`coredns`, `vpc-cni`, `kube-proxy`). This was implemented in a patch, through a workaround where the provider would:
1. Create an EKS addon on cluster creation
2. Delete it immediately after
3. This effectively brought the cluster components under Pulumi's management before destroying them

## Current Situation
AWS has now introduced native support for this functionality through the `bootstrapSelfManagedAddons` flag, which creates a blank cluster without default components.

Key differences:
- `defaultAddonsToRemove`: Allows selective removal of specific addons
- `bootstrapSelfManagedAddons`: All-or-nothing approach (creates a blank cluster). Is faster as it doesn't have to first install the components, adopt them and finally destroy them

## Proposed Changes
1. Mark `defaultAddonsToRemove` as deprecated
2. Remove the property in the next major version

## Required Investigation
Before proceeding with deprecation:
1. Analyze current usage patterns of `defaultAddonsToRemove`
   - Identify how users are utilizing the selective removal capability
   - Determine if there are common patterns that would be broken by moving to an all-or-nothing approach
2. Evaluate potential migration paths for users who need selective addon removal
3. Consider if we need to maintain both options or if AWS's solution sufficiently covers all use cases

## References
- Original implementation: [Patch #0028](https://github.com/pulumi/pulumi-aws/blob/c0d3ae1a2688e1eb17bf173fcd5f2a65525e074e/patches/0028-aws_eks_cluster-implement-default_addons_to_remove.patch)
- [AWS Announcement](https://aws.amazon.com/about-aws/whats-new/2024/06/amazon-eks-cluster-creation-flexibility-networking-add-ons/)

t0yv0:
There was some pushback on https://github.com/pulumi/pulumi-aws/issues/3009 before but I think what was missing is allowing users to still not get the addons wihch is now made available with `bootstrapSelfManagedAddons`. Thanks for filing!

PR #5716, opened 2025-08-01
Add back eks defaultAddonsToRemove property

corymhall:
Adding back the patch for `defaultAddonsToRemove` by popular demand.

fixes https://github.com/pulumi/pulumi-aws/issues/5715
re https://github.com/pulumi/pulumi-aws/issues/4755

### DIFF
--- a/examples/eks-remove-addons/index.ts
+++ b/examples/eks-remove-addons/index.ts
@@ -1,0 +1,91 @@
+// Copyright 2016-2018, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


examples/eks-remove-addons/index.ts
```diff
@@ -0,0 +1,105 @@
...
+
+const cluster = new aws.eks.Cluster('test-cluster', {
+    roleArn: clusterRole.arn,
+    vpcConfig: {
+        subnetIds: [subnet1.id, subnet2.id],
+    },
+    bootstrapSelfManagedAddons: false,
+});
+
+// TODO: example needs node group
```
mjeffryes, line 72: For sake of example, we could pick one of the addons that doesn't require a nodegroup?

mjeffryes (approved):
Thanks for spinning this up quickly. I see you're also adding an example for `bootstrapSelfManagedAddons`. Do we have any tests that exercise `defaultAddonsToRemove`?

examples/eks-remove-addons/index.ts
corymhall: > Thanks for spinning this up quickly. I see you're also adding an example for bootstrapSelfManagedAddons. Do we have any tests that exercise defaultAddonsToRemove?

This test also adds a cluster with `defaultAddonsToRemove`.

> For sake of example, we could pick one of the addons that doesn't require a nodegroup?

I think all addons need a node group to launch in

PR #5725, opened 2025-08-06
Fix assumeRole validation in validateCredentials

corymhall:
This fixes the validation around the `aws:assumeRoles` config. We were
validating the old `assumeRole` config instead of the new `assumeRoles`.

This also adds some additional validation so that we throw some nice
errors rather than panics.

### DIFF
--- a/examples/examples_nodejs_test.go
+++ b/examples/examples_nodejs_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/pulumi/providertest/pulumitest/opttest"
 	"github.com/pulumi/pulumi-aws/provider/v7/pkg/elb"
 	"github.com/pulumi/pulumi/pkg/v3/testing/integration"
+	"github.com/pulumi/pulumi/sdk/v3/go/auto"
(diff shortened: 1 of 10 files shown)


examples/examples_nodejs_test.go
```diff
@@ -1345,6 +1346,62 @@ func TestUpstreamWarningsPropagated(t *testing.T) {
...
+	newStack.SetConfigWithOptions(test.Context(), "aws:assumeRoles[0].roleArn",
+		auto.ConfigValue{Value: roleArn},
+		&auto.ConfigOptions{Path: true},
+	)
+
+	_, err = newStack.Up(test.Context())
+	assert.NoError(t, err)
+
+	// destroy this stack first since it uses the role created in the first stack
+	newStack.Destroy(test.Context())
```
mjeffryes, line 1402: It doesn't look like we destroy the first stack at all? or does pulumitest automatically clean it up for you anyway.

mjeffryes (approved):

PR #5905, opened 2025-10-16
Update the APN 1.1 Marketplace ID used on User Agent headers.

pose:
Closed in favor of https://github.com/pulumi/pulumi-aws/pull/5910 since it was getting messy to review.

---

Part of pulumi/home#4348

---

* ~~Currently, this is not working. This change is appending to the User Agent string, not replacing: `aws-sdk-go-v2/1.39.2 ua/2.1 os/macos lang/go#1.24.6 md/GOOS#darwin md/GOARCH#arm64 api/sts#1.38.6 m/t,u APN/1.1 (c7qiae2l6usvzoynupds6v7hf)`~~ This is now (2025-10-20) working correctly and replacing the whole User Agent string.

### DIFF
--- a/patches/0016-user-agent-set-to-Pulumi-apn-1.1-info.patch
+++ b/patches/0016-user-agent-set-to-Pulumi-apn-1.1-info.patch
@@ -28,7 +28,6 @@ index 0000000000..ddf3aed4ba
 +
 +func UserAgentHeaderRoundTripper(next http.RoundTripper) http.RoundTripper {
 +	return roundTripperFunc(func(r *http.Request) (*http.Response, error) {
-+		fmt.Println("Adding header")
 +		r.Header.Set("User-Agent", "APN/1.1 (c7qiae2l6usvzoynupds6v7hf)")
 +		resp, err := next.RoundTrip(r)
 +		return resp, err


patches/0006-Revert-WAF-schema-changes.patch
```diff
@@ -6,7 +6,7 @@ Subject: [PATCH] Revert WAF schema changes
 - This causes far too many types to be generated downstream.
```
pose: Is this normal? I didn't change anything here.

patches/0016-user-agent-set-to-Pulumi-apn-1.1-info.patch
```diff
@@ -0,0 +1,73 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: corymhall <43035978+corymhall@users.noreply.github.com>
```
pose: I would be happy change to me or add me as Co-Author (once I learn how). Thoughts?

provider/provider_endpoint_test.go
```diff
@@ -0,0 +1,93 @@
...
+			default:
+				t.Fatalf("Unexpected request count: %d", requestCount)
+			}
+			requestCount++
+		}))
+		t.Cleanup(server.Close)
+
+		// Using AWS_ENDPOINT_URL environment variable as a workaround.
+		// TODO Investigate why using aws:endpoints is not working.
+		t.Setenv("AWS_ENDPOINT_URL", server.URL)
```
pose: Somehow, `aws:endpoints` was always being ignored. I discussed with @Zaid-Ajaj but couldn't find a way to make it work so I decided to opt for this Terraform env var.

patches/0016-user-agent-set-to-Pulumi-apn-1.1-info.patch
corymhall: How did you generate the patch? I think it should have just picked up your git info

patches/0016-user-agent-set-to-Pulumi-apn-1.1-info.patch
```diff
@@ -0,0 +1,73 @@
...
+Date: Thu, 11 Apr 2024 08:06:27 -0400
+Subject: [PATCH] user agent set to Pulumi apn 1.1 info
```
corymhall: When you create the commit you can add a commit message with details as well. Can you add additional details to the commit so that in the future we can see all the context in the patch?

e.g.

- Why do we need to do this?
- Where did you get the APN id?
- ...

patches/0006-Revert-WAF-schema-changes.patch
blampe: I think GitHub's syntax highlighting makes these patches really difficult to look at. It's highlighting it as red, but you didn't actually change anything... the patch just happens to remove something. So you can ignore the stuff that doesn't also have the full-line-width background red color.

patches/0016-user-agent-set-to-Pulumi-apn-1.1-info.patch
pose: I intended to replace the patch with the new code so we don't have two patches for APN setting.

pose: What I did was:

1. `./scripts/upstream.sh rebase -i`
2. Placed `edit` next to commit pointed byt `patches/0016-user-agent-set-to-Pulumi-apn...`
3. Added files and made edits.
4. Staged my changes and `git commit --amend`
5. `./scripts/upstream.sh check_in`

I believe those steps are equivalent with what you described on _What i've done in the past..._.

patches/0016-user-agent-set-to-Pulumi-apn-1.1-info.patch
```diff
@@ -0,0 +1,73 @@
...
++	if rt == nil {
++		return http.DefaultTransport.RoundTrip(r)
++	}
++	return rt(r)
++}
++
++func UserAgentHeaderRoundTripper(next http.RoundTripper) http.RoundTripper {
++	return roundTripperFunc(func(r *http.Request) (*http.Response, error) {
++		fmt.Println("Adding header")
++		r.Header.Set("User-Agent", "APN/1.1 (c7qiae2l6usvzoynupds6v7hf)")
```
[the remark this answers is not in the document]
pose: Discussed offline. We will go back to the first approach. @blampe we can sync up during standup as @corymhall raised valid concerns in terms of how brittle this change would be.

Issue #5878, opened 2025-10-07
Re-introduce write-only supporting fields that were skipped alongside write-only fields


guineveresaenger:
Currently this provider suppresses fields such as `passworkd_wo_version` becuase they're nonsensical in the absence of write-only fields. We should remove this logic and allow all such fields to persist.

PR #5906, opened 2025-10-16
Re-introduce write-only version support fields

guineveresaenger:
This pull request removes logic that removed write-only supportive fields from the schema and SDKs. 

Once https://github.com/pulumi/pulumi-terraform-bridge/pull/3215 has been released and a new bridge version has landed in this provider, write-only fields will be available, and these supportive fields are necessary to use them properly.

Until the bridge changes land, fields such as `masterPasswordWoVersion` will not be usable, but they're necessary for the use of `masterPasswordWo` so they should be available before write-only fields are enabled.

Fixes https://github.com/pulumi/pulumi-aws/issues/5878
- **Remove logic to keep writeOnly supporting fields from being generated**
- **Make schema**
- **Build SDKs**

### DIFF
--- a/sdk/dotnet/Bedrock/AgentcoreApiKeyCredentialProvider.cs
+++ b/sdk/dotnet/Bedrock/AgentcoreApiKeyCredentialProvider.cs
@@ -59,6 +59,12 @@ namespace Pulumi.Aws.Bedrock
         public Output<ImmutableArray<Outputs.AgentcoreApiKeyCredentialProviderApiKeySecretArn>> ApiKeySecretArns { get; private set; } = null!;
 
         /// <summary>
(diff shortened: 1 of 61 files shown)


guineveresaenger:
> Shouldn't they be part of the same release? I think we should hold off on merging this until we merge in the bridge upgrade so that we make sure we don't do an aws release with this and without the bridge upgrade.

Ideally it would be part of the same upgrade, yes. But it's harmless to have these fields available, while if we release and ship the bridge _first_, we have the opposite problem - and it would be more potentially harmful to ship write-only fields without the means to control them. They might still work the first time - and then the user has no way of updating them until the next AWS update.

The ideal order here is:
1. Merge this change
2. Update the bridge
3. Release the provider

If 3) happens before 2) it is much safer than if 1) happens after 3).

PR #5940, opened 2025-10-31
fix: ordered list should start at 1

stooj:
Quick fix to sort an ordered list. It looks like there are missing steps, but the steps were removed in https://github.com/pulumi/pulumi-aws/commit/6bc468cca67e8d4c7424d6d994410c65ed141507 because they're covered by the previous section (the list of resources to rename).


ringods:
@corymhall is the `docs` folder not copied over to the registry at every release of the provider?

corymhall:
> is the `docs` folder not copied over to the registry at every release of the provider?

No I don't think it is, I've had to copy the content over manually.

stooj:
OK, also opened https://github.com/pulumi/registry/pull/9101

PR #5961, opened 2025-11-13
chore: update Go SDK references from v6 to v7

dirien:


pose:
Thanks @dirien for raising this. Seems the SDKs need to be re-generated to reflect that code change (that's why I believe the Python tests are failing). Let @pulumi/providers know if you need help with that. Thanks!

guineveresaenger (approved):

PR #6051, opened 2025-12-11
Extend schema to support java25 lambda runtime

bhurlow:
Not entirely sure but looks like the respective language runtimes will be auto generated?


corymhall:
Thanks for the contribution.

Closing this PR because support for java25 Lambda runtime is already included in a released provider version, so this change is no longer needed on top of current master.

If you still see a gap in the latest release, please open a new issue with a minimal repro and the exact provider version.

PR #6080, opened 2025-12-23
feat: add NodeJS24dX

adrians5j:
Adding `NodeJS24dX` to `Runtime` enum.

### DIFF
--- a/sdk/nodejs/types/enums/lambda/index.ts
+++ b/sdk/nodejs/types/enums/lambda/index.ts
@@ -12,6 +12,7 @@ export const Runtime = {
     NodeJS18dX: "nodejs18.x",
     NodeJS20dX: "nodejs20.x",
     NodeJS22dX: "nodejs22.x",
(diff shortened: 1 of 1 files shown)


adrians5j:
I see the warnings re auto generation, but still... 

Node 24 for Lambda was released +month  ago, and this file was not regenerated, which tells me it maybe still needs to be edited by hand?

In any case, would appreciate any extra info in this regard. :)

clara-btf:
Any updates on this? On my team we’re migrating our Lambdas to Node 24 and we need the library to support this enum 👀

adrians5j:
Ultimately we went with the plain string here. All works.

corymhall:
Thanks for the contribution.

Closing this PR because support for nodejs24.x is already included in a released provider version, so this PR is now superseded.

If the latest release is still missing expected behavior, please open a new issue with a repro and provider version details.

Issue #6115, opened 2026-01-14
[chore] re-enable `TestAccSQSQueue_List_basic`


pgavlin:
This test is broken as of v6.28.0 and should be reenabled.

PR #6116, opened 2026-01-14
[chore] disable TestAccSQSQueue_List_basic

pgavlin:
This test is broken as of upstream@v6.28.0.
https://github.com/pulumi/pulumi-aws/issues/6115 tracks re-enabling the test.


pgavlin:
This should fix https://github.com/pulumi/pulumi-aws/issues/6105.

pose (approved):

PR #6267, opened 2026-04-08
Don't nest the Region type

Frassle:
Core doesn't want to support modules nested under the "index" module. It makes little sense, and this appears to be the only instance of a type trying to do it. 

This needs updating so we can tighten the token validation in the CLI to dis-allow patterns like this without breaking aws-native builds.


Frassle:
oh wait, this is an "aws" type not "aws-native".... I need to think about this more 😕

Frassle:
Ignore me, I was in totally the wrong repo

PR #6524, opened 2026-07-09
Workaround upstream bug: aws_route53domains_delegation_signer_record dropped from state on refresh

### LINKED ISSUE #6526: Track upstream aws_route53domains_delegation_signer_record - Read/refresh removes resource from state even when it exists in AWS
### Describe what happened

Track patch and fix for upstream Terraform provider bug: https://github.com/hashicorp/terraform-provider-aws/issues/47928.

### Sample program

See PR

### Log output

_No response_

### Affected Resource(s)

_No response_

### Output of `pulumi about`

N/A

### Additional context

_No response_

### Contributing

smithrobs:
## Summary

Customer reported that `aws_route53domains_delegation_signer_record` is affected by an upstream Terraform provider bug: [hashicorp/terraform-provider-aws#47928](https://github.com/hashicorp/terraform-provider-aws/issues/47928).

**Symptom:** Immediately after a successful `Create`, the very next `refresh`/`plan` removes the resource from state. The following `apply` then fails with:
```
unexpected state 'FAILED', wanted target 'SUCCESSFUL'. last error: [Parameters in request are not valid]
```
because Terraform/Pulumi tries to re-create a DS record the registry already has.

**Root cause:** `Create` stores `DnssecKey.Id` (format `DS:keytag-algorithm-digesttype-digest`) as returned by `AssociateDelegationSignerToDomain`, but `Read` looks the key back up via `GetDomainDetail`, which does not reliably echo that `Id` back in the same form. The lookup by `Id` in `findDNSSECKeyByTwoPartKey` therefore never matches, and the resource is treated as gone.

## Fix

This adds a new upstream patch (`patches/0028-...patch`) to `internal/service/route53domains/delegation_signer_record.go`, based on the most complete upstream fix candidate ([hashicorp/terraform-provider-aws#48165](https://github.com/hashicorp/terraform-provider-aws/pull/48165)):

- **Key on `DnssecKey.Digest` instead of `Id`** in Create/Read/Delete/`findDNSSECKeyByTwoPartKey`. `Digest` is the value that is stable across `AssociateDelegationSignerToDomain` and `GetDomainDetail`.
- **Stop overwriting `signing_attributes` in Read.** `GetDomainDetail` doesn't populate `Algorithm`/`Flags`/`PublicKey` on `DnssecKey`, so the previous code was zeroing out the configured values and producing a permanent `RequiresReplace` diff.
- **Delete now resolves the internal `DnssecKey.Id`** from the digest before calling `DisassociateDelegationSignerFromDomain`, since that API still requires the internal `Id`, not the digest.
- **Adds a schema v0→v1 state upgrader** that rewrites any existing `dnssec_key_id`/`id` values from the old `DS:keytag-algorithm-digesttype-digest` shape down to the bare digest, so existing Pulumi stacks with state from a previously-broken provider aren't dropped on their first refresh after this fix lands. The upgrader is idempotent and leaves malformed values untouched rather than corrupting them.
- Adds a `PlanOnly` acceptance test step asserting an empty plan immediately after `Create` — the exact regression this fixes — plus a unit test for the digest-parsing helper used by the state upgrader.

## Testing

- `go build ./internal/service/route53domains/...` — clean
- `go vet ./internal/service/route53domains/...` — clean
- `go build ./...` (full upstream module) — clean
- `go test ./internal/service/route53domains/... -v` — all unit tests pass (acceptance tests skip without `TF_ACC`/AWS credentials, as expected)
- Verified `./scripts/upstream.sh init -f` re-applies all 28 patches (including the new one) cleanly from a clean checkout.

No SDK/schema regeneration was needed since the resource's public attribute set (`dnssec_key_id`, `domain_name`, `signing_attributes`, `id`) is unchanged — only the internal Create/Read/Delete implementation and schema version changed.

Fixes #6526

---
*Created with [Eon](https://eon.corp.pulumi.com/session/aa1cfcdd7ee105dc82130a45beae00c0)*

### DIFF
--- a/patches/0028-Fix-aws_route53domains_delegation_signer_record-drop.patch
+++ b/patches/0028-Fix-aws_route53domains_delegation_signer_record-drop.patch
@@ -1,0 +1,382 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Eon <eon@pulumi.com>
+Date: Thu, 9 Jul 2026 14:35:52 +0000
+Subject: [PATCH] Fix aws_route53domains_delegation_signer_record dropping from
+ state on refresh
+
+The Read function keyed on DnssecKey.Id, which GetDomainDetail does not
+reliably echo back in the form the resource stored it (the format returned
+by AssociateDelegationSignerToDomain), causing every refresh to fail to
+match the existing DS record and remove the resource from state. A
+subsequent apply would then fail because the registry already has the
+record.
+
+- Key Create/Read/Delete off DnssecKey.Digest, the value that is stable
+  across AssociateDelegationSignerToDomain and GetDomainDetail.
+- Stop flattening GetDomainDetail's DnssecKey onto signing_attributes in
+  Read: GetDomainDetail does not populate Algorithm/Flags/PublicKey on
+  DnssecKey, so the previous code was overwriting configured values with
+  zeros and producing a permanent RequiresReplace diff.
+- Delete now resolves the internal DnssecKey.Id from the digest before
+  calling DisassociateDelegationSignerFromDomain, which requires the
+  internal Id rather than the digest.
+- Bump schema to v1 with a state upgrader that rewrites existing
+  'DS:keytag-algorithm-digesttype-digest' shaped dnssec_key_id/id values
+  down to the bare digest, so users who already have state from a broken
+  provider version aren't affected on their next refresh.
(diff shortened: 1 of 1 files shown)


corymhall (changes requested):
Marking as request changes so it doesn't merge. We are discussing internally.

choeflake:
> Marking as request changes so it doesn't merge. We are discussing internally.

Hi, any progress on this issue?

PR #6547, opened 2026-07-17
fix: Fix aws_route timeouts-only update fix from upstream PR #47413

### LINKED ISSUE #6549: aws_route (aws:ec2/route:Route) update fails with "route target attribute not specified" when only timeouts changes
### What happened?

`aws.ec2.Route` fails to update with an error like:

    route target attribute not specified

when the only diff on the resource is to the `timeouts` block (e.g. changing `create`/`delete` timeout values, or a diff that surfaces spuriously after a provider upgrade). No route target attribute (`gatewayId`, `natGatewayId`, `transitGatewayId`, etc.) actually changed.

### Root cause

`resourceRouteUpdate` in the upstream provider always calls `routeTargetAttribute`, which requires at least one target attribute to have `HasChange` return `true`. When only `timeouts` differs, none of the target attributes have changed, so this call errors instead of being a no-op.

pierskarsenbarg:
## Summary
- Adds patch 0028 backporting hashicorp/terraform-provider-aws#47413, fixing `aws_route` update failures with "route target attribute not specified" when only the `timeouts` block changes (or when a provider upgrade produces a timeouts-only diff).
- Adds an early return in `resourceRouteUpdate` when no route target attributes have changed, avoiding an unnecessary/erroring API call.
- Includes the upstream acceptance test `TestAccVPCRoute_timeoutsOnlyChange` and its supporting config helper.

## Test plan
- [ ] Run `./scripts/upstream.sh init -f` and confirm the patch applies cleanly
- [ ] Run acceptance tests for `TestAccVPCRoute_timeoutsOnlyChange` / `TestAccVPCRoute_basic` in `internal/service/ec2`

Fixed: #6549

### DIFF
--- a/.github/workflows/aws-upstream-tests.yml
+++ b/.github/workflows/aws-upstream-tests.yml
@@ -56,7 +56,7 @@ jobs:
           - service: sqs
             tests: TestAccSQSQueue
           - service: ec2
-            tests: TestAccEC2KeyPair_publicKey
+            tests: 'TestAccEC2KeyPair_publicKey|TestAccVPCRoute_timeoutsOnlyChange'
           - service: docdb
             tests: TestAccDocDBCluster_basic
           - service: redshiftserverless
--- a/examples/examples_nodejs_test.go
+++ b/examples/examples_nodejs_test.go
@@ -694,19 +694,12 @@ func TestRegress5219(t *testing.T) {
 // Regress aws_route update failing with "route target attribute not
 // specified" when only customTimeouts changes, see pulumi/pulumi-aws#6549
 func TestRegress6549(t *testing.T) {
-	skipIfShort(t)
-	t.Parallel()
+	opts := nodeProviderUpgradeOpts()
+	opts.skipDefaultPreviewTest = true
 	dir := filepath.Join("test-programs", "regress-6549")
-	cwd, err := os.Getwd()
-	require.NoError(t, err)
-	options := []opttest.Option{
-		opttest.LocalProviderPath("aws", filepath.Join(cwd, "..", "bin")),
-		opttest.YarnLink("@pulumi/aws"),
-	}
-	test := pulumitest.NewPulumiTest(t, dir, options...)
-	test.Up(t)
+	test, _ := testProviderUpgrade(t, dir, opts,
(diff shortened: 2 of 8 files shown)


guineveresaenger:
Thank you for this pull request @pierskarsenbarg. I have a few preliminary questions:

1. Why are we "backporting" an upstream PR? It looks like there was some back and forth - can you give us some context?
2. Please review the [patch policy guidelines](https://github.com/pulumi/platform-providers-team/blob/main/playbooks/Patching%20Upstream.md#patching-policy) to see if a patch is really the best way to resolve this issue. If it is in fact necessary, please file an [additional tracking issue](https://github.com/pulumi/platform-providers-team/blob/main/playbooks/Patching%20Upstream.md#documenting-patches) so that we can trace the reasoning and remove it when the underlying issue is resolved.

Thanks again!

pierskarsenbarg:
@guineveresaenger 

1. "Backport" shouldn't be in the issue title - it's not backporting at all.
2. Do you need another issue other than https://github.com/pulumi/pulumi-aws/issues/6549 ?

corymhall (changes requested):
Thanks for putting this together. The patch itself looks pretty low-risk because it does not introduce any schema or public API changes, and it matches the proposed upstream fix. 

Before merging we should add some more tests:

1. Run the new upstream acceptance test in CI.    The EC2 job in `.github/workflows/aws-upstream-tests.yml` currently runs only `TestAccEC2KeyPair_publicKey`, so `TestAccVPCRoute_timeoutsOnlyChange` was not exercised despite the EC2 job passing. Add the new Route test to the EC2 test expression, ideally alongside `TestAccVPCRoute_basic`.
2. Add a Pulumi-side regression test. We need an example that creates an `aws:ec2:Route`, changes only its `customTimeouts`, performs an actual `pulumi up`, and verifies that the update succeeds without replacing the Route. The test should fail without the patch and pass with it.


Once those are covered, I think this is a reasonable patch to carry. The risk of permanent API divergence is low because it changes only internal update behavior, although upstream may still choose a somewhat different implementation since hashicorp/terraform-provider-aws#47413 has not yet been reviewed or merged.

pierskarsenbarg:
@corymhall I've added in the tests you suggested.

corymhall (approved):

PR #6266, opened 2026-04-07
Add r8 family to instance-types.txt

bastirlnd:
## Summary
- Problem: Instance types of the r8 family are missing at the moment.
- Change: Adds r8* instance types

## Change Type
- [ ] Provider logic only (`provider/`)
- [X] Schema or mapping change (may require SDK regeneration)
- [ ] Upstream or patch pipeline change (`upstream/`, `patches/`, `scripts/upstream.sh`)
- [ ] CI workflow source change (`.ci-mgmt.yaml`)

## Rollback
- Revert plan: Revert change -> rolls back to previous provider schema & SDK files
- Follow-up cleanup: Potentially re-run `make schema && make provider` as well as `make build_sdks` if other changes were merged on top of this.

PR #6652, opened 2026-09-10
Sync instance-types.txt with the AWS EC2 InstanceType enum

SeanWhelan:
## Summary
- Problem: `provider/instance-types.txt` is hand-maintained and has drifted. It lists 854 instance types; the `InstanceType` enum in AWS's EC2 service model lists 1405. So `ec2.InstanceType` cannot express current-generation hardware: no `m8*`, no `c8*`, no `i8g` or `x8g`, nothing from gen-9. #6266 brought the r8 family current, but the M, C, I and X families were missed. This came up at Estuary, where we provision managed AWS data planes with pulumi-aws and wanted `m8id` for a compute-heavy workload. Our own config models use `ec2.InstanceType` as a validation type, so an instance type missing from the enum is rejected at config-parse time rather than being passable as a plain string.
- Change: Adds the 551 missing types from the `InstanceType` enum in `botocore` 1.43.91 (`botocore/data/ec2/2016-11-15/service-2.json`), the list the AWS CLI and SDKs validate against. Additions only, nothing removed or reordered, still sorted under `LC_ALL=C sort`. Nothing already in the file is absent from AWS's enum, so there was nothing to prune.

## Change Type
- [ ] Provider logic only (`provider/`)
- [x] Schema or mapping change (may require SDK regeneration)
- [ ] Upstream or patch pipeline change (`upstream/`, `patches/`, `scripts/upstream.sh`)
- [ ] CI workflow source change (`.ci-mgmt.yaml`)

## Carried Upstream Patch
Not applicable. No patch under `patches/` is added or changed.

## Validation Evidence
- [ ] `make lint` (fails identically on unmodified `master`, see below)
- [x] `make test_provider`
- [x] If schema-affecting: `make schema`
- [x] If schema-affecting: `make build_sdks`
- [ ] If CI source changed: `make ci-mgmt` (not applicable)

### Command output snippets
```text
make test_provider
--- PASS: TestUpstreamLint (243.58s)
ok  github.com/pulumi/pulumi-aws/provider/v7  252.320s  coverage: 3.2% of statements

make build_sdks
BUILD SUCCESSFUL   (all five SDKs, including dotnet build and gradle build)

make lint
Error: can't load config: the Go language version (go1.25) used to build
golangci-lint is lower than the targeted Go version (1.26.6)
```

`make lint` fails the same way on a clean checkout; I stashed the change and re-ran it to check. The pinned `golangci-lint` 2.7.0 is built against an older Go than the repo targets, and this PR adds no Go source. `.ci-mgmt.yaml` sets `lint: false` for this provider anyway.

`instanceTypeName()` in `provider/enums.go` panics on a value with more than one `.`, and a duplicate generated name would break codegen, so I ported it and checked the list before generating. It reproduces all 854 existing names, and the full 1405 give no parse errors and no collisions, including shapes with no precedent in the file such as `p6e-gb200.36xlarge` and `u7inh-32tb.480xlarge`. The generator agrees: `m8id.2xlarge` becomes `M8id_2XLarge` and `M8ID_2_X_LARGE`.

The diff also touches a javadoc example in the Java SDK's `wafv2/RuleGroup.java`, which `pulumi convert` rewrites on regeneration. The same churn appears in #6266 and #5834. It is committed rather than reverted so the tree matches what a fresh `make build_sdks` produces.

## Risk
- Blast radius: Additive. Existing members keep their values and generated names, so nothing that works today stops working. `ec2.InstanceType` already accepts a plain string alongside the enum member in every language.
- Edge cases: Six retired families (`cc1`, `cc2`, `cg1`, `cr1`, `hi1`, `hs1`) are included on the same basis as everything else, in that AWS's enum lists them and the file already carries previous-generation types like `m1`, `c1` and `c3`. Keeping the file a straight mirror also means the next sync against AWS is a no-op instead of re-adding them.

## Rollback
- Revert plan: Revert the commit, restoring the previous `instance-types.txt`, schema and SDK files. No migration or state change is involved.
- Follow-up cleanup: Re-run `make schema && make provider` and `make build_sdks` if other schema-affecting changes land on top.


dyaffe:
We specifically need access to m8id.2xlarge instances and would appreciate a turn around on this.  Let us know if you'd like us to limit it to just adding that instance type.

corymhall:
@SeanWhelan thanks for the PR, and sorry it took so long to review! I merged in some changes before this which caused conflicts on your branch and we can't run our integration tests on forks so I had to merge this change in a separate PR https://github.com/pulumi/pulumi-aws/pull/6658

Issue #6659, opened 2026-09-18
aws.mq.Broker: perpetual empty diff on sharedResources (Computed-only attribute, null in state)


mengesb:
### Describe what happened

`aws.mq.Broker` proposes an update on every `preview`/`up`, permanently, with an empty diff:

```
~  aws:mq:Broker hatchet-dev-mq-broker      update [diff: +sharedResources]
~  aws:mq:Broker hatchet-sandbox-mq-broker  update [diff: +sharedResources]
~  aws:mq:Broker hatchet-prod-mq-broker     update [diff: +sharedResources]
Resources:
    ~ 3 to update
    131 unchanged
```

Detailed diff:

```
+ sharedResources: [unknown]
```

Applying does not settle it. We applied it across two separate deploys; the next preview proposes the same updates again. Every merge performs a no-op in-place update against production brokers, and our scheduled drift-check has gone permanently red as a result.

This started when we upgraded pulumi-aws 7.26.0 → 7.46.0.

### Root cause

`shared_resources` was added to `aws_mq_broker` in terraform-provider-aws **v6.55.0**, reaching pulumi-aws in **v7.37.0** (bisected with `git grep BrokerSharedResource` over `provider/cmd/pulumi-resource-aws/schema.json` across tags: absent through v7.36.0, present from v7.37.0).

Upstream declares it **`Computed: true` with no `Optional`** (`internal/service/mq/broker.go:326`), and the Pulumi schema mirrors that correctly — `sharedResources` is in `properties`, absent from `inputProperties`. It is therefore output-only, and `ignoreChanges` cannot suppress it (verified: neither `"sharedResources"` nor `"shared_resources"` has any effect).

For a RabbitMQ broker with no RAM shares the AWS API returns an empty list, not an error:

```
$ aws mq describe-shared-resources --broker-id b-… --region us-west-1
{ "SharedResources": [] }
```

so the provider takes the `default:` branch and `flattenSharedResources` returns `[]any{}`. But Pulumi state records `sharedResources = null` (key present, value null), and a null value for a `Computed` attribute is indistinguishable from "not yet computed" — so each plan emits NewComputed.

I confirmed this is **not** the update path failing to re-read: a targeted `pulumi refresh --preview-only --target <broker urn>` reports `1 unchanged` against the null-holding state. If Read produced `[]` while Update had written null, refresh would have proposed `null → []`.

[...]

PR #6660, opened 2026-09-19
Omit aws_mq_broker.shared_resources to stop a perpetual empty diff

mengesb:
Fixes the perpetual empty diff on `aws.mq.Broker` reported in #6659.

### The problem

`shared_resources` is `Computed: true` with no `Optional` (`internal/service/mq/broker.go:326`). For a RabbitMQ broker with no RAM shares, `DescribeSharedResources` returns an empty list — no error:

```console
$ aws mq describe-shared-resources --broker-id b-… --region us-west-1
{ "SharedResources": [] }
```

That reaches Pulumi state as `null`, and a null value for a `Computed` attribute is indistinguishable from "not yet computed", so every plan emits NewComputed:

```text
~  aws:mq:Broker my-broker  update [diff: +sharedResources]
      + sharedResources: [unknown]
```

Applying never settles it. Affected brokers propose a no-op update on every `preview` and `up`, which also keeps scheduled drift-checks permanently red.

`ignoreChanges` cannot suppress it — the property is output-only (in `properties`, absent from `inputProperties`), so there is no user-side workaround.

### This change

Adds `"shared_resources": {Omit: true}` to the `aws_mq_broker` override, following the existing `aws_quicksight_*` `{Omit: true}` precedent, with a comment linking the bridge issue so it can be reverted when that lands.

The underlying defect is bridge-level and filed separately as pulumi/pulumi-terraform-bridge#3633 — a Computed-only attribute whose settled value is empty should not generate a diff. This is the stopgap.

The data source type `GetBrokerSharedResource` is deliberately untouched; data sources don't diff.

### Generated code

`make tfgen` and `make generate_{nodejs,python,dotnet,go,java}` were run in a clean Linux container with the toolchain from `mise.toml` (mise 2026.3.7, Go 1.26.6, converter 1.4.0, and the eight pinned provider plugins installed into the workspace `PULUMI_HOME`).

Regeneration also produced ~10 files per language of pre-existing example-formatting drift unrelated to this change — the same files appear when regenerating unmodified `master`. **Those are excluded from this PR** so the diff stays reviewable; happy to include them if you'd rather they landed here.

`make build_go` is not included: the Go SDK verification compile OOMs at the 7.7 GB available to me. Generation completed; only the compile step was skipped, so that's worth a CI check.

### Verification

Schema diff touches exactly one token:

```text
provider/cmd/pulumi-resource-aws/schema.json | 12 +++--- 61 ------
only: "aws:mq/BrokerSharedResource:BrokerSharedResource"
```


corymhall:
@mengesb I'm going to close this PR. See https://github.com/pulumi/pulumi-aws/issues/6659#issuecomment-5763146729

